PR #66, opened 2016-04-19
Remove lowercase flag (Line 155)

Fvbor:
@jasonmunro: Wasn't it planned to remove the lowercase flag from line 155? In the de.php the line is alöready removed ;)

### DIFF
--- a/language/en.php
+++ b/language/en.php
@@ -152,7 +152,6 @@ return array(
     'forward' => false,
     'attach' => false,
     'raw' => false,
-    'flag' => false,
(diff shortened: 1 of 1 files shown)


jasonmunro:
It was the plan, but I was wrong :) I actually do need both versions. Sorry about that!

jasonmunro:
changed my mind again :) I'm going to use css to convert strings to lowercase where I need to, this way we don't have duplicate strings in the translation files for a lowercase version.

PR #68, opened 2016-04-19
Add a missing Parameter

Fvbor:
Add Missing => false to 'Event Deleted'


jasonmunro:
Thanks for catching this!

PR #73, opened 2016-05-19
Update de.php

Fvbor:
Sorry for this Pull Request...
I know, it is only a single row :D


jasonmunro:
Thanks for the update! :)

PR #130, opened 2016-10-05
Add dutch language

DylanMeeus:
Hey there,

I have added translations for the Dutch language. I based myself on the translations file of the French language. There are some double keys in the lists though, should they be removed?  I haven't verified wether or not the doubles appear in all script files, but I'm assuming so.

I have not yet set up the project so I can't verify whether or not this is working. But at least the language file is already present.


jasonmunro:
Thanks for doing this! I will happily test this out and merge it. You are correct, there should not be any duplicate keys - unless they have different capitalization (which I need to fix in the code, but have not gotten around to yet).

DylanMeeus:
@jasonmunro Hey Jason, you're welcome. I don't have an apache server etc set up right now so I couldn't test it myself. There are duplicate keys, most have different capitalization but some have identical capitalization. I don't mind looking around to fix this either when I have some time. 

Made the issue here #132

jasonmunro:
Tested and merged. Found a few typos in the file, easily fixed. Thanks again for doing this!

PR #145, opened 2016-10-30
fix differences between language file and label used in trans()

hiroshitoda:


jasonmunro:
This is great, thanks for doing this!

PR #147, opened 2016-10-30
fix POP3 parser in makeshift.

hiroshitoda:


jasonmunro:
Thank you for doing this!

PR #150, opened 2016-11-17
Create hu.php

dunstn:
Hungarian language file for Cypht


jasonmunro:
Thanks for doing this!

PR #161, opened 2016-12-17
Adding translation for Portuguese Brazil (pt-BR)

ramirovjr:
Copied the en.php file and added the translations for pt-BR locale.


jasonmunro:
Thank you for doing this! I will test this out and get it merged tomorrow.

ramirovjr:
I've translated as best as I could. I think that with some context, some of these translations could be improved.

PR #172, opened 2017-03-12
Remove all click listeners before adding

sjhernes:
Not sure if this would be the best solution, but every time a click-event loads it adds another listener making a simple scroll through N mails gives N² requests on the server. Removing all listeners on click before adding solves this problem.


jasonmunro:
Thanks for the feedback! I will take a closer look. Seems likely I'm doing the wrong thing here, I often do with js (!). I should have some time tomorrow to address the issue.

jasonmunro:
Looks good, thanks again!

PR #182, opened 2017-03-31
hm3.sample.ini: fix ldap_base_dn to ldap_auth_base_dn

iheinrich:
Hi Jason,

Found a typo in hm3.sample.ini: ldap_base_dn should actually be ldap_auth_base_dn.
See lib/auth.php:378.

Thanks for providing Cypht!


jasonmunro:
Thank you for this!

PR #207, opened 2017-06-20
Russian localization and Office365 nux

hartois:


jasonmunro:
Thanks for doing this!

hartois:
I found some more inaccuracies in the translation, but did not have time to make a commit. I'll do it a little later

PR #321, opened 2019-02-22
Fix problems adding imap, smtp and probably more

Razvan0925:
## Pullrequest
When saving a new imap/smtp the email address is encoded ( @ is converted in %40 ) and because of this the authentication fails and it cannot be saved. This is an elegant solutions, otherwise every $form['field'] in every module should be urldecode() -ed.


jasonmunro:
Thank you for this! This bug is something I recently introduced when moving from zepto.js to cash.js. Zepto had a serialize method for forms that cash does not exactly mirror so I wrote our own and did not consider the urlencoding aspects. I will merge this shortly!

PR #326, opened 2019-03-13
allow easier integration into 3rd party systems

kroky:
Hi Jason,

Thanks for the great work on Cypht! We are integrating Cypht into [Tiki](https://tiki.org/) with Marc Laporte. I have found some missing pieces in the core, so we can embed Cypht into Tiki or reuse Tiki classes for session handling, authentication and similar tasks.

In essence, this PR solves 2 issues:
- inability to override session class. Yes, we have site module lib but it is located in Cypht modules dir which we plan to install via composer and do not want to change anything once installed - i.e. don't want to override Custom_Session class there with our own but specify another session class from another location. This is solved via injected config variable containing the session class name.
- inability to return the output of the dispatcher and decide what to do with it on a later stage of the script execution. We need this in order to embed Cypht into Tiki pages as a wiki plugin where we cannot simply echo the contents of Cypht output modules but need to hold on until the whole page is constructed.

I have done the rest of the modifications through config tweaks (mainly turning off specific modules/functions), as to keep the needed changes to Cypht core at the minimum. I think Marc Laporte can show you a demo of the resulting product if you are interested.

If you see any problems with the code or improvements are necessary, I will be glad to discuss!

Regards,
Victor

### DIFF
--- a/lib/dispatch.php
+++ b/lib/dispatch.php
@@ -156,6 +156,7 @@ class Hm_Dispatch {
     public $session;
     public $module_exec;
     public $page;
+    public $output;
 
     /**
      * Setup object needed to process a request
@@ -257,9 +258,10 @@ class Hm_Dispatch {
      */
     private function render_output() {
         $formatter = new $this->request->format($this->site_config);
-        $renderer = new Hm_Output_HTTP();
+        $class = $this->site_config->get('output_class', 'Hm_Output_HTTP');
+        $renderer = new $class;
         $content = $formatter->content($this->module_exec->output_response, $this->request->allowed_output);
-        $renderer->send_response($content, $this->module_exec->output_data);
+        $this->output = $renderer->send_response($content, $this->module_exec->output_data);
     }
 
     /**
--- a/lib/session_base.php
+++ b/lib/session_base.php
@@ -394,6 +394,7 @@ class Hm_Session_Setup {
      * @return string
      */
     private function get_session_class() {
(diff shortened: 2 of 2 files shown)


jasonmunro:
Hi Victor,
This looks great, thanks so much for doing this! I see no reason not to merge the changes, but I want to of course do some testing first :). Also I will add some commented out entries with descriptions in the hm3.sample.ini file for the new settings for others to reference. My only (very minor) concern is saving the output when using the default HTTP renderer as it serves no purpose in that context and uses memory, but I can easily run some analytics to see if it has any measurable impact.

Thanks again for this! Should have this merged tomorrow.

kroky:
Thanks for the quick answer!

I didn't want to do greater changes to the core but if we want to avoid storing unnecessary output as instance variable, maybe we can think about an additional method that outputs the content? Default HTTP renderer will destroy the output after echo-ing to browser but custom ones can just return it.

Also, do you want me to do the same class overriding for custom auth_type (Custom_Auth) class as it seems a bit inconsistent to be able to override only the session one now?

I am open to add more commits here if you prefer to polish this change...

jasonmunro:
Changes are merged. I will follow up with adding some ini comments and using a similar pattern for custom auth. I don't think the output will be a problem but once I have the rest done I will dig into it a bit more. Thanks again!

kroky:
Great, thanks! I will also probably follow up with more updates as we continue the integration.

PR #335, opened 2019-04-02
add cookie_path configuration parameter

kroky:
## Pullrequest
Add ability to specify the cookie path the same way cookie domain can be specified through configuration. It is useful in scenarios where mod_rewrite is used and current Cypht cookie path retrieval mechanism incorrectly puts the cookie inside a directory which is actually a webpage, hence the inability to access those cookies later. E.g. request URI like /tiki-git/Webmail?page=compose will end up with cookie path /tiki-git/Webmail/ which is incorrect as the path is /tiki-git/ only.

### Issues
- [X] None

### Checklist
- [ ] Config Update

### How2Test
Specify a specific cookie path and check cookies are still set - e.g. compose message returns back the proper Message Sent response.

### Todo
- [ ] Changelog

### DIFF
--- a/hm3.sample.ini
+++ b/hm3.sample.ini
@@ -203,11 +203,22 @@ admin_users=
 ; By default Cypht uses the server name used in the request to determine
 ; the domain name to set the cookie for. Configurations that use a reverse
 ; proxy might need to define the domain name used for cookies. Leave this
-; blank to let Cypht auotmatically determine the domain. You can also use
+; blank to let Cypht automatically determine the domain. You can also use
 ; the special value of "none" to force Cypht to NOT set the cookie domain
 ; property at all. This is not recommended unless you know what you are
 ; doing!
 cookie_domain=
+
+
+; Cookie Path
+; -------------
+; By default Cypht uses the request URI to determine the cookie path to set
+; the cookie for. Configurations that use mod_rewrite might need to define
+; the path used for cookies. Leave this blank to let Cypht automatically
+; determine the path. You can also use the special value of "none" to force
+; Cypht to NOT set the cookie path property at all. This is not recommended
+; unless you know what you are doing!
+cookie_path=
 
 
 ; Outbound E-mail Domain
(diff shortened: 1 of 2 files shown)


lib/session_base.php
```diff
@@ -304,6 +304,22 @@ private function cookie_domain($request) {
...
+    /**
+     * @param Hm_Request $request request object
+     * @return string
+     */
+    private function cookie_path($request) {
+        $path = $this->site_config->get('cookie_path', false);
+        if (!$path) {
+            $path = $request->path;
+        }
+        if ($path == 'none') {
```
dumblob: Is this a good idea to check for a special value `none` (which is valid just for the server-side configuration) if it might be forged by the client (see line `314`)?

hm3.sample.ini
```diff
@@ -203,13 +203,24 @@ admin_users=
...
 
 
+; Cookie Path
+; -------------
+; By default Cypht uses the request URI to determine the cookie path to set
+; the cookie for. Configurations that use mod_rewrite might need to define
+; the path used for cookies. Leave this blank to let Cypht automatically
+; determine the path. You can also use the special value of "none" to force
+; Cypht to NOT set the cookie path property at all. This is not recommended
+; unless you know what you are doing!
```
dumblob: The example with `tiki-git` would be of help here as well :wink:.

lib/session_base.php
kroky: I just followed the same way as cookie_domain variable is set. Indeed, if we end up with a URI path of 'none' we might default to an empty cookie path which will be a difference with the old code - it would have used 'none' as the cookie path. I don't mind removing this 'none' functionality from here. Just wanted to keep it consistent with cookie domain one.

hm3.sample.ini
kroky: Sure, added one.

lib/session_base.php
dumblob: @jasonmunro what do you think?

jasonmunro: Sorry for the delay on this. @dumblob makes a good point, though I don't think there is any directly actionable attack vector as the result will likely be you get logged out if you forge 'none' for both cookie path or cookie domain. Regardless, I'm going to go ahead and merge this then fix both methods to eliminate that possibility.

jasonmunro:
Thanks for the PR. I will follow up with the small refactor as discussed in the comments and add some unit tests.

PR #342, opened 2019-04-29
allow usage of custom class for implemeting user configuration persistence

kroky:
We needed a way to store user configuration inside Tiki database as user preferences as we already control authentication using Tiki. Saving to a separate table using DB method did not accomplish what we needed as we didn't go through the normal authentication login form (to trigger initial configuration load) and we also didn't use save settings page. Basically, I needed a way to override `load_user_config_object` function but there was no easy way to do it. I added another configuration option for custom implementation of the user configuration persistence class. Also extended the unit test for that function. Hope all this makes sense!

### DIFF
--- a/hm3.sample.ini
+++ b/hm3.sample.ini
@@ -141,16 +141,19 @@ default_smtp_no_auth=
 
 ; Settings Storage
 ; ----------------
-; Cypht supports 2 methods for saving user settings between logins. File based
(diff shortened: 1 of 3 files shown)


jasonmunro:
Looks good to me, thanks for doing this. I will merge it shortly.

kroky:
Much appreciated!

PR #346, opened 2019-05-21
fix compose page handler module precedence

kroky:
This is a minor fix for the following bug:
- profiles page had the smtp module load_smtp_servers_from_config function executed before profile_data which correctly loaded the smtp server list from config
- compose page was executing load_smtp_servers_from_config after compose_profile_data and thus defined servers in config did not correctly auto-created the default profile when we had exactly one smtp and imap server

Both pages defined the relevant modules to be executed after load_user_data, however alphabetic sort of the names or something else prevented compose page to get the right order.
I merely changed compose_profile_data load requirement to be after load_smtp_servers_from_config as it depends on it to function correctly.


jasonmunro:
Thanks for doing this!

PR #347, opened 2019-05-29
use separate constant for VENDOR_PATH

kroky:
We are storing vendor files in Tiki in a few places and already have the Cypht dependencies installed, so I just needed to make the vendor path configurable by this straightforward update. Hope it makes sense!


jasonmunro:
Looks good, merging now. Thanks!

PR #358, opened 2019-10-23
show alert when trying to sign message without existing profile

yvesdaxmaz:
## Pullrequest

When composing a message under the message textarea there are "Sign" button which as purpose to add signature to a message but, the button isn't self explanatore. So to add some explanation when the button is clicked instead of showing nothing to showcase some interaction of the button. I add an alert which tell the user what to do before that feature to be avalable.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] compose a new message without defining any profile in cypht. And under the message textarea click on the sign button.

### Todo

- [X] None


jasonmunro:
This sounds reasonable, I will review and let you know. Thanks for the PR!

PR #369, opened 2020-01-22
use protected variables in Hm_MIME_Msg class to allow inheritance and…

kroky:
… changes of defined methods

## Pullrequest
Using protected variables to allow inherited classes reuse them and alter behavior of defined methods.

### Issues
I am adding RSVP actions to incoming messages in IMAP module and need to respond with special mime part 'text/calendar' message but current Hm_MIME_Msg class does not support it. Integration with Tiki will support RSVP actions but we can make the code available upstream if you want it. It will be a matter of another PR...

### DIFF
--- a/modules/smtp/hm-mime-message.php
+++ b/modules/smtp/hm-mime-message.php
@@ -11,14 +11,14 @@
  * @subpackage smtp/lib
  */
 class Hm_MIME_Msg {
-    private $bcc = '';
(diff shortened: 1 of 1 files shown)


kroky:
Jason, I managed to reuse Hm_MIME_Msg the way it is, so no need for this PR now. If I have something else, I will submit another one. Thanks!

PR #373, opened 2020-02-11
[Privacy] Don't send referrer information when clicking links

t-m-w:
## Pullrequest
This PR adds a meta tag to pages that prevents referrer information from being transmitted in HTTP request headers when you click links, such as links in emails. Sending that information to external sites unnecessarily exposes the full URL of the cypht page you were on, which includes your cypht instance domain/address, message ID, and other IDs which might be usable for unknown nefarious purposes. The URL often looks something like this: `https://cypht.example.com/?page=message&uid=12345&list_path=imap_0_1234567890&list_parent=imap_0_1234567890&list_page=1`


jasonmunro:
I love this, thanks for the contribution! I'm on PR patrol this weekend and will merge. Thanks again!

PR #375, opened 2020-02-16
Replace loading gradient icon with spinner

yvesdaxmaz:
## Pullrequest

To replace the loading gradient which appear each time cypht performs some background action to load message. I made some update to integrate a spinner which will more visible than the loading gradient

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] Configure a mail server
- [X] Open an inbox and the spinner will appear

### Todo

- [X] None

### DIFF
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -1399,7 +1399,7 @@ class Hm_Output_content_section_start extends Hm_Output_Module {
      * Opens a main tag for the primary content section
      */
     protected function output() {
-        return '<main class="content_cell"><div class="offline">'.$this->trans('Offline').'</div>';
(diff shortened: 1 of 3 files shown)


jasonmunro:
I can't merge this as is. I agree that the current loading gradient is not very noticeable, but this is actually TOO noticeable :) For quick updates this results in a "filcker" of gray over the content cell, and the overlay has some style issues as the spinner is not centered and while visible it creates a horizontal scroll bar.

yvesdaxmaz:
I still working on this to try to make It appear only in certain circumstance and not so often as there are many AJAX request which is run in background

marclaporte:
@yvesdaxmaz: one important use case is to select all emails on the page, and move them to another account. This understandably takes some time.

marclaporte:
Closing this for inactivity. As soon as someone (hopefully @Revisto) is ready to take this on, they can start a new merge request.

PR #376, opened 2020-02-19
Fix user reply to all

adrienmaloba:
## Pullrequest

Remove the REPLY-ALL link when the message comes from a single sender or when the TO and/or CC option has only one email address. By clicking on the REPLY-ALL button, the user's email must not appear in the list of recipients.
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [ ] Configure a mail server
- [ ] Click on the REPLY-ALL link and the sender's email will no longer appear in the list of recipients.

### Todo

- [X] None

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -156,9 +156,9 @@ function format_reply_address($fld, $excluded, $delivered_to) {
             }
         }
         if (!$skip) {
-			if ($v['email'] != $delivered_to) {
-				$res[] = $v;
-			}
+            if ($v['email'] != $delivered_to) {
+                $res[] = $v;
+            }
         }
     }
     if ($res) {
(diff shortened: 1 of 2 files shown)


jasonmunro:
Thanks for doing this! I will review tomorrow.

jasonmunro:
Sadly this is now conflicting due to some other changes. I would like to add these improvements however there are a couple issues with the code I would like  refactored

- First can we re-submit this as 2 PRs, one to remove the delivered-to address from the reply all list, and one to disable the reply-all link when it would produce the same result as reply
- To excluded the delivered to address, lets use the existing $excluded argument rather than add another one. so we could add the delivered-to address to this array https://github.com/jasonmunro/cypht/blob/master/modules/core/message_functions.php#L108 and make sure it's passed into all calls to format_reply_address()
- The check for multiple recipients should use some existing more robust address field parsing instead of just split on comma as that could be a comma in a label and not really > 1 address. Counting the result of split_address_fld() should work here. And lets use count() rather than sizeof() please.

Thanks again for doing this. Please don't hesitate to let me know if you have any questions about the above! I will happily merge new PRs with the above changes.

adrienmaloba:
Thank you for reviewing the code.
I will take the time to make the changes according to your remarks.

PR #383, opened 2020-04-04
prevent last recipient from disappearing

adrienmaloba:
## Pullrequest
When we click on REPLY-ALL, all recipients at **To** should remain on the mailing list.

### Issues
- [X] None

### Checklist

- [X] None

### How2Test

- [ ] Configure a mail server
- [ ] Click on the REPLY-ALL link and and there, on the page for sending a new email, all recipients should be in the mailing list.

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -251,12 +251,9 @@ function process_address_fld($fld) {
         }
         if (!$email) {
             foreach ($vals as $i => $v) {
-                $v = trim($v, '<>\'"');
-                if (is_email_address($v)) {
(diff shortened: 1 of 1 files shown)


jasonmunro:
Merge to master. Thanks for the fix!

jasonmunro:
while this fixes the bug, it actually breaks proper address parsing. Working on a fix now

marclaporte:
@jasonmunro: Is this the fix? https://github.com/jasonmunro/cypht/commit/2b3a87f77e07140f405c4b05310906fcc746fd89

PR #385, opened 2020-04-10
add WEB_ROOT constant to allow theme and static asset files loaded fr…

kroky:
…om a different web root URL

Hey Jason,

Just realized that static assets from smtp module were not properly integrated into Tiki as we serve cypht from a different URL than the site/ one you assume as the default URL. I had to add an empty WEB_ROOT constant and prepend all those <link> and <script> tag urls. By default, it stays empty, so previous behavior stays intact. For us with Tiki, I override it to point to the proper vendor place where cypht resides.

Let me know if you see any issues with that!

Not sure we need it for all scripts/*.php files but added there just in case. Also spotted a missing VENDOR_PATH in the api and added that as well.

Btw, running unit tests with phpunit 5 and 6 fails for different reasons. Which version of phpunit do you use to run tests?

### DIFF
--- a/index.php
+++ b/index.php
@@ -14,6 +14,7 @@
  */
 define('APP_PATH', '');
 define('VENDOR_PATH', APP_PATH.'vendor/');
+define('WEB_ROOT', '');
 define('DEBUG_MODE', true);
 define('CACHE_ID', '');
 define('SITE_ID', '');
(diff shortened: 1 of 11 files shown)


jasonmunro:
@kroky this here is how we deal with phpunit version vs PHP versions. This file also shows some of the other bits that need to be in place for the tests to run. Not ideal I know but I have not had any time to work on cleaining it up.

https://github.com/jasonmunro/cypht/blob/master/.travis/setup.sh#L83-L102

jasonmunro:
Thanks for the PR. Merged to master

kroky:
Thanks, I will try running the tests locally next time!

PR #386, opened 2020-04-15
make actions copy and move available on mobile devices

yvesdaxmaz:
## Pullrequest

When select message in on the page message_list, copy and move action are not available as It's on wide screen, I made update to remove code which check if the page is viewed on mobile device to not render copy and move actions.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] Open cypht on a mobile device browse to any mailbox and select one or multiple message
- [X] You'll notice the presence of copy and move buttons in the action list

### Todo

- [X] None


jasonmunro:
merged to master, thanks for the PR!

PR #387, opened 2020-04-16
Add warning went send message with sent folder undefined

yvesdaxmaz:
## Pullrequest

Made update to alert user when he try to send a message without defining a SENT FOLDER where sent messages should be saved.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] Compose a new message without defining sent folder 
- [X] There should be a warning alert which show "Unable to send message: Please configure your folder to save sent messages (Sent Folder)"
### Todo

- [X] None

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -412,14 +412,12 @@ class Hm_Handler_process_compose_form_submit extends Hm_Handler_Module {
         if (!array_key_exists('smtp_send', $this->request->post)) {
             return;
         }
-
         /* missing field */
         list($success, $form) = $this->process_form(array('compose_to', 'compose_subject', 'compose_smtp_id', 'draft_id'));
         if (!$success) {
             Hm_Msgs::add('ERRRequired field missing');
             return;
         }
-
         /* defaults */
         $smtp_id = server_from_compose_smtp_id($form['compose_smtp_id']);
         $to = $form['compose_to'];
@@ -431,7 +429,6 @@ class Hm_Handler_process_compose_form_submit extends Hm_Handler_Module {
             'draft_subject' => $form['compose_subject'],
             'draft_smtp' => $smtp_id
         );
-
         /* msg details */
         list($body, $cc, $bcc, $in_reply_to, $draft) = get_outbound_msg_detail($this->request->post, $draft, $body_type);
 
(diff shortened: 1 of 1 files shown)


marclaporte:
What about when I delete a message, does it go to trash? If trash folder is not defined, then what?

marclaporte:
I lost a message again today because I forgot to set the sent on a new mail account.

jasonmunro:
@marclaporte deleting a message with no trash setup while similar, should have it's own PR. @yvesdaxmaz I can't merge this as is because it will make it impossible to send a message without a configured sent folder. I agree the user should be aware of the missing sent folder, but it should not be impossible to send without it.

marclaporte:
@adrienmaloba: Since Yves is unavailable nowadays, please take over.

Thanks!

henrique-borba:
> @marclaporte deleting a message with no trash setup while similar, should have it's own PR. @yvesdaxmaz I can't merge this as is because it will make it impossible to send a message without a configured sent folder. I agree the user should be aware of the missing sent folder, but it should not be impossible to send without it.

In this case @adrienmaloba @marclaporte,  if we just need to warn the user maybe we can just replace 
`Hm_Msgs::add("Message Sent");` (Line 507) with:

```PHP
 $specials = $this->user_config->get('special_imap_folders', array());
        if(array_key_exists($smtp_id, $specials) || empty($specials)){
            if(!array_key_exists('sent', $specials[$smtp_id]) || empty($specials[$smtp_id]['sent'])) {
                Hm_Msgs::add('Message Sent.');
                Hm_Msgs::add('ERRPlease configure your folder to save sent messages (Sent Folder)');
         }
 }
```

It is simple and won't require JS or ajax calls, but you will lose your first email.

dumblob:
> It is simple and won't require JS or ajax calls, but you will lose your first email.

Oh no, please. I've lost already several emails during the past few years due to MUAs doing such mistakes. It's of the highest priority to not lose any information the user has created or received.

henrique-borba:
> > It is simple and won't require JS or ajax calls, but you will lose your first email.
> 
> Oh no, please. I've lost already several emails during the past few years due to MUAs doing such mistakes. It's of the highest priority to not lose any information the user has created or received.

I just chatted with @adrienmaloba and he is actually pretty advanced on the solution to check it before the email is sent so don't mind my last comment 👍

henrique-borba:
Fixed in https://github.com/jasonmunro/cypht/pull/493

PR #389, opened 2020-04-22
make sure to exclude delivered to address from reply-all link

adrienmaloba:
## Pullrequest

By clicking on the REPLY-ALL link, the user's email must not appear in the list of recipients.
### Issues

- [X] None

### Checklist

- [x] None

### How2Test

- [ ] Configure a mail server
- [ ] Click on the REPLY-ALL link and the sender's email will no longer appear in the list of recipients.

### Todo

- [X] None

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -105,18 +105,22 @@ function reply_to_address($headers, $type) {
     $msg_cc = '';
     $headers = lc_headers($headers);
     $parsed = array();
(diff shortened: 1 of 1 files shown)


jasonmunro:
This looks great, thanks! I'm going to merge this shortly. The only change I will make after that is to check to make sure delivered-to is actually present in $headers. It _should_ always be there but this is E-mail so you never know :)

PR #390, opened 2020-04-23
Remove REPLY-ALL link when there is only one recipient

adrienmaloba:
## Pullrequest

Remove the REPLY-ALL link when the message comes from a single sender or when the TO and/or CC option has only one email address.
### Issues

- [X] None

### Checklist

- [x] None

### How2Test

- [ ] Configure a mail server

### Todo

- [X] None

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -224,6 +224,9 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                 if ($reply_all || $size > 1) {
                     $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
                 }
+                else {
+                    $txt .= ' | <a class="reply_all_link hlink disabled_link">'.$this->trans('Reply-all').'</a>';
+                }
                 ' | <a class="forward_link hlink" href="?page=compose&amp;forward=1'.$reply_args.'">'.$this->trans('Forward').'</a>';
             if ($msg_part === '0') {
                 $txt .= ' | <a class="normal_link hlink msg_part_link normal_link" data-message-part="" href="#">'.$this->trans('normal').'</a>';
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -208,13 +208,22 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
             if ($this->get('list_headers')) {
                 $txt .= format_list_headers($this);
             }
+            $reply_all = false;
+            if (array_key_exists('cc', lc_headers($headers))) {
+                $reply_all = true;
+            }
+            $addr = split_address_fld($headers['To']);
+            $size = count($addr);
+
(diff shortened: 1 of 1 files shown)


dumblob:
Do we really want this? What is the motivation?

I'm asking, because this violates the basic user experience rule to not change layout of widgets automatically (layout of widgets shall be allowed to change only when directly triggered by the user, not by some "smart").

This is actually a perfect counter-example to this basic UI rule as there is even no way the user could guess why the layout changed and why the button isn't there.

IMHO a better solution would be to inactivate the button instead of hiding it. But even in that case I still don't see any reason why would we want it anyway as it's perfectly valid to reply to all while getting just one (or even none in some cases) recipient. That's why I'm really interested what is the driver behind this change?

adrienmaloba:
Hi @dumblob,

This change is made in the sense of keeping only **REPLY** in the interface when the mail concerns only one recipient. Since both links lead to a single result, it would be a good idea to keep only one link and also to avoid confusion. This would make a difference when there are several recipients and when there is only one.

The **REPLY-ALL** link will be removed from the interface, that's why I changed the title of the Pull Request from **Hide** to **Remove**.

Thank you for your remarks.

marclaporte:
Here is an example from Gmail:

marclaporte:
@dumblob: Thanks for sharing your concern.

"What is the driver behind this change?"

I want a cleaner interface, and not have options that add no value. I see what you mean about changes to the interface, and yes, we could gray it out to keep interface more consistent.

It's a trade-off. What do others think?

Thanks!

P.S. Later on, I want to improve the UI more by hiding rarely used actions, like raw, all headers and copy. But that will be another discussion at that time.

dumblob:
>P.S. Later on, I want to improve the UI more by hiding rarely used actions, like raw, all headers and copy. But that will be another discussion at that time.

I'm not against hiding whatever stuff if it won't be different across the "default view" of different emails. In the google interface, the default view (i.e. the view one sees for each email right after opening it) doesn't change at all. What changes is the content of the menu, which is fine, because that's not what a user has "burned" in her head as expectation "what comes next" when opening an email.

I think I should have written the UX/UI rule more precisely, so here we are :wink::

>**Never change** spatial properties (layout, dimensions, shape, ...) of widgets automatically across instances of a **default view** of the domain in question (e.g. email reading) unless explicitly requested by the user (do not confuse "requested" with "triggered").

marclaporte:
Ok, let's see what @jasonmunro thinks :-)

jasonmunro:
@marclaporte I agree I would prefer to keep it present and disable it. Honestly this is a pretty minor UI tweak, but I do agree with @dumblob on the importance of consistent positioning of UI elements. I have no problem with the way it is now, but I'm also fine with disabling it for only 1 recipient.

marclaporte:
@dumblob @jasonmunro: I like it!

@adrienmaloba: please adjust PR for present and disabled.

Thank you!

adrienmaloba:
> @dumblob @jasonmunro: I like it!
> 
> @adrienmaloba: please adjust PR for present and disabled.
> 
> Thank you!

I just added a new commit that disables the link instead of removing it.

jasonmunro:
Looks good, thanks for this!

PR #408, opened 2020-06-18
Fix theme issue when selecting a folder

lyabs243:
## Pullrequest

On the "Folders" page in the "Archive Folder" section, we see that the "Cancel" link is not easy to notice because the font size is very small, we also see that it is not only too far away, but also is also not responsive, when we reduce the browser window it changes the appearance of the other components hence the importance of placing it in a fixed place easily visible to the user.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/imap_folders/modules.php
+++ b/modules/imap_folders/modules.php
@@ -323,21 +323,21 @@ class Hm_Output_folders_archive_dialog extends Hm_Output_Module {
         if (!$archive_folder) {
             $archive_folder = $this->trans('Not set');
         }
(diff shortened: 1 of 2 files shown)


jasonmunro:
Thanks for doing this! I'm happy to merge this change however the indentation in modules/imap_folders/modules.php appears to be off by 1 (the lines are all indented 4 spaces too far). Could you fix that bit? Thanks!

lyabs243:
Hi @jasonmunro i just updated code and fix indentation issue. 
Thank you.

marclaporte:
For the record, I am not seeing a change of behavior.

Issue #368, opened 2020-01-11
Saved search: should we be able to pick our own label?


marclaporte:
The saved search module is great!
https://github.com/jasonmunro/cypht/tree/master/modules/saved_searches

The name of the link/bookmark is the search term. But perhaps someone would want something more descriptive?

marclaporte:
I confirm is works with latest code. Thanks!!

PR #409, opened 2020-06-23
Add form to update token of the saved search

yvesdaxmaz:
## Pullrequest

After performing a search we are able to save the search term in cypht for later usage. The token used to reference the saved search is based on the search term. I made update which provide to user a form where they can provide a custom name to represent that saved search.

### Issues

- [X] #368 

### Checklist

- [X] None

### How2Test

- [X] Perform a search
- [X] Save the search results by clicking on the Save search Button
- [X] Click on the Edit Icon to show the form
- [X] Enter a new token which will represent the saved search and click on the UPDATE button under the form
- [X] Check the list of saved search in the sidebar you'll see the new token listed 

### Todo

- [X] None


modules/saved_searches/setup.php
```diff
@@ -10,7 +10,7 @@
 add_handler('ajax_save_search', 'save_search', true, 'core');
 add_handler('ajax_save_search', 'language', true);
 add_handler('ajax_save_search', 'date', true, 'core');
-add_handler('ajax_save_search', 'http_headers', true, 'core');
```
jasonmunro, line 13: This looks like an inadvertent change ... ?

yvesdaxmaz: exactly, I made few changes and fix that mistake

jasonmunro:
@yvesdaxmaz looks better, I will test it out locally and let you know if I find any problems. Thanks!

PR #418, opened 2020-07-31
add default sort order for imap folders - arrival or sent date

kroky:
This PR actually tries to solve 3 issues coming from discussion: https://github.com/jasonmunro/cypht/issues/378

1. Add a site setting for default date sort order - up to now it defaults to arrival date but in some cases, the actual sent date (date from msg header) is preferred. Now it can be configured.

2. Show arrival date in extended headers. We were only showing date header which is a bit confusing when coming from message list where it lists the arrival date.

3. Fix sorting by sent date. Server correctly sorted messages and returned them in a formatted message list. However, there was one more sorting on the client side which used the timestamp of each row which came from the arrival date. Thus, sorting by sent date did not actually produce any different results - it was still sorting by arrival date. I have fixed this by actually changing the date column in the message list to display sent date when sorting by sent date. This requires clearing the imap page cache before reloading the form as the returned formatted message list is going to be different.

Let me know if you see any issues.

### DIFF
--- a/modules/core/functions.php
+++ b/modules/core/functions.php
@@ -517,6 +517,18 @@ function start_page_opts() {
 }}
 
 /**
+ * List of valid default sort order options
+ * @return array
+ */
+if (!hm_exists('default_sort_order_opts')) {
+function default_sort_order_opts() {
+    return array(
+        'arrival' => 'Arrival Date',
+        'date' => 'Sent Date',
+    );
+}}
+
+/**
  * See if a host + username is already in a server list
  * @param class $list class to check
  * @param int $id server id to get hostname from
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -243,6 +243,25 @@ class Hm_Handler_process_start_page_setting extends Hm_Handler_Module {
 }
 
 /**
+ * Process input from the the default sort order setting in the general settings section.
+ * @subpackage core/handler
+ */
+class Hm_Handler_process_default_sort_order_setting extends Hm_Handler_Module {
+    /**
+     * Can be one of the values in start_page_opts()
+     */
+    public function process() {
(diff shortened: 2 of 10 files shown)


marclaporte:
Thank you Victor!  That was way more work than I would have expected :-)

So now let's see what @jasonmunro thinks.

Thanks again!

jasonmunro:
Initial review looks good, thanks @kroky I want to test it out locally but will either merge or follow up with concerns once I do.

jasonmunro:
@kroky nicely done, very thorough! The only thing jumping out at me is that the client side sort options on combined views always say "arrival" even if I have sort by date as my default sort order. I'm pretty sure this is just the label. I'm going to merge and confirm. Thanks again!

jasonmunro:
@kroky just FYI small fix out of the gate - if you have never set the default sort order it defaults to false which causes IMAP mailbox page views to fail (combined pages still work). This is almost handled properly in process_sort_arg() however the value for default is from the user settings, and if that is unset its default is false. Changing that to arrival fixes the issue (line 586 imap/handler_modules.php

jasonmunro:
@kroky FYI, here is what I committed after merging this PR:

https://github.com/jasonmunro/cypht/commit/0964b6628cbb15fceff4a7ec6239c54d4c64804a
https://github.com/jasonmunro/cypht/commit/3920c50b6b67abebc21d6edd2842dd43cb0ec25a

I don't expect contributors to update unit tests and the fixes were minimal just wanted you to see the follow up.
Thanks again!

kroky:
Thanks Jason! I saw the updates and agree it's better to have a sensible default value like 'arrival' (date) instead of false.
I am usually running the unit tests before doing a PR but maybe I didn't do this after the final round of changes here. Will take a note on this one as well. We need to keep the unit tests in good shape.

PR #422, opened 2020-08-13
make copy/move buttons work in advanced search

kroky:
In connection with 
https://github.com/jasonmunro/cypht/pull/374

Copy/Move buttons were missing their event handlers in advanced search. This PR should add those.


jasonmunro:
merged, thanks!

PR #434, opened 2020-09-24
Expose default sort order even on combined message views

kroky:
This is because imap message list can include the right timestamp to sort by. If we don't expose the list_sort parameter which is the default sort order in combined message views, we end up with using internal "arrival" date by default even if the default sort order is changed to Send date... see also #418


jasonmunro:
Thanks for the PR!

Issue #439, opened 2020-10-06
Add Archive action to lists (All Mail, Basic Search, Advanced Search, etc.)


marclaporte:
We have https://github.com/jasonmunro/cypht/issues/228   Great!

![2020-10-06_152437](https://user-images.githubusercontent.com/1004261/95250739-829bfc80-07e8-11eb-889f-001c97c5ee2d.png)

Please keep in mind that archiving acts differently in each account (it is configurable) so if we are in "All Mail" interface, each mail needs to be archived as if we launched the action directly on the email.

jasonmunro:
PR for this has been merged so pretty sure we can close this out @marclaporte

marclaporte:
Testing with latest Cypht from Packagist bundled in Tiki:
I see the buttons in advanced search and all mail, but clicking on it does nothing. Perhaps I have some sticky cache?

jasonmunro:
I actually did not test the archive process. Do you have archive folders setup for your accounts?

marclaporte:
When archiving from an email view, Cypht uses a default pattern (which I am OK with) and it creates folders as needed.

So I usually do not set anything. I expect archive from search or list to work the same way.

jasonmunro:
I respectfully disagree about that. An archive folder is IMO like a special folder (sent/trash/drafts) and auto-creating those is I think a presumptuous action for a webmail client. Not to say we could not auto-create one via the nux module or a different avenue, but at the time of an archive action is IMO too late. Auto-creating also makes some assumptions I don't think we want to make (i18n for example).

marclaporte:
I see what you mean. The current behavior is not what I was expecting, but that is what @yvesdaxmaz coded and I felt it was good enough (and we could always improve later). The archive discussion was here: https://github.com/jasonmunro/cypht/issues/228

I use "Archive" all the time. But other users may prefer to only have the archive button if the archive feature is set up. But then, we'll get feedback from @dumblob about how buttons should not change :-)

So @jasonmunro, how shall we proceed now?

marclaporte:
Thank you @jasonmunro for https://github.com/jasonmunro/cypht/commit/268ec8a5d8c1f8a2d9296e2a5cb5907b476434d9

I will test.

marclaporte:
When I try to archive a few messages from a list, I get "An error occurred trying to archive some messages!"

jasonmunro:
[...]

PR #441, opened 2020-10-07
Add Archive action to lists (All Mail, Basic Search, Advanced Search, etc.)

yvesdaxmaz:
## Pullrequest

Add **Archive Button** in the message heading list to allow archiving of any message which has their checkbox is checked
### Issues

- [X] #439

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -819,7 +819,7 @@ class Hm_Handler_imap_message_action extends Hm_Handler_Module {
     public function process() {
         list($success, $form) = $this->process_form(array('action_type', 'message_ids'));
         if ($success) {
-            if (in_array($form['action_type'], array('delete', 'read', 'unread', 'flag', 'unflag'))) {
+            if (in_array($form['action_type'], array('delete', 'read', 'unread', 'flag', 'unflag', 'archive'))) {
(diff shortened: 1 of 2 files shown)


modules/imap/handler_modules.php
```diff
@@ -836,6 +837,21 @@ public function process() {
                             }
                         }
                     }
```
jasonmunro, line 839: If I'm reading this correctly the failure to find a configured archive folder for the IMAP server results in using a default folder named "Archive", then attempting to move a message to this folder. If that folder does not exist it will fail (or maybe some IMAP servers will auto-create I don't know).

Please remove this default. If the server does not have a configured archive folder just skip it.

jasonmunro:
Thanks for doing this. Please see the inline comment on the Code page: https://github.com/jasonmunro/cypht/pull/441/files

modules/imap/handler_modules.php
yvesdaxmaz: I have made changes, I remove the previouse behavior which create the Archive folder when the Archive Folder is not defined. Now when the folder is not set and you try to archive mail within the message_list page no feature action will be perform the the action will be skipped.

PR #442, opened 2020-10-20
Added support for the database port

KosuDS:
Added support for the database port


jasonmunro:
Thanks for doing this. I'm slightly refactoring the code without any functional changes, adding a default to the hm3.sample.ini file, and checking the unit tests, but it all looks good.

PR #443, opened 2020-10-22
fix default sort order label (Arrival vs Sent Date) throughout the codebase

kroky:
We need to load the default sort order setting in several output modules and  thus I moved it out from imap module into the core module and loaded it where needed - search, advanced search, message list headings, etc. - all forms that contain the sort dropdown. This PR complements #418 and should be the final one to finish this update.

### DIFF
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -243,6 +243,19 @@ class Hm_Handler_process_start_page_setting extends Hm_Handler_Module {
 }
 
 /**
+ * Check the default sort order
+ * @subpackage core/handler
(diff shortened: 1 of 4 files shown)


kroky:
Jason, let me know if there is an easier way to access user settings in output modules - I think we need to pass them by a handler but sometimes the setting is needed in multiple modules, so it turns out that the handler needs to be called in several places. Have you considered accessing user or site configuration directly in output modules? Sounds logical to be able to access configuration items that are directly related to output in output modules. Maybe you have other objections of not supplying them in the output modules?

PR #445, opened 2020-11-10
 Allow Cursor to be placed according to action

mnsuccess:
## Pullrequest

After this button/action, cursor should be in

Forward: "to"
Reply: top of text area

### DIFF
--- a/modules/smtp/site.js
+++ b/modules/smtp/site.js
@@ -210,18 +210,6 @@ var delete_attachment = function(file, link) {
     return false;
 };
 
-var get_url_parameter = function (sParam){
-    var sPageURL = window.location.search.substring(1);
-    var sURLVariables = sPageURL.split('&');
-    for (var i = 0; i < sURLVariables.length; i++){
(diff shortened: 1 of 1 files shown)


modules/smtp/site.js
```diff
@@ -226,5 +244,9 @@ $(function() {
             toggle_recip_flds();
         }
         $('.delete_attachment').on("click", function() { return delete_attachment($(this).data('id'), this); });
-    }
+        if (get_url_parameter('reply') === '1'){
```
jasonmunro, line 247: I think get_url_paramater is a bit much for this simple purpose, also we should check for both reply and reply_all. 

Something simple like this should work fine:

```
if (window.location.href.search('&reply=1') !== -1 || window.location.href.search('&reply_all=1') !== -1) {
            replace_cursor_positon ($('textarea[name="compose_body"]'));
}
```

jasonmunro:
Thanks for the follow up, looks good!

PR #446, opened 2020-11-17
[NEW]  add mark as unread on the inline message

mnsuccess:
## Pullrequest
When we read an email, there should be a way to mark as unread.

Screenshot 2020-11-18 at 00 09 31
 this MR add support of that


jasonmunro:
Great idea and thanks for the PR!

PR #449, opened 2021-01-04
use default sort order for search/advanced search pages

kroky:
Search and Advanced Search ajax pages did not correctly load the default sort order setting and thus were still using Arrival date instead of Sent date even if the setting was enabled. This PR tries to correct the problem by exposing default sort order in case list_sort is empty - e.g. when search pages are using the imap message list functions.


jasonmunro:
Thanks for doing this!

PR #454, opened 2021-02-23
use uniqid instead of php array indexing for server list

kroky:
(imap, jma, smtp, pop3, feeds) but keep backward compatibility - should help keeping special folder assignments when deleting imap servers.

https://github.com/jasonmunro/cypht/issues/349

Newly added servers use uniq ID, so deleting or reordering server list won't negatively affect other servers or special folder assignments. Existing settings are kept as integer IDs.

Fixed an issue with tests but please double check!


marclaporte:
For the record, a fix of this commit: https://github.com/jasonmunro/cypht/commit/b0a7f86d22173d6c52536873823fa41b065b2dd2

marclaporte:
Another: https://github.com/jasonmunro/cypht/commit/f5c3585cac53b4743d1023d1f67a3cb45fe2c11f

marclaporte:
Another attempt: https://github.com/jasonmunro/cypht/pull/591

PR #462, opened 2021-04-04
add a default text to the combined inbox page

azibom:
Hi, I am azibom from tikiwiki company
I done this task "https://avan.tech/item38376-Cypht-in-Tiki-no-page-2-for-messages-in-combined-view-in-one-mailbox-is-OK"
with is about adding a default text in the end of the combined index page and I add it now and you can test it with going to "Everything" page ("/?page=message_list&list_path=combined_inbox")

Screenshot from 2021-04-04 18-27-03


jasonmunro:
thanks!

PR #473, opened 2021-05-07
Revert "use uniqid instead of php array indexing for server list"

jasonmunro:
Reverts jasonmunro/cypht#454

This PR was well-intentioned and I appreciate the effort but it has broken a lot of things since and I don't know what else might be effected that I have not discovered yet. Not sure if it is even possible to revert properly at this point but looking into it.


henrique-borba:
@jasonmunro @marclaporte for me is a go! New PR's won't be affected as I tested with. Anyway, after the merge if any bug show up I will fix ASAP

kroky:
Jason, there were already bugs with the previous approach - like special folder assignment getting wrong when adding/removing mail servers. I think ordering by integer array indexes rather than a unique hash key would introduce more problems down the road than the ones you possibly found now. Can you share what kind of problems did you find? I could have fixed those...

jasonmunro:
@kroky
I admit that it was an initial bad design choice to use integer indexes as "ids" for servers, however they actually work fine for all cases except one: persistent relationships to servers that use the id across logins. I thought about whether to revert this change for a long time and honestly almost did not. The bugs that came up were mostly due to input expecting an integer but instead getting a string which is not hard to fix but easy to miss. If we were to make a change similar to this PR, it should be global for all code that uses the server list base class, and I would lean toward using the existing "id" index and making it an integer id property such that it does not change when adding/removing servers to the list but keeping it a unique integer id.  For the special folders problem I will probably fix that code to associate special folders with an IMAP account using the IMAP server name and username rather than by id. 

I very much appreciate your work here so please don't be discouraged! There is no real right or wrong here just trying to do the best I can to keep things working as well as possible.

kroky:
Sure, I was just suggesting that I could fix the issues. If you have a simpler idea to fix the special folder assignment problem, that's great! No worries on my side.

PR #477, opened 2021-05-27
Send and archive button when replying emails

henrique-borba:
This update allow users to easily archive emails after reply by providing a "Send & Archive" button on the compose page.

It won't be affected by https://github.com/jasonmunro/cypht/pull/473


marclaporte:
Here is related feature request: https://github.com/roundcube/roundcubemail/issues/5937

PR #494, opened 2021-06-30
Draft attachment support

henrique-borba:
## Pullrequest
This PR allows attachments to be used with the IMAP draft feature.


jasonmunro:
Awesome thanks for doing this! Looks like maybe missed updating a call to `attache_file()` in modules/imap/handler_modules.php

henrique-borba:
> attache_file()

I can't find any `attach_file` call in my imap/handler_modules.php

henrique-borba:
> Awesome thanks for doing this! Looks like maybe missed updating a call to `attache_file()` in modules/imap/handler_modules.php

Found it

jasonmunro:
> > Awesome thanks for doing this! Looks like maybe missed updating a call to `attache_file()` in modules/imap/handler_modules.php
> 
> Found it

I don't think that fix is correct. No need to set imap_draft to true there it defaults to false which is fine. However looks you changed the signature of attach_file() to have a new first argument ("$atts"), so I assumed you need to update the call in the imap module set to pass whatever is expected there, or am I missing something?

henrique-borba:
> > > Awesome thanks for doing this! Looks like maybe missed updating a call to `attache_file()` in modules/imap/handler_modules.php
> > 
> > 
> > Found it
> 
> I don't think that fix is correct. No need to set imap_draft to true there it defaults to false which is fine. However looks you changed the signature of attach_file() to have a new first argument ("$atts"), so I assumed you need to update the call in the imap module set to pass whatever is expected there, or am I missing something?

True, I also need to test this one while forwarding the email

henrique-borba:
I preferred to update the function signature and ensure compatibility with other calls. See if it looks right now.

jasonmunro:
> I preferred to update the function signature and ensure compatibility with other calls. See if it looks right

Agreed and it looks better thanks. Seems like we are falling back to legacy draft saving when IMAP is not available is that right? I would prefer we remove all legacy draft support :)

jasonmunro:
merging to master to test out, thanks again! We can remove legacy draft support in another PR

PR #499, opened 2021-07-13
Fix reply to issue

Revisto:
Hello, I am Revisto from WikiSuite project.
I did this [Cypht: When we reply to an email we have sent, it proposes our address](https://avan.tech/item60589-Cypht-When-we-reply-to-an-email-we-have-sent-it-proposes-our-address) which was related to this [issue](https://github.com/jasonmunro/cypht/issues/384). I handled the situation when replying to myself, it shows the other person's email in the 'message to' section.

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -114,11 +114,16 @@ function reply_to_address($headers, $type) {
     if ($type == 'forward') {
         return $msg_to;
     }
-    foreach (array('reply-to', 'from', 'sender', 'return-path') as $fld) {
-        if (array_key_exists($fld, $headers)) { 
-            list($parsed, $msg_to) = format_reply_address($headers[$fld], $parsed);
-            if ($msg_to) {
-                break;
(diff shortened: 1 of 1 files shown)


marclaporte:
There are commits related to your factory reset enhancement.

Revisto:
> There are commits related to your factory reset enhancement.

Yes, I made a mistake, I made a branch from the factory reset branch. I will fix it.

jasonmunro:
Not sure this is a good idea, and even if so, I'm pretty sure this change has unintended consequences. If I am replying to a message, even one sent from one of my own addresses, it should pick the sender of the message from the available header fields. Why would I want it to populate the To: address when I selected reply? Also this change prioritizes return-path in all cases which I am pretty sure is not what we want.

marclaporte:
@Revisto Please research and report on how other webmails do it, especially Gmail and Roundcube.

Thanks!

marclaporte:
Fixed here: https://github.com/jasonmunro/cypht/pull/530

PR #489, opened 2021-06-02
Cypht  view on a folder in mobile mode: Should have a search box

mnsuccess:
## Pullrequest
This  PR  allow Cypht  view on a folder in mobile mode: to have a search box

the search box was missing in mobile mode

PR #501, opened 2021-07-17
Farsi Cypht

Revisto:
## Pull request
Hi. This is Revisto from Wikisuite project.
I've worked on adding Farsi language to Cypht that included translating words and fixing many problems in RTL.
This is the [task](https://avan.tech/item59043-Review-status-of-Farsi-in-Tiki-RTL-is-perhaps-good-perhaps-broken) which I went further than that and I started fixing bugs that happened in RTL not reporting them only.


jasonmunro:
Thanks for doing this! Looks great except for how the font files are being included.  We need to move the /assets/fonts/ directory to /modules/core, and move the font-face CSS to PHP and use the WEB_ROOT constant and new path in the url() definitions. Here is the code change (https://gist.github.com/jasonmunro/9621d58601abb189a2fa38e3bd6bac84) This way the includes will work in both debug and production mode along with various URL schemes. Also are the font files absolutely needed? I would imagine someone wanting to use Farsi would have the correct fonts installed? If we need them no problem with the above changes. If you can update your PR that would be great, otherwise I will merge it and apply the fixes. Thanks again!

jasonmunro:
Also looks really great :)
farsi

jasonmunro:
Just noticing one more issue, the sort dropdown on message list view is now misplaced - looks like a CSS rule problem

Revisto:
> Thanks for doing this! Looks great except for how the font files are being included. We need to move the /assets/fonts/ directory to /modules/core, and move the font-face CSS to PHP and use the WEB_ROOT constant and new path in the url() definitions. Here is the code change (https://gist.github.com/jasonmunro/9621d58601abb189a2fa38e3bd6bac84) This way the includes will work in both debug and production mode along with various URL schemes. Also are the font files absolutely needed? I would imagine someone wanting to use Farsi would have the correct fonts installed? If we need them no problem with the above changes. If you can update your PR that would be great, otherwise I will merge it and apply the fixes. Thanks again!

Great, I will fix the font files location and move the font-face to PHP.
And I believe the fonts are kind of necessary. Farsi People should have a web extension that changes the webpage's fonts to a Farsi one which I believe only a small percentage of people have that. And if we don't include a Farsi/Arabic font, not only the appearance of Cypht would be kind of bad, but also some problems will occur with the selection that some parts of text in vertical will not fit so not appear like this:
without Behdad Font:
image
with Behdad Font:
image

And also the font that I found is totally free and open source and it can be used for any usage.

Revisto:
> Just noticing one more issue, the sort dropdown on message list view is now misplaced - looks like a CSS rule problem

You mean that the selection should be more right?
image
to be like this:
image

jasonmunro:
> > Just noticing one more issue, the sort dropdown on message list view is now misplaced - looks like a CSS rule problem
> 
> You mean that the selection should be more right?
> image
> to be like this:
> image

Correct, also I now have 2 sort dropdowns on some pages:
bug

Revisto:
> > > Just noticing one more issue, the sort dropdown on message list view is now misplaced - looks like a CSS rule problem
> > 
> > 
> > You mean that the selection should be more right?
> > image
> > to be like this:
> > image
> 
> Correct, also I now have 2 sort dropdowns on some pages:
> bug

I checked and one of these two does the sortation with JS‌ and the page doesn't reload, but the other one does the search by reloading the page and searching through all the email history. and also one of them has one more sort option. should I remove one of them?

jasonmunro:
> I checked and one of these two does the sortation with JS‌ and the page doesn't reload, but the other one does the search by reloading the page and searching through all the email history. and also one of them has one more sort option. should I remove one of them?

They are not both present on any page before your PR so yes :) The client side one with less options is for combined views, the one that reloads and has more opitons is for IMAP mailbox message lists.

Revisto:
> > I checked and one of these two does the sortation with JS‌ and the page doesn't reload, but the other one does the search by reloading the page and searching through all the email history. and also one of them has one more sort option. should I remove one of them?
> 
> They are not both present on any page before your PR so yes :) The client side one with less options is for combined views, the one that reloads and has more opitons is for IMAP mailbox message lists.

actually I checked and this problem occurred in [#489](https://github.com/jasonmunro/cypht/pull/489) (in site.css when "display: flex" was added into content_title).

Revisto:
> > > I checked and one of these two does the sortation with JS‌ and the page doesn't reload, but the other one does the search by reloading the page and searching through all the email history. and also one of them has one more sort option. should I remove one of them?
> > 
> > 
> > They are not both present on any page before your PR so yes :) The client side one with less options is for combined views, the one that reloads and has more opitons is for IMAP mailbox message lists.
> 
> actually I checked and this problem occurred in [#489](https://github.com/jasonmunro/cypht/pull/489) (in site.css when "display: flex" was added into content_title).

I removed combined_sort_dialog being called. with this change, the bug with multi sort option is fixed.

PR #503, opened 2021-07-20
some clean up after PR 501

Revisto:
## Pull request
Hello, this is Revisto from the Wikisuite project.
I did some clean-up around the javascript based on the things that Jason mentioned on Gitter. I shorted syntaxes using JQuery and used built-in functions for getting URLs and added a function for getting the interface direction. I also moved the page load section I added myself to an existing one.


jasonmunro:
much better, thanks for doing this!

PR #505, opened 2021-08-24
Autocrypt base implementation

henrique-borba:
# Autocrypt

> This PR is a preview and it is not ready for merging.

This PR aims to implement the sending and receiving of public PGP keys through message headers using Autocrypt Level 1 standards.

When enabling the PGP module, the user can choose to include a public key in the message header on the compose page. When receiving a message with Autocrypt headers, the user can choose to import the received key on the message preview page.

Reference: https://autocrypt.org

Screenshot_20210824_143717

Screenshot_20210824_143813

# Currently mandatory items implemented according to specification

#### The Autocrypt Header

- The addr attribute is mandatory, and contains the single recipient address this header is valid for. If this address differs from the one in the From header, the entire Autocrypt header **MUST** be treated as invalid.
- The keydata attribute is mandatory, and contains the key data for the specified addr recipient address. The value of the keydata attribute is a Base64 representation of the binary OpenPGP “Transferable Public Key”. For ease of parsing, the keydata attribute **MUST** be the last attribute in this header.

#### OpenPGP Based key data

- The keydata sent by an Autocrypt-enabled Level 1 MUA **MUST** consist of an OpenPGP “Transferable Public Key”
- These packets **MUST** be assembled in binary format (not ASCII-armored), and then base64-encoded.
- A Level 1 MUA **MUST** be capable of processing and handling Ed25519 public keys for signatures, as well as Cv25519 for encryption. 

#### Header injection in outbound mail

- This header **MUST** contain the corresponding public key material (accounts[from-addr].public_key) as the keydata attribute, and from-addr as the addr attribute. The most minimal Level 1 compliant MUA will only include these two attributes. If accounts[from-addr].prefer_encrypt is set to mutual, then the header **MUST** have a prefer-encrypt attribute with the value mutual.
- If the From address changes during message composition (e.g., if the user selects a different outbound identity), then the MUA **MUST** change the Autocrypt header accordingly.
- The MUA **MUST NOT** include more than one valid Level 1 Autocrypt header (see Updating Autocrypt Peer State).

#### Message Encryption

- An Autocrypt MUA **MUST NOT** create an Autocrypt Setup Message without explicit user interaction

# Solves

- https://github.com/jasonmunro/cypht/issues/241

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -276,9 +276,23 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
             $txt .= '<input type="hidden" class="move_to_string1" value="'.$this->trans('Move to ...').'" />';
             $txt .= '<input type="hidden" class="move_to_string2" value="'.$this->trans('Copy to ...').'" />';
             $txt .= '<input type="hidden" class="move_to_string3" value="'.$this->trans('Removed non-IMAP messages from selection. They cannot be moved or copied').'" />';
-            $txt .= '</th></tr>';
-            $txt .= '</table>';
+            $txt .= '</th></tr>';       
 
+
+            $headers = $this->get('msg_headers', array());
+            if (array_key_exists('Autocrypt', $headers)) {
+                $exploded_autocrypt_header = explode('keydata=', $headers['Autocrypt']);
+                $exploded_autocrypt_email_header = explode('addr=', $headers['Autocrypt']);
+                if (count($exploded_autocrypt_header) >= 1) {
+                    $pgp_key = end($exploded_autocrypt_header);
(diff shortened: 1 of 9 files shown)


dumblob:
Just saw notification about this PR - I'm really glad someone is working on this!

Just one thing from my side - feel free to test this with https://delta.chat/en/ as it's by far the major user of Autocrypt (though they used some **extensions** of Autocrypt and it's possible they still use some rests of those - but I believe they were all backwards compatible with pure Autocrypt).

henrique-borba:
Closed MR in favor of another more up-to-date branch and solution

marclaporte:
For the record: https://github.com/cypht-org/cypht/pull/652

PR #506, opened 2021-08-24
Profiles counter on home page

henrique-borba:
## Profiles counter on home page

This PR adds the profile counter on the main page, a shortcut for managing profiles with the main objective of reminding the user that he does not have registered profiles yet.

Screenshot_20210824_143620

## Solves

- https://github.com/jasonmunro/cypht/issues/380

### DIFF
--- a/modules/nux/modules.php
+++ b/modules/nux/modules.php
@@ -72,9 +72,6 @@ class Hm_Handler_nux_homepage_data extends Hm_Handler_Module {
         if (data_source_available($modules, 'smtp')) {
             $smtp_servers = count(Hm_SMTP_List::dump(false));
         }
-        if (data_source_available($modules, 'smtp')) {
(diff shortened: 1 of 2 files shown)


modules/nux/modules.php
```diff
@@ -71,12 +72,20 @@ public function process() {
         if (data_source_available($modules, 'smtp')) {
             $smtp_servers = count(Hm_SMTP_List::dump(false));
         }
```
jasonmunro, line 74: Looks like the smtp section got duplicated here and we should set a default empty value for $profiles if the site does not have that module set enabled.

henrique-borba:
@jasonmunro there you go

jasonmunro:
> @jasonmunro there you go

excellent thanks for doing this!

marclaporte:
This is a very very good idea!

PR #507, opened 2021-08-24
Fixed combined inbox not updating messages after combined folder removal

henrique-borba:
## Fixed combined inbox not updating messages after combined folder removal

This PR fixes a problem where, after removing a mailbox from the combined mailbox, messages from it are not removed from the combined mailbox.

## Steps to reproduce

- Open the "Everything" combined mailbox
- Remove one of your mailbox from the combined mailbox using the top right button
- Open the "Everything" combined mailbox again
- Messages from the removed mailbox are still shown

## How the fix works

- The clear_imap_page_combined_inbox function was created in the site.js of the IMAP module which allows clearing the combined mailbox cache in any other mailbox
- When update_imap_combined_source is triggered, it calls clear_imap_page_combined_inbox to clear the combined mailbox cache

### DIFF
--- a/modules/imap/site.js
+++ b/modules/imap/site.js
@@ -297,7 +297,9 @@ var imap_message_list_content = function(id, folder, hook, batch_callback) {
             if (res.auto_sent_folder) {
                 add_auto_folder(res.auto_sent_folder);
             }
+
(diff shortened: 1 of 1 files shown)


jasonmunro:
looks great thanks!

PR #510, opened 2021-09-05
fix of the overflow of array elements containing profile information.

NasserNgandu:
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


NasserNgandu:
Il y'avait un debordement des informations de profile.

jasonmunro:
I think I would rather have an ellipsis overflow with mouse over like the fix you made for the folder list, but this is still better thanks!

NasserNgandu:
Thank you!

PR #518, opened 2021-09-14
Fix draft attachment and remove legacy draft fallback

henrique-borba:
This PR removes the fallback of the legacy draft and fixes the ID issues that prevented the correct functioning of the IMAP drafts attachment.

**SOLUTION**
- Fixed the way the attachment cache IDs are updated when the draft ID is changed while editing
- When a previously saved draft with attachments is opened for editing, the attachments are downloaded to the attachments folder and added to the uploaded file cache. That way, when editing is complete, the new draft version will contain the files previously uploaded in the previous draft version.

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -58,14 +58,14 @@ class Hm_Handler_load_smtp_is_imap_draft extends Hm_Handler_Module {
                 
                 $attached_files = [];
                 $this->session->set('uploaded_files', array());
-                if (array_key_exists('subs', $msg_struct[0])) {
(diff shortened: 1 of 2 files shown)


jasonmunro:
Happy to merge when you are ready. thanks for doing this!

henrique-borba:
@jasonmunro @marclaporte I think it is ready

marclaporte:
Thanks!

Is https://github.com/jasonmunro/cypht/pull/508 still necessary?

henrique-borba:
> Thanks!
> 
> Is #508 still necessary?

https://github.com/jasonmunro/cypht/pull/508#issuecomment-927345955

PR #529, opened 2021-11-03
Cypht: Indicate which server when this error is shown: "Could not authenticate to the selected IMAP server"

jeanpaulHamuli:
Indicate which server when this error is shown: "Could not authenticate to the selected IMAP server"

### Checklist

- [X] None

### How to Test

- [ ] Add a few IMAP and SMTP accounts
- [ ] Enable google Less secure apps of one account 
- [ ] Then try to create a folder
- [ ] You will get a warning message

For more details : [Click](https://avan.tech/item57989-Cypht-Indicate-which-server-when-this-error-is-shown-Could-not-authenticate-to-the-selected-IMAP-server#threadId=52711)
### Todo
- [X] None

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -617,7 +617,7 @@ class Hm_Handler_imap_folder_expand extends Hm_Handler_Module {
                 $this->out('imap_expanded_folder_path', $path);
             }
             else {
-                Hm_Msgs::add(sprintf('ERRCould not authenticate to the selected %s server ()', $imap->server_type));
+                Hm_Msgs::add(sprintf('ERRCould not authenticate to the selected %s server (%s)', $imap->server_type, get_failed_email($imap->show_debug())));
             }
         }
     }
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -968,6 +968,23 @@ function imap_authed($imap) {
 
 /**
  * @subpackage imap/functions
+ * @param $debug_msg
+ * @return string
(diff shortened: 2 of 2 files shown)


modules/imap/handler_modules.php
```diff
@@ -633,7 +633,7 @@ public function process() {
                 $this->out('imap_expanded_folder_path', $path);
             }
             else {
-                Hm_Msgs::add(sprintf('ERRCould not authenticate to the selected %s server (%s)', $imap->server_type, $form['imap_user']));
+                Hm_Msgs::add(sprintf('ERRCould not authenticate to the selected %s server (%s)', $imap->server_type, get_failed_email($imap->show_debug())));
```
henrique-borba, line 636: This solves the problem however it creates a direct dependency between a string generated by the IMAP Debugger and the listing method, wich can eventually cause other problems in the future.

You can get the server informations directly from the ID that was sent at the beginning of the method using the module's `user_config`:

```
$this->user_config->get('imap_servers')[$form['imap_server_id']]
```

henrique-borba (changes requested):

modules/imap/handler_modules.php
jeanpaulHamuli: Thank you @henrique-borba, I solve it using `user_config` module

henrique-borba (approved):

marclaporte:
For the record, this works. (previous attempts had not worked for me)

PR #530, opened 2021-11-03
[FIX] Cypht: When we reply to an email we have sent, it proposes our address

jeanpaulHamuli:
Correct this to propose the right address when we reply to an email we have sent.
More details about the fixed issue (https://github.com/jasonmunro/cypht/issues/384)


henrique-borba:
I had that problem, I fixed in #524 by changing the email header field priority array

PR #532, opened 2021-11-08
[FIX] Cypht: Check the phone number field to make sure it contains only numbers. If not, issue a warning to the user the user once.

NasserNgandu:
…ers.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/contacts/site.js
+++ b/modules/contacts/site.js
@@ -1,8 +1,15 @@
 'use strict';
 
+var verification_number=false;
 function contact_phone_verify(){
-    var contact_phone=document.getElementById("contact_phone");
-    contact_phone.value=contact_phone.value.replace(/\D/g,'');
+    if(!verification_number){
+        var contact_phone=document.getElementById("contact_phone");
+        const regex = new RegExp('^\\d+$');
+        if(!regex.test(contact_phone.value)){
+            alert("This phone number appears to contain invalid character (s).\nIf you are sure ignore this warning and continue!");
+            verification_number=true;
+        }
+    }
 }
 
 var delete_contact = function(id, source, type) {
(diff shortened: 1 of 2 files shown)


NasserNgandu:
In the contact menu of cypth when registering a new contact there is no control on the telephone number, it can contain anything whereas normally the telephone number is composed of digits.

modules/contacts/site.js
```diff
@@ -1,5 +1,10 @@
 'use strict';
 
+function contact_phone_verify(){
+    var contact_phone=document.getElementById("contact_phone");
+    contact_phone.value=contact_phone.value.replace(/\D/g,'');
```
dumblob, line 5: Please no. People often write spaces (even UTF-8 ones, not just ASCII space character), `+`, `#`, `*`, "short pause", "long pause" (both of these represented by quite different characters on different phones...), etc. - needed for conference calls, special systems, operator messages, etc. Even overall length doesn't make much sense because many people use fixed **one-digit numbers** for certain important contacts (e.g. inside a small company).

So, I for myself would be very afraid to impose basically any restrictions here without knowing the context better. And Cypht does **not** have any context IMHO, so rather no restrictions :wink:.

Just my 2 cents.

marclaporte: Also, there are numbers like 1 877 Wiki WEB

marclaporte:
@NasserNgandu: If we want to do this, it should be a soft warning like "This data seems to contain invalid characters. Are you sure?" and not prevent the user.

Also, as @dumblob points out, it should accept special characters.

NasserNgandu:
I modify the code then, to warn the user once, if he wants he can ignore message and type whatever he wants.

NasserNgandu:
> I modify the code then, to warn the user once, if he wants he can ignore message and type whatever he wants.

I made a change to the commit.

PR #533, opened 2021-11-11
[FIX] Cypht reply-all could have a better UX

jeanpaulHamuli:
Issue fixed : Cypht reply-all could have a better UX
Decription of issue : https://avan.tech/item57670-Cypht-reply-all-could-have-a-better-UX

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -239,12 +239,7 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                 '<a class="hlink small_headers" style="display: none;" href="#">'.$this->trans('Small headers').'</a>';
             if (!isset($headers['Flags']) || !stristr($headers['Flags'], 'draft')) {
                 $txt .= ' | <a class="reply_link hlink" href="?page=compose&amp;reply=1'.$reply_args.'">'.$this->trans('Reply').'</a>';
-                if ($reply_all || $size > 1) {
-                    $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
-                }
-                else {
-                    $txt .= ' | <a class="reply_all_link hlink disabled_link">'.$this->trans('Reply-all').'</a>';
-                }
+                $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
                 $txt .= ' | <a class="forward_link hlink" href="?page=compose&amp;forward=1'.$reply_args.'">'.$this->trans('Forward').'</a>';
             }
             if ($msg_part === '0') {


modules/imap/output_modules.php
```diff
@@ -243,7 +243,7 @@ protected function output() {
                     $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
                 }
                 else {
-                    $txt .= ' | <a class="reply_all_link hlink disabled_link">'.$this->trans('Reply-all').'</a>';
+                    $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
```
henrique-borba, line 246: It looks good, have you checked if the code below can add the `disabled_link` class again without you wanting to?

https://github.com/jasonmunro/cypht/blob/9a61e208fbd462c02dc561bb2db6625ea4e7d353/modules/imap/site.js#L721-L724

henrique-borba (changes requested):

modules/imap/output_modules.php
jeanpaulHamuli: It can only add the `disabled_link` when the page has not yet loaded

henrique-borba: Good, I just realized that the `else` below is unnecessary:

https://github.com/jasonmunro/cypht/blob/8aadcbfe1cde41043c571af96992ba7f62670fd7/modules/imap/output_modules.php#L242-L247

The code is the same for both conditions

henrique-borba (changes requested):

modules/imap/output_modules.php
jeanpaulHamuli: I didn't pay attention to that, let me remove one and the condition as it is no longer in use

modules/imap/output_modules.php
```diff
@@ -239,12 +239,7 @@ protected function output() {
                 '<a class="hlink small_headers" style="display: none;" href="#">'.$this->trans('Small headers').'</a>';
             if (!isset($headers['Flags']) || !stristr($headers['Flags'], 'draft')) {
                 $txt .= ' | <a class="reply_link hlink" href="?page=compose&amp;reply=1'.$reply_args.'">'.$this->trans('Reply').'</a>';
-                if ($reply_all || $size > 1) {
-                    $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
-                }
-                else {
-                    $txt .= ' | <a class="reply_all_link hlink disabled_link">'.$this->trans('Reply-all').'</a>';
-                }
+                $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
```
henrique-borba, line 242: We still need to check if there is more than one sender to reply

```php
if ($reply_all || $size > 1) {
   $txt .= ' | <a class="reply_all_link hlink" href="?page=compose&amp;reply_all=1'.$reply_args.'">'.$this->trans('Reply-all').'</a>';
}
```

jeanpaulHamuli: If we still check without `else`, the **reply all** button will disappear when there is less than one message. 

It should be like that ?

henrique-borba: You're right, the previous pattern was to show so let's keep it that way. Thanks Jean

henrique-borba: You're right, the previous behaviour was to show it anyway so let's keep it that way. Thanks Jean

henrique-borba (approved):

modules/imap/output_modules.php
jeanpaulHamuli: Thank you too Henrique

PR #534, opened 2021-11-14
[FIX] Cypht webmail view on a folder in mobile mode: Should have a search box

jeanpaulHamuli:
## Pullrequest

Cypht webmail view on a folder in mobile mode: Should have a search box 

### Issues

https://avan.tech/item54308-Cypht-webmail-view-on-a-folder-in-mobile-mode-Should-have-a-search-box

### Checklist

- [X] None

### How2Test

- [ ] Go to Email->All
- [ ] In the search box write something then click Enter

### Todo

- [X] None


jasonmunro:
The problem here is the search box overlaps on top of the message list title. Depending on the width of the mobile device it can completely cover it. I specifically removed search (and sort) from mobile because of this reason. If we want to add this back we need to figure out a better layout, either a second row in the heading area or a search icon that makes the input box visible. We should fix the sort dialog as well as it has the same issue since be re-enabled on mobile.

marclaporte:
ok, let's go for an improved layout as search is so critical.

jeanpaulHamuli:
Understand, let's improve it

jeanpaulHamuli:
This is what I suggest, adding a second row was not good. How do you find the @jasonmunro idea?
cyphtImg

jasonmunro:
Looks great, thanks for reworking this!

PR #536, opened 2021-11-16
[FIX] Cypht: Check the phone number field to make sure it only contains numbers and a few allowed characters. If not, send a warning to the user only once.

NasserNgandu:
…ns numbers and a few allowed characters. If not, send a warning to the user only once.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/contacts/site.js
+++ b/modules/contacts/site.js
@@ -1,5 +1,21 @@
 'use strict';
 
+var verification_number=false;
+function contact_phone_verify(){
+    if(!verification_number){
+        var contact_phone=document.getElementById("contact_phone");
+        const regex_number = new RegExp('^\\d+$');
+        const allowed_characters = ['+','-','(',')'];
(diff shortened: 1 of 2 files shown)


jasonmunro:
Just a few changes and I am happy to merge this:
- remove the onkeyup and onchange handlers from the html and add them to the javascript here: https://github.com/jasonmunro/cypht/blob/master/modules/contacts/site.js#L167
- use cachjs rather than getElementById
- dont use a global variable for state, instead inside the function remove the event handler after it is executed so it does not execute after the first time

Happy to explain in more detail if needed.
thanks!

NasserNgandu:
Thank you ! I take note of remarks. I fix it.

NasserNgandu:
I just made another merge request, following your comments: https://github.com/jasonmunro/cypht/pull/546.

PR #541, opened 2021-12-22
Cypht: addition of the selection with ctrl click and of the multiple selection with shift clicks of the mails while clicking on all the entry.

NasserNgandu:
…selection with shift clicks of the mails while clicking on all the entry.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/core/message_list_functions.php
+++ b/modules/core/message_list_functions.php
@@ -178,7 +178,7 @@ function human_readable_interval($date_str) {
  */
 if (!hm_exists('message_list_row')) {
 function message_list_row($values, $id, $style, $output_mod, $row_class='') {
-    $res = '<tr class="'.$output_mod->html_safe($id);
+    $res = '<tr onclick="set_row_mail_callback(this)" class="'.$output_mod->html_safe($id);
     if ($row_class) {
         $res .= ' '.$output_mod->html_safe($row_class);
     }
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -1,5 +1,71 @@
 'use strict';
 
+document.addEventListener('keydown', checkKeyDown);
+document.addEventListener('keyup', checkKeyUp);
+var ctrlPressed = false;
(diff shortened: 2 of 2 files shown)


NasserNgandu:
As it was asked on this task: https://avan.tech/item38529-Cypht-SHIFT-click-and-CTRL-click-should-permit-to-select-several-entries-like-in-Roundcube, to make a selection of the mails while clicking on all the entry like on Rundcube.
I had suggested that the selection of an email be done with ctrl click and the multiple selection of emails is done with shift click. This also made it possible to bypass the links zone which allows to open the mails, to make the selection possible by clicking on the whole entry.

marclaporte:
Here is context:
https://github.com/jasonmunro/cypht/issues/362

jasonmunro:
I appreciate your work on this but there are several issues with this implementation:
- inline onclick event triggers should not be used, I much prefer for events to be managed entirely in javascript
- the keyup and keydown events are globally active for every page meaning they fire on every keystroke
- this uses global variables in javascript (we use some globally scoped variables in javascript however only where we need to)
- this seems to ignore that we already have shift-click support for the checkboxes which can be re-used for row level clicks
- we have jquery like support (cashjs) so we should use that where possible instead of vanilla javascript (not a big deal but typically the cashjs version of an action is less code/cleaner)

Implementing this with all of the above considerations is tricky so I took a stab at it and pushed a change that:
- adds shift click multi select to any click on the row
- adds control click select/unselect to any click on the row
- loads the message on a click with no modifiers
- maintains the checkbox behavior without conflicts from the new shift/ctrl click code
- integrates with the existing code base and does not use onclick, global variables, or global events

Again thank you for working on this. Here is the change I pushed: https://github.com/jasonmunro/cypht/commit/a4a9f6f610ac978f2d70986bfcd444fc5be8edf4

PR #543, opened 2021-12-30
Adding drag and drop of mails using the SortableJs library

NasserNgandu:
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/core/Sortable.js
+++ b/modules/core/Sortable.js
@@ -1,0 +1,3788 @@
+/**!
+ * Sortable 1.14.0
+ * @author	RubaXa   <trash@rubaxa.org>
+ * @author	owenm    <owen23355@gmail.com>
+ * @license MIT
+ */
+(function (global, factory) {
+  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
+  typeof define === 'function' && define.amd ? define(factory) :
+  (global = global || self, global.Sortable = factory());
+}(this, (function () { 'use strict';
+
+  function ownKeys(object, enumerableOnly) {
+    var keys = Object.keys(object);
+
+    if (Object.getOwnPropertySymbols) {
+      var symbols = Object.getOwnPropertySymbols(object);
+
+      if (enumerableOnly) {
+        symbols = symbols.filter(function (sym) {
+          return Object.getOwnPropertyDescriptor(object, sym).enumerable;
+        });
+      }
+
+      keys.push.apply(keys, symbols);
+    }
+
+    return keys;
+  }
+
+  function _objectSpread2(target) {
+    for (var i = 1; i < arguments.length; i++) {
+      var source = arguments[i] != null ? arguments[i] : {};
+
+      if (i % 2) {
+        ownKeys(Object(source), true).forEach(function (key) {
+          _defineProperty(target, key, source[key]);
+        });
+      } else if (Object.getOwnPropertyDescriptors) {
+        Object.defineProperties(target, Object.getOwnPropertyDescriptors(source));
+      } else {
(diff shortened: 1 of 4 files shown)


NasserNgandu:
Here is the link of the task: https://avan.tech/item44205-Cypht-Add-drag-and-drop.

dumblob:
Thanks for taking stab at this!

Before anyone will dive into to the source I wonder whether ~3.7k SLOC library is worth including?

Are there other **real** (not just hypothetical, not "for future") benefits of the (quite huge I'd say) JS library for Cypht?

marclaporte:
Related feature request: https://github.com/jasonmunro/cypht/issues/216

marclaporte:
https://github.com/SortableJS/Sortable/blob/master/Sortable.js is bigger than https://github.com/bevacqua/dragula/blob/master/dist/dragula.js

Last commit in 2020-09  (15 months is not a disqualifier but a concern)
https://github.com/bevacqua/dragula/commits/master


Another topic: should we use minified version?
https://github.com/SortableJS/Sortable/blob/master/Sortable.min.js

marclaporte:
@dumblob There is more analysis at https://github.com/jasonmunro/cypht/issues/216

I feel the decision is "good enough" and not worth investing more time. But if you or someone come with a better option, we can look into swapping.

Thanks!

dumblob:
@marclaporte I see - thanks for the analysis of "liveness" of the biggest contenders.

On the other hand I'd still prefer evaluating the following 3 criteria (in this order) in addition to project liveness:

1. the user feel on a smartphone/tablet of the given library (yes, touch support is different and some implementations really feel very weird & unintuitive and some do feel much better)
2. the simplicity and "smartness" of the API (i.e. how much it'll influence Cypht future design & decisions, how easy is it to learn, how orthogonal to other Cypht JS code it is, etc.)
3. whether it dictates anything peculiar about the underlying html structure or not (if it does, then it goes the "React" route of constrained bulky widgets which is something Cypht IMHO doesn't want to go)

On the other hand, if the library you want to choose **now** will not influence Cypht design and APIs and thus will be easily exchangeable, then I'd take the first such library which also suits your devs as that's what counts the most at the beginning. We can later see if we could optimize (do better) in the sense of point (1) above or at least in the sense of the library size.

marclaporte:
With the currently proposed integration, 4 lines of code are modified. So the lib is easily interchangeable later. I do expect us to add drag and drop in a few other places like folder management but I don't expect it to influence Cypht.

I tested https://sortablejs.github.io/Sortable/ in touch mode, and it works well. Only issue I noticed in the "Handle" mode: the handle is small and not easy to catch.

SortableJS has MultiDrag, which makes a lot of sense for email:
https://github.com/SortableJS/Sortable/wiki/Dragging-Multiple-Items-in-Sortable

jasonmunro:
I'm a bit confused by this PR - What exactly is the point of this? To re-order the list on the client side? The order is not sticky (and will most likely break new message arrival placement in the list) so I really don't understand the use case. @marclaporte can you help me out here?

marclaporte:
This is the goal: https://github.com/jasonmunro/cypht/issues/216


So pick 3 messages and drag and drop to a folder.

NasserNgandu:
I'm sorry, I misunderstood the task. The code just changes the order of the emails. I'll fix that ASAP.

jasonmunro:
> I'm sorry, I misunderstood the task. The code just changes the order of the emails. I'll fix that ASAP.

I think this is going to be _maybe_ a bit tricky. Let me know if you get stuck on anything I'm happy to help!

PR #546, opened 2022-01-19
[FIX] Cypht: Check the phone number field to make sure it only contains numbers and a few allowed characters. If not, send a warning to the user only once.

NasserNgandu:
…ns numbers and a few allowed characters. If not, send a warning to the user only once.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/contacts/site.js
+++ b/modules/contacts/site.js
@@ -179,6 +179,18 @@ if (hm_page_name() == 'contacts') {
     $('.server_title').on("click", function() {
         $(this).next().toggle();
     });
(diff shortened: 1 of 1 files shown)


jasonmunro:
Awesome work, thank you for updating the code on this!

jasonmunro:
FYI I changed the js alert() to our Hm_Notices system so the warning comes up like other Cypht messages.

NasserNgandu:
Thank you !

PR #547, opened 2022-01-20
Cypht "reply to all": Should be available if there are more than one email in from, to and cc

jeanpaulHamuli:
Cypht "reply to all": Should be available if there are more than one email in from, to and cc

## Pullrequest
Sending message to multiple emails, the reply all button will appear if there is only one email, the button is hidden.
link to the task: https://avan.tech/item72360-Cypht-reply-to-all-Should-be-available-if-there-are-more-than-one-email-in-from-to-and-cc 

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -236,7 +236,12 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                 '<a class="hlink small_headers" style="display: none;" href="#">'.$this->trans('Small headers').'</a>';
             if (!isset($headers['Flags']) || !stristr($headers['Flags'], 'draft')) {
                 $txt .= ' | <a class="reply_link hlink" href="?page=compose&amp;reply=1'.$reply_args.'">'.$this->trans('Reply').'</a>';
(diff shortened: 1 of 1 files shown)


jasonmunro:
This code is fine but I don't want to hide the reply all button, disabling it is fine but hiding changes the layout of the links which I prefer to maintain. Could you please change this to disable the reply all link instead of hiding it?

marclaporte:
@jeanpaulHamuli GitHub says "This branch cannot be rebased due to conflicts"

dumblob:
We already discussed this some years ago here in this tracker and I thought we all settled at the agreement that *individual* interface widgets shall **not** change their position nor dimensions nor visibility depending on the content.

What is the motivation to break this rule No. 1 in UX?

marclaporte:
@jeanpaulHamuli What do you think?

jeanpaulHamuli:
We have a lot of ideas on this, I remember the last time we obeyed the rule, don't change anything, just leave the button there in any situation [[533]](https://github.com/jasonmunro/cypht/pull/533), that was good too, but for me, to make this more user friendly what we can do is check if there is only one email from, to and cc we just disable the button.

jasonmunro:
looks good thanks!

PR #558, opened 2022-03-03
[FIX] Fix the forwarding a message which was broken by calling the function which is no longer defined

### LINKED ISSUE #557: Forwarding a message is broken
When trying to forward a message the code is calling attach_file() which is no longer defined. Looks like this bit was missed when we changed how attachments are handled. Easy to trigger, just try forwarding a message.

```[01-Mar-2022 19:22:01 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function attach_file() in /home/jason/cypht/modules/imap/handler_modules.php:58
Stack trace:
#0 /home/jason/cypht/lib/modules_exec.php(175): Hm_Handler_imap_forward_attachments->process()
#1 /home/jason/cypht/lib/modules_exec.php(153): Hm_Module_Exec->run_handler_module()
#2 /home/jason/cypht/lib/dispatch.php(220): Hm_Module_Exec->run_handler_modules()
#3 /home/jason/cypht/lib/dispatch.php(190): Hm_Dispatch->process_request()
#4 /home/jason/cypht/index.php(50): Hm_Dispatch->__construct()
#5 {main}
  thrown in /home/jason/cypht/modules/imap/handler_modules.php on line 58```

### PR DESCRIPTION
**Pull request by jeanpaulHamuli**
## Pullrequest

Remove the call of function `attach_file()` which is no longer defined.   
### Issues

- fix #557

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -107,6 +107,52 @@ class Hm_Handler_load_smtp_is_imap_draft extends Hm_Handler_Module {
 /**
  * @subpackage smtp/handler
  */
+class Hm_Handler_load_smtp_is_imap_forward extends Hm_Handler_Module
+{
+    public function process() {
+        if (!$this->module_is_supported('imap')) {
+            return;
+        }
+        
+        if (array_key_exists('forward', $this->request->get)) {
+            $path = explode('_', $this->request->get['list_path']);
+            $imap = Hm_IMAP_List::connect($path[1]);
+            if ($imap->select_mailbox(hex2bin($path[2]))) {
+                $msg_struct = $imap->get_message_structure($this->request->get['uid']);
+                list($part, $msg_text) = $imap->get_first_message_part($this->request->get['uid'], 'text', 'plain', $msg_struct);
+                $msg_header = $imap->get_message_headers($this->request->get['uid']);
+                if (!array_key_exists('From', $msg_header) || count($msg_header) == 0) {
+                    return;
+                }
+
+                # Attachment Download
+                $attached_files = [];
+                $this->session->set('uploaded_files', array());
+                if (array_key_exists(0, $msg_struct) && array_key_exists('subs', $msg_struct[0])) {
+                    foreach ($msg_struct[0]['subs'] as $ind => $sub) {
+                        if ($ind != '0.1') {
+                            $new_attachment['basename'] = $sub['description'];
+                            $new_attachment['name'] = $sub['description'];
+                            $new_attachment['size'] = $sub['size'];
+                            $new_attachment['type'] = $sub['type'];
+                            $file_path = $this->config->get('attachment_dir') . DIRECTORY_SEPARATOR . $new_attachment['name'];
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jasonmunro** [comment]
This "fixes" the bug but forwarding now does not include attachments in the original message.

**Comment by jeanpaulHamuli** [comment]
I see, means that instead of removing the call to that function, I should redefine it?

**Comment by jasonmunro** [comment]
Tested and there were no errors but it did not work. The original message should appear as an attachment on the forwarded version on the compose page. When I forward a message with this PR there is no original message attachment

**Comment by jeanpaulHamuli** [comment]
Alright, let me check again why

**Comment by jeanpaulHamuli** [comment]
Please @jasonmunro, is that the original message I circled there ?
OrMsg

**Comment by jasonmunro** [comment]
The entire original message should be an attachment on the forwarded message. You should also have the primary message body in the compose page as you circled, but the entire message should also be an attachment.

For example, what if I forward a message that has an image attached to it? I would expect the forwarded message to include that image. Since E-mail can have complicated formats with many attachments, the easiest way to handle forwarding is to capture the entire original message and attach it to the forwarded message.

**Comment by jeanpaulHamuli** [comment]
I think I now understand the need.
Just a clarification, before it was working as you explain or is an improvement we need to implement?
I'm trying to go behind some commits, to see if I'll see any changes but nothing so far.

**Comment by marclaporte** [comment]
Yes @jeanpaulHamuli it was working like that before. Just try Cypht code from a few months ago.

**Comment by henrique-borba** [comment]
The implementation is similar to editing drafts. Basically you should download the attachments to the attachments folder and set the uploaded_files session variable. Below you can see how it was implemented for the drafts:

https://github.com/jasonmunro/cypht/blob/56679db25d696eecab8e6441f733c80a48d6c3c2/modules/smtp/modules.php#L59-L80

**Comment by jeanpaulHamuli** [comment]
This was also solved for me, I just updated this PR

File: modules/smtp/modules.php [hunk 1]
```diff
@@ -1651,6 +1698,33 @@ function get_draft($id, $session) {
     return false;
 }}
 
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('attach_file')) {
+function attach_file($content, $file, $filepath, $draft_id, $mod, $imap_draft=false, $atts=array()) {
```
jasonmunro, line 1705: This function is unused I think so should be removed from the PR. Thanks!

modules/smtp/modules.php
```diff
@@ -104,6 +104,53 @@ public function process() {
...
 }
 
+/**
+ * @subpackage smtp/handler
+ */
+class Hm_Handler_load_smtp_is_imap_forward extends Hm_Handler_Module
+{
+    public function process() {
+        if (!$this->module_is_supported('imap')) {
+            Hm_Msgs::add('ERRIMAP module unavailable.');
```
jasonmunro, line 114: You could forward a POP3 message so while I don't care about attachments in that case we probably don't need this user facing warning, just the early exit.

jeanpaulHamuli: I remove the warning message, @jasonmunro can you review please?

PR #559, opened 2022-03-09
Fix bug in PHP 8.1 when require lib/request.php due to $GLOBALS modif…

xorti:
…ication changes

## Pullrequest

This pull request fixes an issue in PHP 8.1 that was failing with **Fatal Error** when requiring the file `lib/request.php` due to `$GLOBALS` modification.

```
Fatal error: $GLOBALS can only be modified using the $GLOBALS[$name] = $value syntax in /var/www/html/vendor/jason-munro/cypht/lib/request.php on line 153
```

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] Run the following code with PHP8.1 to see the issue: `php -r "require 'lib/request.php';"`

### Todo

- [X] None

### DIFF
--- a/lib/request.php
+++ b/lib/request.php
@@ -150,7 +150,10 @@ class Hm_Request {
         $_FILES = array();
         $_REQUEST = array();
         $_ENV = array();
-        $GLOBALS = array();
+        
+        foreach (array_keys($GLOBALS) as $key) {
+            unset($GLOBALS[$key]);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


jasonmunro:
@xorti thanks for doing this, I will merge it but I'm confused as to why the version_compare() check in that function isn't working (not working with a direct require from the CLI for me either). Weird!

PR #569, opened 2022-05-25
Server side filters with Sieve

henrique-borba:
This PR adds a new module that allows the creation, construction and management of server side filters using Sieve and the ManageSieve protocol, enabling automatic filters and building the foundations for intelligent filter integration based on artificial intelligence executed directly by the Sieve service (see https://doc.dovecot.org/configuration_manual/sieve/plugins/extprograms/)

## Screenshots

Screenshot_20220525_164840
Screenshot_20220525_165007
Screenshot_20220527_103534
Screenshot_20220527_103631
Screenshot_20220527_103704

## Dependencies

Thinking of the native and minimalist code followed by Cypht, this change requires few external dependencies. The ManageSieve communication library was custom made to our needs and allows us to add capabilities more conveniently and quickly.

- **henrique-borba/php-sieve-manager** -  Native implementation of the ManageSieve protocol. Does the heavy lifting and transforms logical frameworks to generate valid Sieve scripts https://github.com/henrique-borba/php-sieve-manager
- **ext-socket** - Already used

## Affected Modules

- **IMAP** - The settings for each mailbox may contain a Sieve configuration, so the IMAP module has been changed to allow the user to configure connection and authentication information.


dumblob:
How did you solve compatibility with other systems (especially those proprietary ones ran by huge cloud email providers)?

How does your Sieve lib behave in case it gets some slightly modified Sieve script from major email providers (I do not know whether the UI in this PR allows this, but I suppose you want to support pure text input)?

henrique-borba:
> How did you solve compatibility with other systems (especially those proprietary ones ran by huge cloud email providers)?
> 
> How does your Sieve lib behave in case it gets some slightly modified Sieve script from major email providers (I do not know whether the UI in this PR allows this, but I suppose you want to support pure text input)?

There are two options for adding a filter, one of them "Add Script" allows you to add a Sieve script  using a textarea input without restrictions. The module only manages scripts and filters created through Cypht and the user could add their old filters manually (copy and paste).

This is necessary because as only one script is activated in Sieve per account, the module needs to create a script that includes all others in the user-defined priority order. 

btw, I will add more screenshots

marclaporte:
We have a test server up at https://sieve.cypht.dev1.evoludata.com/?page=sievefilters

To help with testing, please contact me, and I will provide you with a login.

marclaporte:
Next todos (for July 7th)
- Mailbox list for copy and move actions
- Change the name of copy action to move
- Add the copy action
- Check the option of bounce emails
- Block List (ban list)
- Tweak for mobile

modules/imap/handler_modules.php
```diff
@@ -1692,7 +1692,12 @@ public function process() {
                     $this->out('msg_text', $msg_text);
                     $this->out('msg_download_args', sprintf("page=message&amp;uid=%s&amp;list_path=imap_%s_%s&amp;imap_download_message=1", $form['imap_msg_uid'], $form['imap_server_id'], $form['folder']));
                     $this->out('msg_show_args', sprintf("page=message&amp;uid=%s&amp;list_path=imap_%s_%s&amp;imap_show_message=1", $form['imap_msg_uid'], $form['imap_server_id'], $form['folder']));
-                    $this->out('sieve_filters_enabled', $this->module_is_supported('sievefilters'));
+
+                    $server = $this->user_config->get('imap_servers')[$this->request->post['imap_server_id']];
+
+                    if (array_key_exists('sieve_config_host', $server)) {
+                        $this->out('sieve_filters_enabled', $this->module_is_supported('sievefilters'));
+                    }
```
dumblob, line 1700: Hm, previously it was always shown correctly describing the boolean choice. But now, there is nothing shown if I am not mistaken. That is bad UX. It shall be **always shown** but greyed-out/deselected (meaning "unavailable") or shown normally (meaning "available").

marclaporte:
Next steps (for 2022-07-14)
1- In filters, add "To or CC"

2- In filters, add custom headers

3- filters: if easy, add Regex

4- addflag, removeflag (as part of RFC)

5- In block list, a new option next to each email: "Convert to blocking the whole domain"

6- In block list, make it more obvious the action of unblocking (maybe more like a delete button)

7- Remove Sieve username and password since it will use the main IMAP credentials.

8- Investigate how does Roundcube do "Reply with message"
Message body (vacation reason):
Message subject:
Reply sender address:
My e-mail addresses:
Fill with all my addresses
How often send messages: x days

Is this the vacation extension of the spec?

9- Investigate what "Set variable". Roundcube has:
Variable name:
Variable value:
Modifiers:
lower-case
upper-case
first character lower-case
first character upper-case
quote special characters
length

10- Tweak for mobile

marclaporte:
@jasonmunro Please merge this when you are ready. No need to wait for these other todos (If you are faster, they will come in distinct merge requests as soon as ready):

1. Sieve accounts: when adding accounts, propose default account to be the same as IMAP (In JS) which is likely to be the case 99% of the time (Henrique)
2. Filters: Add a spinner when waiting for mailbox (because it can task a few seconds when there are lots of folders, and we don't want user to think it failed) (Henrique)
3. Make Sieve lib more findable on Packagist and Github so keywords/tags + add a readme (Henrique)
4. Make mobile version look OK (Elvis)

marclaporte:
@jasonmunro This is ready for you. Henrique and I don't have anymore todos on this. Elvis will make it mobile-friendly in a future merge request.

Thanks!

jasonmunro:
Only thing I notice on a quick review is that in debug mode we are including both tingle and litegraph javascript, but in the config_gen.php file only tingle is added. Is that correct @henrique-borba ?

marclaporte:
@jasonmunro [Litegraph](https://github.com/jagenjo/litegraph.js) has been removed. At the beginning of the project, we were exploring to use Litegraph as a GUI for filters, but in the end, we didn't implement that idea.

jasonmunro:
Looks like I need to setup Sieve so I can use this too! Thanks for all the work on this!

PR #570, opened 2022-05-27
Fix bug #439: archive action does not work in list mode

Steven--:
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None


jeanpaulHamuli:
Looks good, what do you say Jason?

PR #574, opened 2022-06-28
Update bacon/bacon-qr-code to also support v2

xorti:
- Also remove composer.lock

## Pullrequest
The library `endroid/qr-code` had a dependency of `bacon/bacon-qr-code` between `1.0.0` and `^2.0.0` depending on the installed version of the  `endroid/qr-code` library.

Cypht can still use any version of the `bacon/bacon-qr-code` since the class names are exactly the same.

---
Note: The `composer.lock` was not updated with the changes in composer.json, but also it should not be versioning, allowing composer to manage the packages depending on the software dependencies.


marclaporte:
@jasonmunro I merged this in because it's causing issues with Tiki's inclusion of Cypht.

We use
`"bacon/bacon-qr-code": "^2.0.0",`
here: https://gitlab.com/tikiwiki/tiki/-/blob/master/vendor_bundled/composer.json#L13

PR #584, opened 2022-08-04
[FIX] Increase z-index to make move to dropdown visible

josaphatim:
## Pullrequest

**Move to dropdown** displays under messages text and users are unable to move folders.

Related: https://avan.tech/item81392

### Issues
- [X] None


jasonmunro:
looks good, thanks!

PR #586, opened 2022-08-04
[ENH] Nux: set special folders

josaphatim:
## Pullrequest
Related: https://avan.tech/item60588

### Issues
- fixes https://github.com/jasonmunro/cypht/issues/397

### DIFF
--- a/modules/imap_folders/modules.php
+++ b/modules/imap_folders/modules.php
@@ -125,6 +125,40 @@ class Hm_Handler_process_special_folder extends Hm_Handler_Module {
 /**
  * @subpackage imap_folders/handler
  */
+class Hm_Handler_process_accept_special_folders extends Hm_Handler_Module {
+    public function process() {
+        list($success, $form) = $this->process_form(array('imap_server_id', 'imap_service_name'));
+        if ($success) {            
+            $cache = Hm_IMAP_List::get_cache($this->cache, $form['imap_server_id']);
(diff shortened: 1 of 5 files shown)


modules/imap_folders/modules.php
```diff
@@ -122,6 +122,40 @@ public function process() {
...
+
+            if (!is_object($imap) || $imap->get_state() != 'authenticated') {
+                Hm_Msgs('ERRUnable to connect to the selected IMAP server');
+                return;
+            }
+
+            $specials = $this->user_config->get('special_imap_folders', array());
+            if ($exposed = $imap->get_special_use_mailboxes()) {
+                $specials[$form['imap_server_id']] = array('sent' => $exposed['sent'], 'draft' => '', 'trash' => $exposed['trash'], 'archive' => '');
+            } else if ($form['imap_service_name'] == 'gandi') {
```
marclaporte, line 143: Would be cleaner to store here: https://github.com/jasonmunro/cypht/blob/master/modules/nux/services.php

josaphatim: I thought that as imap folder is in charge of setting special folders it could be better to put the code there.

marclaporte: ok, let's see what @jasonmunro thinks.

jasonmunro: this looks fine to me, thanks!

PR #587, opened 2022-08-06
Enabled Copy/Move when picking an email in message table

josaphatim:
## Pullrequest
This consists of allowing user move/copy when he selects one or more emails in message table.

### Issues
-- relates https://avan.tech/item81123


jasonmunro:
Looks good, thanks!

PR #588, opened 2022-08-07
Made next & previous emails optional in combined view

josaphatim:
## Pullrequest
I created show previous & next emails setting to let user choose either he wants to see those links or not.

### Issues
- fixes https://github.com/jasonmunro/cypht/issues/402
- relates https://avan.tech/item60586

### DIFF
--- a/hm3.sample.ini
+++ b/hm3.sample.ini
@@ -771,6 +771,10 @@ modules[]=keyboard_shortcuts
 ; Defaults to full structure
 ; default_setting_simple_msg_parts=false
 
+; Next and Previous emails on the message view page
+; Defaults to full structure
+; default_setting_pagination_links=true
+
 ; Show icons for each IMAP message part type
 ; Defaults to true
 ; default_setting_msg_part_icons=true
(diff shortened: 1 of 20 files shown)


jasonmunro:
This looks great, and thanks for updating the language files. Only one issue I ran into. When I first login after merging this code, the new feature is not checked on the Settings -> Site page, but the links are showing. I think the logic in Hm_Output_imap_pagination_links needs to be updated to check the box if the setting does not yet exist in the user settings. Once I check the box, then uncheck it everything works as expected. Thanks!

josaphatim:
> This looks great, and thanks for updating the language files. Only one issue I ran into. When I first login after merging this code, the new feature is not checked on the Settings -> Site page, but the links are showing. I think the logic in Hm_Output_imap_pagination_links needs to be updated to check the box if the setting does not yet exist in the user settings. Once I check the box, then uncheck it everything works as expected. Thanks!

Thanks. I added one more commit please check.

jasonmunro:
Looks good thanks for the update!

PR #589, opened 2022-08-09
Added archive to original folder option

josaphatim:
## Pullrequest
I created an option **Archive to original folder** which will keep original folder in archive when checked otherwise all messages will be in archive folder.

### Issues
- fixes https://github.com/jasonmunro/cypht/issues/228#issuecomment-540259005
- relates https://avan.tech/item79532


jasonmunro:
Just one question - does this still work if the target folder already exists? I don't see a check for that but maybe we don't need it? Thanks!

josaphatim:
> folder

I'm working on it. Thanks.

josaphatim:
> Just one question - does this still work if the target folder already exists? I don't see a check for that but maybe we don't need it? Thanks!

Hello. I have added another commit please check.

marclaporte:
@josaphatim I get This branch has conflicts that must be resolved

marclaporte:
@josaphatim I get "This branch has conflicts that must be resolved"

Details:

<<<<<<< folder-pattern
    'Archive to the original folder' => false
=======
    'Show next & previous emails when reading a message' => false
>>>>>>> master

language/az.php
```diff
@@ -584,7 +584,8 @@
     'Search Results' => false,
     'feed item' => false,
     'Allow handling of mailto links' => false,
-    'Show folders' => false
+    'Show folders' => false,
```
marclaporte, line 587: @jasonmunro Do committers need to do this? Seems we shouldn't worry about this until a translator wants to work on a specific language.

In Tiki, we do like this: https://gitlab.com/tikiwiki/tiki/-/blob/master/lib/core/Tiki/Command/GetStringsCommand.php

josaphatim:
> @josaphatim I get "This branch has conflicts that must be resolved"
> 
> Details:
> 
> ```
> <<<<<<< folder-pattern
>     'Archive to the original folder' => false
> =======
>     'Show next & previous emails when reading a message' => false
> >>>>>>> master
> ```

Think it is now good

language/az.php
jasonmunro: Actually I do like when this done, it makes it easier for a translator to be able to convert all the strings from a single source.

PR #590, opened 2022-08-10
Fix unremovable text preferences/settings

josaphatim:
## Pullrequest
Updated code to support emptying settings with string values.

### Issues
- relates https://avan.tech/item81566

### DIFF
--- a/lib/module.php
+++ b/lib/module.php
@@ -394,7 +394,7 @@ abstract class Hm_Handler_Module {
     private function check_field($val) {
         switch (true) {
             case is_array($val):
-            case trim((string) $val) !== '':
+            case is_string($val):
             case $val === '0':
             case $val === 0:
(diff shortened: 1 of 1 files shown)


jasonmunro:
Sorry, I merge this thinking it looked innocent but it is breaking unit tests. I honestly don't know exactly what this change was supposed to fix but I think we need to find another solution. Happy to help if you can explain what the specific issue is!

marclaporte:
It's possible to edit values, but not remove (empty) them completely. When you try, they just come back. Seen on "Gmail Client ID"

jasonmunro:
Thanks for the clarification, pushed a fix to master

marclaporte:
Sorry for breaking the tests. I added https://github.com/jasonmunro/cypht/issues/596

PR #591, opened 2022-08-11
Stop reuse of server ids

henrique-borba:
If server 1 is removed, index 1 must not be used anymore. Reusing entities ids can cause many problems.

https://github.com/jasonmunro/cypht/issues/573

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -1342,9 +1342,12 @@ class Hm_Handler_load_imap_servers_from_config extends Hm_Handler_Module {
         $added = false;
         $updated = false;
         $new_servers = array();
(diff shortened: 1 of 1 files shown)


jasonmunro:
Thanks for this, merging, however we should probably consider doing this for other sets of servers like SMTP etc?

marclaporte:
@jasonmunro @henrique-borba Makes sense to me.

marclaporte:
@jasonmunro Henrique is working to move everything to stable IDs: 
https://github.com/jasonmunro/cypht/pull/625

marclaporte:
Major cleanup for Cypht 2.0: https://github.com/cypht-org/cypht/pull/625

PR #595, opened 2022-08-12
Delete attachments but keep email

NasserNgandu:
Delete attachments but keep emails. Here is the link of the task: https://avan.tech/item60587.
I added a delete button that allows you to delete an attached file if it exists on the server.
delete_attachment

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -338,7 +338,7 @@ function process_imap_message_ids($ids) {
  * @return string
  */
 if (!hm_exists('format_msg_part_row')) {
-function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $use_icons=false, $simple_view=false, $mobile=false) {
+function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $rm_link, $use_icons=false, $simple_view=false, $mobile=false) {
     $allowed = array(
         'textplain',
         'texthtml',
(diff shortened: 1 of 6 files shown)


marclaporte:
@NasserNgandu Just to be sure (I didn't test)

- This needs to be optional (default off)
- There should be a confirmation message when deleting an attachment (to avoid accidents)

NasserNgandu:
@marclaporte.
There is already a confirmation message, I can just make it optional.

marclaporte:
@NasserNgandu : no need to make the confirmation message option. This is a rare operation, and it makes sense that the user always confirms.

NasserNgandu:
@marclaporte. Alright I leave it like that.

jasonmunro:
This is not functionally correct I'm afraid (I like the idea and appreciate the work), however attachments are only stored on the web server when attaching them to outgoing E-mail. Attachments on messages you have in your IMAP account and stored in the IMAP server. To remove an attachment requires the following steps:
- read in the entire message and parse the MIME structure
- identify the portion of the message that represents the attachment to delete
- save a new message with all other parts except the deleted part
- delete the original message

marclaporte:
"save a new message with all other parts except the deleted part" -> And not change the date so the sort order remains as before

NasserNgandu:
@marclaporte @jasonmunro.
Could you enlighten me on how to save a new message without touching the timestamp. I did a search and couldn't find a way to do this.

jasonmunro:
> @marclaporte @jasonmunro. Could you enlighten me on how to save a new message without touching the timestamp. I did a search and couldn't find a way to do this.

You cannot. E-mail messages have a date header, this you should maintain from the original E-mail. However IMAP servers manage a created timestamp (commonly referred to as "arrival time") and you cannot change that.

dumblob:
I did not look thoroughly at the RFC but it would be great if we could keep at least some header or an "empty space" at the original place.

If that is entirely impossible, then we shall save the original timestamp e.g. as a (potentially hidden) tag with a predefined prefix to be able to search & find & sort the mails according to the original order which for myself is very important.

Half a year ago I was quickly scripting something using curl and IMAP and I remember some hacky things like issuing [COPY](https://www.atmail.com/blog/imap-commands/ ) to a temporary folder to preserve time and other attributes. IDK but maybe this could be somehow combined with sieve filtering during the "COPY" to the special temporary folder.

PR #599, opened 2022-08-16
Made copy/move buttons clickable also in combined view

josaphatim:
### Issues
- relates https://avan.tech/item81123


jasonmunro:
If we are going to move the function call for this to the core module set we need to also move the function definition - the core module set should work without the imap module set enabled. Thanks!

josaphatim:
> If we are going to move the function call for this to the core module set we need to also move the function definition - the core module set should work without the imap module set enabled. Thanks!

Hello @jasonmunro. I have made changes. Please check. Thanks.

jasonmunro:
thanks, looks good!

PR #607, opened 2022-08-29
Fix mobile look for sieve filter

ElvisAns:
### Issues

- The sieve filter page was not looking good on mobile devices

### Checklist

- To make sure the new styles are applied run the config_gen.php file (eg . `php ./scripts/config_gen.php`)

### How2Test

- Pull this branch, apply the config gen then navigate where you add filters/scripts  for your emails
- Make sure you have configured sieve filters when adding the mail 

### Todo

- Other things related to the whole web app look and feel eg. the login page...

### DIFF
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -1311,11 +1311,11 @@ if (!hm_exists('get_classic_filter_modal_content')) {
                     <small>Input a name and order for your filter. In filters, the order of execution is important. You can define an order value (or priority value) for your filter. Filters will run from lowest to highest priority value.</small>
                 </div>
             </div>
-            <div style="margin-bottom: 10px; margin-top: 25px; ">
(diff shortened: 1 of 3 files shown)


jasonmunro:
This looks good. One thing I noticed was the @screen css based on width 600px. It's fine, however keep in mind we do a server side check for mobile devices and add a class to the body tag called "mobile" when on a mobile. This can be used to apply your style to any mobile devices. Example: https://github.com/jasonmunro/cypht/blob/master/modules/core/site.css#L208

ElvisAns:
I used media queries for applying styles for mobiles devices especially, so instead i am supposed to use only the .mobile class and the server side will add the class whenever the user is browsing from mobile device?

jasonmunro:
> I used media queries for applying styles for mobiles devices especially, so instead i am supposed to use only the .mobile class and the server side will add the class whenever the user is browsing from mobile device?

That is exactly correct :)

ElvisAns:
@jasonmunro  i hope now it is looking good

jasonmunro:
:+1: thanks for updating this

PR #608, opened 2022-08-29
Added Reply-to-all in bcc

josaphatim:
## Pullrequest
When email is bcc, reply-all will reply to from and to emails.

### Issues
- relates https://avan.tech/item72360

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -237,7 +237,16 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                 $addr_list = process_address_fld($lc_headers['cc']);
                 $size += count($addr_list);
             }
-
+            if (array_key_exists('from', $lc_headers)) {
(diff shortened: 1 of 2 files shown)


jasonmunro:
Could you please explain what this intends to do? Most people don't have access to the tiki internal issue tracker so it would be helpful if you could explain a bit more what this change does. Thanks!

josaphatim:
> Could you please explain what this intends to do? Most people don't have access to the tiki internal issue tracker so it would be helpful if you could explain a bit more what this change does. Thanks!

L'intention est que lorsqu'un utilisateur reçoit un e-mail en bcc, answer-all lui permet de répondre aux e-mails dans les en-têtes **To** et **From**. Thanks.

marclaporte:
@jasonmunro Here is the issue, illustrated:

jasonmunro:
Thanks for the follow up on my question!

PR #610, opened 2022-08-31
Added callback handler to close inline message when archiving it

josaphatim:
## Pullrequest
We have the **show inline message** which allow us to view message side by side with message list. It works good in combined view. But elsewhere inline message closes on the next ajax call. 

Another issue is that Archive does close the inline message even if the message is deleted. And the message list isn't updated. This merge request solves these two issues.


jasonmunro:
thanks for this!

PR #612, opened 2022-09-03
Added Kolabnow to Nux

josaphatim:
## Pullrequest
I added Kolabnow to nux services.

### Issues
- relates https://avan.tech/item82262


marclaporte:
Tested and worked beautifully.

PR #615, opened 2022-09-04
Fix composer 2 issues

ElvisAns:
### Issues
- [X] Installation from a system that doesn't have composer 2.* is causing issues

### Checklist
- [ ] On composer update/install composer will try to update its version by running composer self-update
- [X] No composer operation can proceed if composer 2.* is not installed

### How to Test
- [X] Just run `composer install/update`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
         "ext-openssl": "*",
         "ext-session": "*",
         "ext-fileinfo": "*",
-        "webklex/composer-info": "^0.0.1"
+        "webklex/composer-info": "^0.0.1",
+        "composer" : "^2.0.0"
     },
     "suggest": {
         "ext-PDO": "To use database features, this needs to be installed",
@@ -50,6 +51,8 @@
         }
     },
     "scripts": {
+        "pre-install-cmd":  "composer check-platform-reqs",
+        "pre-update-cmd": "composer check-platform-reqs",
         "post-update-cmd": "composer suggest",
         "post-package-install":"composer suggest",
         "post-install-cmd": "composer suggest"


ElvisAns:
The test is failing saying requiring composer is invalid while on my local computer i was able to run `composer install`
image

ElvisAns:
Still failing test

ElvisAns:
Still failing CI/CD tests
image

@jasonmunro , @marclaporte  any support?

ElvisAns:
Still failing Ci/CD tests

image

@marclaporte , @jasonmunro ... can you support?

marclaporte:
Additional info:

https://cypht.org/install.html doesn't specify how to install Composer for the Cypht instance, and thus just running Composer may lead to using Composer v1 and this generates warning messages.

ElvisAns:
I see @marclaporte , where is the website content hosted? So we can clarify the documentation.
I have added a pre install script with composer self-update  ... I need to see it in action for a live server

marclaporte:
https://github.com/jasonmunro/cypht-website

marclaporte:
> https://github.com/jasonmunro/cypht-website

https://github.com/jasonmunro/cypht-website/issues/34

marclaporte:
Here is what prompted all this:


In Tiki, we fetch our own copy of Composer and can control the version. Cypht could do the same thing. Or just let it be and Cypht users should make sure their system Composer is OK.

ElvisAns:
These days some packages are failing to install using composer 1. So https://getcomposer.org/doc/03-cli.md#self-update-selfupdate- composer self-update is supposed to try install internally an up-to-date composer version i.e >2.
Now what I tried to do was to call `composer `self-update` whenever a user want to update or install packages.
I think the CI/CD failure is coming from a naming convention of composer dependencies.

ElvisAns:
Ok, now the test pass! @marclaporte  can you try to test this branch on a dev server?

ElvisAns:
One more issue, permission... for some cases, like here, composer cannot self-update...
i we will point out all of this aspects in the documentation...
image

marclaporte:
So @ElvisAns : For our server, we'll use root to update Composer server-wide. But for Cypht, it should be resilient to situations like this.

@jasonmunro @henrique-borba @kroky We need your guidance here. What do we do?
- Update documentation to indicate Composer v2 as a dependency?
- Install our own copy of Composer? This is what Tiki Manager does. On first run, if Composer is missing, it installs it. We can copy the code as the licenses are compatible.
- Something else?

jasonmunro:
> So @ElvisAns : For our server, we'll use root to update Composer server-wide. But for Cypht, it should be resilient to situations like this.
> 
> @jasonmunro @henrique-borba @kroky We need your guidance here. What do we do?
> 
>     * Update documentation to indicate Composer v2 as a dependency?
> 
>     * Install our own copy of Composer? This is what Tiki Manager does. On first run, if Composer is missing, it installs it. We can copy the code as the licenses are compatible.
> 
>     * Something else?

I think we just update our deps to require composer 2. I was fighting with an update on my production deployment 2 weeks ago and updating to composer 2 resolved everything easily.

marclaporte:
ok @ElvisAns  Please proceed accordingly.

And please add docs, or a link to docs on how to install Composer (which nowadays will get you v2)

ElvisAns:
`NEXT` : Update the documentation.

ElvisAns:
When requiring composer, Scrutinizer is saying package not found... while on my local pc it is working. I will test this on a dev server to confirm

ElvisAns:
You can see composer has also been checked before processing

image

ElvisAns:
For the documentation update https://github.com/jasonmunro/cypht-website/pull/35

PR #617, opened 2022-09-13
Fixed non-draft message disappearing after reply

josaphatim:
## Pullrequest
When replying to an email it is deleted whether it is draft or not. This fixes the bug.

### Issues
- relates https://avan.tech/item82764


marclaporte:
A bit more context here as there is data loss:

The bug: When one replies to an email, the actual email-we-are-replying-to is deleted from the inbox. At first, I thought the "send and archive" feature was misbehaving and emails were archived. But this is not the case.

This regression was recently introduced in Git. We are looking to find when this was introduced so everyone can avoid the affected Git revisions.

PR #618, opened 2022-09-13
Prevent reloading page each time after hm_reload_folders has executed

josaphatim:
## Pullrequest
Some actions such as adding new server using nux modules, process github authorization force reloading folders. After that actions Hm_Utils.get_from_local_storage is always returning false because session storage is deleted on each page reload. This make previous/next email links impossible to see. This PR aims to delete the cookie after is has reloaded folders.

### Issues
- relates https://avan.tech/item60586

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -1145,6 +1145,7 @@ var Hm_Folders = {
             sessionStorage.clear();
             Hm_Utils.restore_local_settings(ui_state);
             Hm_Utils.expand_core_settings();
+            document.cookie = "hm_reload_folders=1; max-age=0";
             return true;
         }
         return false;


marclaporte:
Ok, we'll soon provide access to a server with the issue.

josaphatim:
> I can't reproduce this. prev/next are showing fine for me. The reload folders cookie is managed on the server side and properly deleted. We can add this additional javascript I guess as it won't hurt anything but I don't know if this is the actual solution to whatever you are seeing wrong. I tested by going to the site page, hitting save, then navigating to a an IMAP folder and the prev/next links are present as expected.

Thanks. I have provided a short video to illustrate

https://user-images.githubusercontent.com/80334370/189981044-c0f261f2-173f-4269-a382-1246de2839cf.mp4

marclaporte:
This is a great video @josaphatim

jasonmunro:
Appreciate you looking into this. Something broke on the backend management of the cookie not sure where/when. I fixed it there. I would prefer to not delete cookies in js as I have had issues with that (even on cookies that are not html only) in the past. Fix is pushed to master. Thanks again!

PR #619, opened 2022-09-14
Offer to block sender only for servers that support sieve

josaphatim:
## Pullrequest
Only offer to block sender if it will work and is greyed out else instead of trying to block knowing it will fail.

### Issues
- relates https://avan.tech/item82181

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -282,9 +282,14 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
             $txt .= ' | <a class="hlink" id="copy_message" href="#">'.$this->trans('Copy').'</a>';
             $txt .= ' | <a class="hlink" id="move_message" href="#">'.$this->trans('Move').'</a>';
             $txt .= ' | <a class="archive_link hlink" id="archive_message" href="#">'.$this->trans('Archive').'</a>';
-
+            
             if ($this->get('sieve_filters_enabled')) {
-                $txt .= ' | <a class="block_sender_link hlink" id="block_sender" href="#"><img src="'.Hm_Image_Sources::$lock.'" width="10px"></img> <span id="filter_block_txt">'.$this->trans('Block Sender').'</span></a>';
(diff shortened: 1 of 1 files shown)


modules/imap/output_modules.php
```diff
@@ -282,9 +282,14 @@ protected function output() {
...
             $txt .= ' | <a class="archive_link hlink" id="archive_message" href="#">'.$this->trans('Archive').'</a>';
-
+            
             if ($this->get('sieve_filters_enabled')) {
-                $txt .= ' | <a class="block_sender_link hlink" id="block_sender" href="#"><img src="'.Hm_Image_Sources::$lock.'" width="10px"></img> <span id="filter_block_txt">'.$this->trans('Block Sender').'</span></a>';
+                $imap_server = Hm_IMAP_List::get($this->get('msg_server_id'), false);
+                if (isset($imap_server['sieve_config_host'])) {
+                    $txt .= ' | <a class="block_sender_link hlink" id="block_sender" href="#"><img src="'.Hm_Image_Sources::$lock.'" width="10px"></img> <span id="filter_block_txt">'.$this->trans('Block Sender').'</span></a>';
+                } else {
+                    $txt .= ' | <span title="This server is not configured with Sieve, which is required for blocking senders"><img src="'.Hm_Image_Sources::$lock.'" width="10px"></img> <span id="filter_block_txt">'.$this->trans('Block Sender').'</span></span>';
```
dumblob, line 291: Maybe rephrase it to include a recommendation what to do to fix it?

>This functionality requires the email server support "Sieve" technology which is not provided. Contact your email provider to fix it.

josaphatim: > 

Thanks. But user can choose to not enable sieve filter while the email provider support sieve technology. I propose: 

> This functionality requires the email server support "Sieve" technology which is not provided. Contact your email provider to fix it or enable it.

jasonmunro:
user facing strings should be wrapped in $this->trans() in an output module. If you could fix that I will merge this. Thanks!

PR #622, opened 2022-09-18
Added subscription to folders

josaphatim:
## Pullrequest
Add support for IMAP folder subscriptions

### Issues
- fixes https://github.com/jasonmunro/cypht/issues/582
- relates https://avan.tech/item81126

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -685,6 +685,19 @@ class Hm_IMAP extends Hm_IMAP_Cache {
         return $attributes;
     }
 
+     /**
+     * Subscribe/Unsubscribe folder
+     * @param string $mailbox IMAP mailbox to check
+     * @param string $action boolean
+     * @return boolean failure or success
+     */
+    public function mailbox_subscription($mailbox, $action) {
+        $command = ($action? 'SUBSCRIBE': 'UNSUBSCRIBE').' "'.$this->utf7_encode($mailbox).'"'."\r\n";
+        $this->send_command($command);
+        $response = $this->get_response(false, true);
+        return $this->check_response($response, true);
+    }
+
     /* ------------------ SELECTED STATE COMMANDS -------------------------- */
 
     /**
@@ -2129,7 +2142,7 @@ class Hm_IMAP extends Hm_IMAP_Cache {
      */
     public function get_folder_list_by_level($level='') {
         $result = array();
-        $folders = $this->get_mailbox_list(false, $level, '%');
+        $folders = $this->get_mailbox_list(true, $level, '%');
         foreach ($folders as $name => $folder) {
             $result[$name] = array(
                 'delim' => $folder['delim'],
@@ -2138,6 +2151,29 @@ class Hm_IMAP extends Hm_IMAP_Cache {
                 'noselect' => $folder['noselect'],
                 'id' => bin2hex($folder['basename']),
                 'name_parts' => $folder['name_parts'],
+            );
+        }
+        return $result;
+    }
+
+    /**
+     * return all the folders with subscribed attribute
+     * @return array list of folders
+     */
+    public function get_mailbox_list_with_subscription() {
+        $result = array();
+        $all_folders = $this->get_mailbox_list();
+        $subscribed_folders = array_column($this->get_mailbox_list(true), 'basename');
+        foreach ($all_folders as $name => $folder) {
+            $result[$name] = array(
+                'name' => $folder['name'],
+                'delim' => $folder['delim'],
+                'basename' => $folder['basename'],
+                'children' => $folder['has_kids'],
+                'noselect' => $folder['noselect'],
+                'id' => bin2hex($folder['basename']),
+                'name_parts' => $folder['name_parts'],
+                'subscribed' => in_array($folder['basename'], $subscribed_folders)
             );
         }
         return $result;
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -431,7 +431,11 @@ class Hm_IMAP extends Hm_IMAP_Cache {
                 $check_for_new = false;
 
                 /* full folder name, includes an absolute path of parent folders */
-                $folder = $this->utf7_decode($vals[(count($vals) - 1)]);
+                if ($vals[(count($vals) - 1)] == ')' && $lsub) {
+                    $folder = $this->utf7_decode($vals[6]);
+                } else {
+                    $folder = $this->utf7_decode($vals[(count($vals) - 1)]);
+                }
 
                 /* sometimes LIST responses have dupes */
                 if (isset($folders[$folder]) || !$folder) {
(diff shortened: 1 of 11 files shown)


jasonmunro:
I appreciate your work on this, but I think we should have discussed this a bit more :) 

1. This changes the default behavior to subscribed folders only being displayed. This should be a user level setting and disabled by default. I would prefer to have to opt-in to showing subscribed folders only. Subscription options should be hidden unless enabled.

2. The subscribe/unsubscribe list on the folders page appears to do a recursive list of all folders, this is considered bad form for an IMAP client. It is possible with some server configurations that this will return a TON  of entries (old IMAP servers used to use the users UNIX home directory as the mailbox, and a recursive list all could easily return 100K of mostly unusable filenames). With that said it may not be as much of an issue these days so we may be able to get away with it. Currently we only list folders at the top level of the mailbox, and if you want to drive down to subfolders clicking the "+" does a list of that parent folder only which we cache in the folder list until it is reset. You will note other components on folders page use the same folder tree by cloning it from the folder list and following the same pattern.

3.  subscribing to a subfolder failed, top level folder subscribe/unsubscribe worked as expected.

4. The log shows that a handful of modules are being re-enabled on the same page id which is incorrect. Modules can only be enabled once for a particular page id.
```
[2] => Already registered module for ajax_imap_folders_create re-attempted: login
    [3] => Already registered module for ajax_imap_folders_create re-attempted: default_page_data
    [4] => Already registered module for ajax_imap_folders_create re-attempted: load_user_data
    [5] => Already registered module for ajax_imap_folders_create re-attempted: language
    [6] => Already registered module for ajax_imap_folders_create re-attempted: date
    [7] => Already registered module for ajax_imap_folders_create re-attempted: http_headers
    [8] => Already registered module for ajax_imap_folders_create re-attempted: load_imap_servers_from_config
    [9] => Already registered module for ajax_imap_folders_create re-attempted: process_folder_create
    [10] => Already registered module for ajax_imap_folders_create re-attempted: imap_bust_cache
    [11] => Already registered module for ajax_imap_folders_create re-attempted: close_session_early
```

5. maybe consider moving the subscription options to their own page. If we keep the recursive list all folders IMAP command then this would reduce the times we issue the command to the IMAP server to only we absolutely need to.

6. After subscribing/unsubscribing the folder list needs to be reset otherwise the changes are not visible.

marclaporte:
"we should have discussed this a bit more"

Indeed, we could have a process to ask more questions beforehand. (Ex.: will a merge request on XYZ be accepted? How would you prefer we do Y?) But a lot of things we start don't necessarily get finished soon after. Priorities shift. And some of the developers are in training. So the risk is that we'd open up a lot of discussions, and that many of them lead nowhere. It's more important that we optimize your time. So I am quite OK that the discussion starts on a merge request. This means the junior developer has a good grasp of the challenge, and a fruitful discussion can ensue. And please do feel comfortable in asking for major changes to a MR. The devs are in training and this is a great way for them to learn.

Thanks!

dumblob:
>Indeed, we could have a process to ask more questions beforehand.

That is definitely an overkill IMHO.

I think Jason simply meant that the effort put into this PR could be better focused if a few things got first described in words instead of code as describing in words would probably be sufficient for understanding and would take less time and less effort for the author :wink:.

I for myself find it actually mostly better to "show the code" as soon as possible as that completely eliminates misunderstandings which are often the major source of delays in implementation of features.

josaphatim:
Hello @jasonmunro. I pushed a new commit. Please check.

jasonmunro:
Looks like we have some conflicts here that need to be resolved before re-checking. Thanks!

josaphatim:
> Looks like we have some conflicts here that need to be resolved before re-checking. Thanks!

Already resolved. Thanks.

henrique-borba (approved):

modules/imap/hm-imap.php
```diff
@@ -425,9 +425,13 @@ public function get_mailbox_list($lsub=false, $mailbox='', $keyword='*') {
                 $marked = false;
                 $folder_sort_by = 'ARRIVAL';
                 $check_for_new = false;
-
+                
                 /* full folder name, includes an absolute path of parent folders */
-                $folder = $this->utf7_decode($vals[(count($vals) - 1)]);
+                if ($vals[(count($vals) - 1)] == ')' && $lsub) {
+                    $folder = $this->utf7_decode($vals[6]);
```
kroky, line 431: Why is this hardcoded value of 6?

modules/imap/hm-imap.php
```diff
@@ -2134,6 +2152,12 @@ public function get_folder_list_by_level($level='') {
                 'name_parts' => $folder['name_parts'],
             );
         }
+        if ($with_subscription_state) {
+            $subscribed_folders = array_column($this->get_mailbox_list(true), 'name');
```
kroky, line 2156: For some reason, it doesn't work locally when I test. I unsubscribe from some folders but the list still shows as subscribed.

modules/imap_folders/site.js
```diff
@@ -9,7 +9,10 @@ var folder_page_folder_list = function(container, title, link_class, target, id_
     $('.imap_folder_link', folders).addClass(link_class).removeClass('imap_folder_link');
     folder_location.prepend(folders);
     folder_location.show();
-    $('.'+link_class, folder_location).on("click", function() { return expand_folders_page_list($(this).data('target'), container, link_class, target, id_dest); });
+    if (subscription) {
+        expand_folders_page_list($('.'+link_class, folder_location).data('target'), container, link_class, target, id_dest, true);
```
kroky, line 13: There is a duplicate trigger of this call - one here when subscription = true and the other one on line 15 when auto-clicking on the folders_subscription page resulting in 2 lists of folders with checkboxes.

modules/imap/hm-imap.php
josaphatim: when lsub is true and a folder has sub folders, $vals[(count($vals) - 1)] does not contain the folder name as it contains more keys. That why I used index 6 because the folder name is always at that key when it comes to lsub.

Screenshot 2023-07-02 at 08 24 26

modules/imap/hm-imap.php
josaphatim: I need to test different accounts for that. On my side It is working. The only thing a bit strange that I noticed is that when a subfolder is subscribe but not it's parent, the parent is still being shown.

modules/imap/hm-imap.php
kroky: It might be always index 6 with your tests but I doubt this is the case for every imap server as I don't see it in the specs - you should use the RFC when it comes to IMAP protocol: https://datatracker.ietf.org/doc/html/rfc3501#section-6.3.9
Maybe the LSUB result contains some more flags at the end but still, there should be a standard way to get the folder name out of the list based on strict reference documentation and not on arbitrary array indexing.

modules/imap/hm-imap.php
kroky: I tested with a somewhat recent postfix +dovecot imap server.

josaphatim:
> if (subscription) {

Hello @kroky I made changes.

modules/imap/hm-imap.php
kroky: Can you paste here the full contents of $vals in your case?

josaphatim:
@kroky can you check please, I fixed some issues :
- When parent is unsubscribed but children subscribed and styling to show that parent folder is disabled
- Folder name issue when parent has children and some other flags

kroky:
All looks good to me but it needs some extensive testing. I might be able to do it after the holidays.

josaphatim:
Hello @kroky. Any update on this ?

kroky:
It still seems a bit shaky. I see + LSUB in the folder list now which doesn't have any children.

I also go to manage folders, create a new test folder under INBOX, manage to move emails in and out. Then, turn on show only subscribed folders. I don't see the test one anymore. I go to manage folders, settings icon on the top right, clicked the checkbox of test folder. I still don't see it in the list. Reloaded the folder list but I don't see it. Now I can't go to subscribed folders anymore. I go to manage folders, click the top right hand side cog icon, it redirects to cypht home page. Something's wrong here.

kroky:
Also, please make sure you merge latest master into this branch, so you can actually test against the .env and other enhancements there.

PR #623, opened 2022-09-20
Login page rework

ElvisAns:
### Issues

- [X] The The login page UI

### Checklist

- [X] None

### Screenshots

image

image

### How2Test

- [X] Just try to login and see the new design

### Todo

- [X] Propose any other change to this proposal


jasonmunro:
I think this looks great, but it needs a couple fixes and I would prefer to make this optional based on a site configuration value. I will definitely use it for my personal site but I would prefer to keep the original plain login as the default and sites can opt-in to use the fancy version.

I did not test the recaptcha module set yet to see if it lines up ok but I can do that after we sort out the issues. Specifically the way the 2 image files are being included is incorrect. In order to work with different site setups includes should be prefixed with the WEB_ROOT constant, and assets should be placed in a module set directory. So please move both images to

```
 modules/core/assets/
```
and change the URLS to:
```
src="'.WEB_ROOT. 'modules/core/asset/images/logo.svg"
url('.WEB_ROOT.'modules/core/asset/images/cloud.png);'
```

Note you have to run the config_gen.php script for the assets to be properly copied over to the production version of the site. Thanks!

ElvisAns:
Now everything follow your guidance. Thank you @jasonmunro .
Next is to know how are we going to make it optional, maybe we need to add switch button somewhere and a file or database entry to save the user setting so he may switch between the login screen accordingly...

ElvisAns:
I got the point.
I am going to work on it soon and give you update

ElvisAns:
@jasonmunro the login page UI is know configurable.

Thanks

modules/core/output_modules.php
```diff
@@ -164,7 +164,13 @@ class Hm_Output_login_end extends Hm_Output_Module {
      * Closes the login form
      */
     protected function output() {
-        return '</form>';
+        $config = new Hm_Site_Config_File(CONFIG_FILE);
```
kroky, line 167: Config is available in handler modules - pass whatever is needed to output modules and don't re-initialize the config class again here in an output module.

modules/core/output_modules.php
```diff
@@ -177,28 +183,60 @@ class Hm_Output_login_start extends Hm_Output_Module {
      * Looks at the current login state and outputs the correct form
      */
     protected function output() {
+        $config = new Hm_Site_Config_File(CONFIG_FILE);
```
kroky, line 186: Same as above - config is available in handler modules - pass whatever is needed to output modules and don't re-initialize the config class again here in an output module.

modules/core/output_modules.php
```diff
@@ -177,28 +183,60 @@ class Hm_Output_login_start extends Hm_Output_Module {
...
+                    '#fff);font-size:1em;height:100%;color:#333;font-family:Arial;padding:0px;margin:0px;min-width:700px;'.
+                    'font-size:100%;}input,option,select{font-size:100%;padding:3px;}textarea,select,input{border:solid '.
+                    '1px #ddd;background-color:#fff;color:#333;border-radius:3px;}.screen_reader{position:absolute'.
+                    ';top:auto;width:1px;height:1px;overflow:hidden;}.login_form{float:left;font-size:90%;'.
+                    'padding-top:60px;height:300px;border-radius:0px 0px 20px 0px;margin:0px;background-color:#f5f5f5;'.
+                    'width:300px;padding-left:20px;}.login_form input{clear:both;float:left;padding:4px;margin-left:20px;'.
+                    'margin-top:10px;margin-bottom:10px;}#username,#password{width:200px;}.err{color:red !important;}.long_session'.
+                    '{float:left;}.long_session input{padding:0px;float:none;}.mobile .long_session{float:left;clear:both;}</style>';
+    
+            return $css.'<form class="login_form" method="POST">';
```
kroky, line 207: Indentation

modules/core/output_modules.php
```diff
@@ -218,13 +256,27 @@ protected function output() {
             ' <label for="stay_logged_in">'.$this->trans('Stay logged in').'</label></div>';
         }
         if (!$this->get('router_login_state')) {
-            return '<h1 class="title">'.$this->html_safe($this->get('router_app_name', '')).'</h1>'.
+            $config = new Hm_Site_Config_File(CONFIG_FILE);
+            $fancy_login = $config->get('fancy_login','n');
+            if($fancy_login=='n'):
```
kroky, line 261: Can you make sure to keep the old login as default and only use the new one in case fancy_login = 'y'?

modules/core/output_modules.php
ElvisAns: Gave me headache, untill i remember we had a 'hello world' module which explained modules in cypht.
It is now implemented from handler module

modules/core/output_modules.php
ElvisAns: Yes, for now, defaultly the setting load with 'n' (even when omitted from hm3.ini)

modules/core/handler_modules.php
```diff
@@ -624,6 +624,8 @@ class Hm_Handler_login extends Hm_Handler_Module {
      */
     public $validate_request = true;
     public function process() {
+        $this->out('fancy_login_allowed', $this->config->get('fancy_login', 'n'));
+        error_log($this->config->get('fancy_login', 'n'));
```
kroky, line 628: Can you remove debugging code please?

modules/core/output_modules.php
kroky: Yes, it is much better now, thanks!

modules/core/output_modules.php
kroky: What I meant by indentation was that you had return statement indented the wrong level compared to the nested if statement. Now the diff changes show automated syntax updates in the whole file which is not desired. Please only update code that you change and leave the whitespace or bracket refactoring for its own task if and when we decide to do so. I see you placed opening brackets on a new line, added spaces around concatenation operators and such things - we don't want to disturb the code diff with so much stuff. If we decide to support PSR-4 or another standard, we will do in a separate PR and do for the whole codebase. For now, please just update the places in code that you do functional updates and try to use the syntax formatting in line with the whole project.

ElvisAns: Thanks @kroky , i now understand what you meant

ElvisAns:
@kroky I have a question, I just want to know more about cypht.
I know module handler classes are like controller and the order in which their process() methods are called depends on where they have been attached from the module setup.php file, now i need to know, output modules are called just in the naming in which they are declared?
- Eg we have handler module called `Hm_Handler_login` (technically `login`) then it means that all classes that have `login` in their name will be automatically called when the login handler is executed? like `Hm_Output_login_end` will automatically be called to close tags that were opened in `Hm_Output_login_start` and continued in `Hm_Output_login` ?

kroky:
@ElvisAns, no, I don't think there is such a logic. The output modules are defined much like handler modules in setup.php and are executed in the order defined.

marclaporte:
For the record: It's fine to start this as "optional based on a site configuration value"

But each option is a potentially untested code path. After it's well tested, and we improve design/themes overall , we may remove the option, and everyone will get the fancy login. This will be in a major release (ex.: from 1.6.x to 1.7.x) so users expect some changes.

modules/core/assets/images/cloud.png
marclaporte, line 1: What is the source and the license of this image? And is it compatible with LGPL?

modules/core/output_modules.php
```diff
@@ -218,13 +253,26 @@ protected function output() {
...
+            }
+
+            return '<svg class="user-icon__signin" viewBox="0 0 20 20"><path d="M12.075,10.812c1.358-0.853,2.242-2.507,2.242-4.037c0-2.181-1.795-4.618-4.198-4.618S5.921,4.594,5.921,6.775c0,1.53,0.884,3.185,2.242,4.037c-3.222,0.865-5.6,3.807-5.6,7.298c0,0.23,0.189,0.42,0.42,0.42h14.273c0.23,0,0.42-0.189,0.42-0.42C17.676,14.619,15.297,11.677,12.075,10.812 M6.761,6.775c0-2.162,1.773-3.778,3.358-3.778s3.359,1.616,3.359,3.778c0,2.162-1.774,3.778-3.359,3.778S6.761,8.937,6.761,6.775 M3.415,17.69c0.218-3.51,3.142-6.297,6.704-6.297c3.562,0,6.486,2.787,6.705,6.297H3.415z"></path></svg>
+                <img src="'.WEB_ROOT. 'modules/core/assets/images/logo.svg" style="height:90px;">'.
+                '<!--h1 class="title">'.$this->html_safe($this->get('router_app_name', '')).'</h1-->'.
                 ' <input type="hidden" name="hm_page_key" value="'.Hm_Request_Key::generate().'" />'.
-                ' <label class="screen_reader" for="username">'.$this->trans('Username').'</label>'.
+                ' <label class="label__signin" for="username">'.$this->trans('Username').'</label>'.
```
marclaporte, line 271: Is the double underscore intentional?

ElvisAns: Yes it was intentional but i will replace them with `_` to keep everything consistent

modules/core/assets/images/cloud.png
ElvisAns: We have 3 days left to reach 1 year since this PL was opened, I'm not sure I remember where I picked it up.
Any way, as we are opting for
Screenshot 2023-09-17 223139
instead of :
Screenshot 2023-09-17 222919

This image will be completed removed!

ElvisAns: Ok, @marclaporte let's decide now if we will keep the white background.
Everything is ready!

ElvisAns: @josaphatim thought?

josaphatim: @ElvisAns can you resolve conflicts ?
Also update php unit tests as they are broken.
Finally take in account the recaptcha modules. Thanks

josaphatim:
@ElvisAns can you rebase locally ?

modules/core/assets/images/cloud.png
marclaporte, line 1: What is the source and license of this image?

ElvisAns: @marclaporte maybe we need to replace it ... TBH i dont really remember where i have picked the image also this PR started more than 16months ago

ElvisAns: I have replaced the image with a free one https://www.pexels.com/photo/heavy-clouds-1828305/

PR #624, opened 2022-09-20
ignore sensitive and user specific files to git ignore

ElvisAns:
### Issues
- fixes : the gitignore file don't include a folder that is supposed to be user-specific

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ scripts/test.php
 /vendor/
 /.idea/
 composer.lock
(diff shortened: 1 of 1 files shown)


jasonmunro:
Not sure why you would want your user storage inside the cypht repo but I have no problem adding this

ElvisAns:
Thank you @jasonmunro , i was able to see a file with my username and a token inside that folder being tracked by Git

PR #625, opened 2022-09-23
Add repository layer to handlers for better and centralized entity and session management

henrique-borba:
An MR to fix once and for all the problem of ids in the cypht, in addition to facilitating all the management and access of entities in each module.

## Repository Layer for Modules
All registered and active modules will have an automatically initialized repository. This repository contains all the base resources needed to add, remove, edit and search for objects stored in the session.

https://github.com/jasonmunro/cypht/blob/9b3e9f7236fd0d11b8b6d3c27f1215bc22d46e6b/modules/profiles/modules.php#L130

> The profiles modules search for a profile on its repository using **findById**. The repository is automatically injected into the module handler. No additional configuration or instances is necessary.

## Fully implemented Profile module
I refactored the entire profiles module to work with this implementation and now I'm implementing it in the other modules. This will remove a lot of duplicate code.

Below is the new code for inserting a profile and how it was before:

### Before
https://github.com/jasonmunro/cypht/blob/07605067dbc996c0f5d9e5ffe512c36aa4f85886/modules/profiles/modules.php#L117-L147

### After
https://github.com/jasonmunro/cypht/blob/9c001c33799e806474b253c7ebdc757806c35279/modules/profiles/modules.php#L118-L137


modules/smtp/modules.php
```diff
@@ -418,8 +435,24 @@ public function process() {
...
+                $servers[$index]['id'] = uniqid();
+                $changed = true;
+            }
+        }
+        if ($changed) {
+            $this->session->record_unsaved('SMTP server ID updated');
+        }
+        return $servers;
+    }
+
```
kroky, line 452: generateId function is repeated in 2 modules - better extract it in one place?

modules/smtp/modules.php
```diff
@@ -407,8 +408,24 @@ public function process() {
...
     public function process() {
         $servers = Hm_SMTP_List::dump();
+        $servers = $this->generateId($servers);
```
kroky, line 428: This is good that we generate the IDs if we have cached server data from config. What about the rest of the lists - pop3, imap, profile, feed?

henrique-borba:
@marclaporte @kroky 
Quick update on this:

There is an exaggerated complexity in ids and the way lists are stored and managed in the cypht currently because there is no centralized way to do this, with the exception of servers, all modules decide how to store and manage their entities by themselves and save it in the session. My goal is to simplify the whole entity registration process (profiles, servers, keys, etc) as well as the search (getById, getByName, getByUsername and so on) of this entities.

## Repository Layer for Modules
All registered and active modules will have an automatically initialized repository. This repository contains all the base resources needed to add, remove, edit and search for objects stored in the session.

https://github.com/jasonmunro/cypht/blob/9b3e9f7236fd0d11b8b6d3c27f1215bc22d46e6b/modules/profiles/modules.php#L130

> The profiles modules search for a profile on its repository using **findById**. The repository is automatically injected into the module handler. No additional configuration or instances is necessary.

## Fully implemented Profile module
I refactored the entire profiles module to work with this implementation and now I'm implementing it in the other modules. This will remove a lot of duplicate code.

Below is the new code for inserting a profile and how it was before:

### Before
https://github.com/jasonmunro/cypht/blob/07605067dbc996c0f5d9e5ffe512c36aa4f85886/modules/profiles/modules.php#L117-L147

### After
https://github.com/jasonmunro/cypht/blob/9b3e9f7236fd0d11b8b6d3c27f1215bc22d46e6b/modules/profiles/modules.php#L118-L138

kroky:
@henrique-borba yes, that will be great to implement this repository layout pattern!

marclaporte:
For anyone wondering: @henrique-borba is focusing on https://github.com/NumPower/numpower nowadays.

kroky:
@marclaporte @josaphatim @Shadow243 This MR is ready. I'd love to get some of you test it out. Will also fix the Tiki integration once this is merged (need a couple of int/validation updates). I also have to see the automated tests but more or less this should be ready to be merged.

kroky:
Btw, one drawback of the switch is that existing config will still use integer IDs. We can make a migration script or try to auto-fix when loading data. The problem will be profiles that link one account to another by id - the profiles will need to be re-linked - i.e. set up again. Is this a big issue and should we think about an automated migration script?

marclaporte:
Some people use hundreds of profiles like here: https://github.com/cypht-org/cypht/issues/882

If we have a safe enough way to convert, let's do it

josaphatim:
These are some issues I found:
- Local contact module is still using integer ids
- title bar content has disappeared in message view page and message box path at the title menu too
- Impossible to edit existing profiles until we add a new one (Already known)

kroky:
@josaphatim updated local contacts to use the repository pattern but had to do a bit of special cases where contacts are objects while all other entities are arrays and also we have external contacts in the same Hm_Contact_Store object that are not saved locally.
I am currently unable to test gmail contacts as they deprecated the Contacts API we are using and we need to switch to People API.
Does anybody has a testing case for LDAP contacts?

Title bar issue seems to be a bootstrap migration issue which we can fix later.

kroky:
@marclaporte latest code here does an auto-migration when users load their config data - it replaces all integer IDs with unique IDs and fixes the links between each other.

PR #629, opened 2022-09-29
Delete attachments (but keep email)

josaphatim:
## Pullrequest
Removes specified attachment part keeping the rest of email.

### Issues
- fixes https://github.com/jasonmunro/cypht/issues/401
- relates https://avan.tech/item60587

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1,5 +1,7 @@
 <?php
 
+use ZBateson\MailMimeParser\Message;
+
 /**
  * IMAP modules
  * @package modules
@@ -338,7 +340,7 @@ function process_imap_message_ids($ids) {
  * @return string
  */
 if (!hm_exists('format_msg_part_row')) {
-function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $use_icons=false, $simple_view=false, $mobile=false) {
+function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $at_args, $use_icons=false, $simple_view=false, $mobile=false) {
     $allowed = array(
         'textplain',
         'texthtml',
@@ -437,7 +439,7 @@ function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $
     if ($mobile) {
         $res .= '<div class="part_size">'.$output_mod->html_safe($size);
         $res .= '</div><div class="part_desc">'.$output_mod->html_safe(decode_fld($desc)).'</div>';
-        $res .= '<div class="download_link"><a href="?'.$dl_args.'&amp;imap_msg_part='.$output_mod->html_safe($id).'">'.$output_mod->trans('Download').'</a></div></td></tr>';
+        $res .= '<div class="download_link"><a href="?'.$dl_args.'&amp;imap_msg_part='.$output_mod->html_safe($id).'">'.$output_mod->trans('Download').'</a></div></td>';
     }
     else {
         $res .= '</td><td class="part_size">'.$output_mod->html_safe($size);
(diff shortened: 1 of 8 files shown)


marclaporte:
@jasonmunro @kroky  Since this is deleting data, I could use a third opinion :-)

kroky:
Code looks good but the remove attachment itself looks a bit hacky with all that string manipulation of a mime message. Isn't it better to use a Mime lib to parse the message, remove the attachment from the message structure and then format the message again? I am afraid the current code may fail in some edge cases or ill-formatted messages and replace more than just the attachment in which case a data loss will occur.

dumblob:
Hm, is this done transactionally? If not, could we at least do a manual two-way commit?

Deleting anything is **always** extremely dangerous...

dumblob:
Got a crazy idea now. If this functionality is important for the end user (for me personally it actually is important), then why not to separate attachments from the email immediately after it was received but just before putting it into INBOX. Maybe by having a special filter or whatever. Then presenting it as one email (perhaps this is impossible and IMAP does not support anything like "merge 2 emails on the go" nor anything like "merge an email with some attachments on the go").

This way the received timestamp, IMAP ordering, etc. everything shall be "preserved". Deleting attachments at any time later would be a trivial operation not affecting the message itself.

Thoughts from IMAP experts?

kroky:
The actual deletion consists of 3 steps done synchronously:
1. Remove attachment from the text representation of the message.
2. Create a new message on the imap server.
3. Remove the old message.

This means that data loss can't occur from the IMAP operations. If an op fails (like number 3), the worst case scenario is leaving 2 mails in the mailbox.

I think data loss could occur from improper search/replace operations on the message itself and thus I proposed to use a MIME message lib to handle the message decomposition and composition better.

josaphatim:
> The actual deletion consists of 3 steps done synchronously:
> 
> 1. Remove attachment from the text representation of the message.
> 2. Create a new message on the imap server.
> 3. Remove the old message.
> 
> This means that data loss can't occur from the IMAP operations. If an op fails (like number 3), the worst case scenario is leaving 2 mails in the mailbox.
> 
> I think data loss could occur from improper search/replace operations on the message itself and thus I proposed to use a MIME message lib to handle the message decomposition and composition better.
> 
> Another potential problem is the creation of the message on step 2 might actually change the date of the message as IMAP stores custom creation datetime besides the Date coming from message headers. Not sure if we can modify that creation date to be the same as the original message.

Thanks for that @kroky. I used https://github.com/zbateson/mail-mime-parser to remove attachment.

kroky:
Looks good to me now, thanks @josaphatim !

modules/imap/functions.php
```diff
@@ -446,11 +448,33 @@ function format_msg_part_row($id, $vals, $output_mod, $level, $part, $dl_args, $
...
+/*
+ * Returns the modified message
+ * @param string $attachment_name filename
+ * @param string $msg raw message
+ * @return string
+ */
+if (!hm_exists('remove_attachment')) {
+    function remove_attachment($msg_id, $msg) {
+        $message = Message::from($msg, false);
+        $message->removeAttachmentPart($msg_id - 2);
```
kroky, line 472: Are you sure it is always 2 here? It seems to depend on the actual mime message structure which might have different parts before the attachments - text only, text and html, html-only, mutlipart with several parts before the attachment.

kroky:
@josaphatim can you please resolve the conflict as well, so we can merge?

josaphatim:
> @josaphatim can you please resolve the conflict as well, so we can merge?

Hello @kroky. Conflicts fixed.

modules/imap/functions.php
kroky: @josaphatim I think we need a specific part number here, not an arbitrary number at the end, don't we?

josaphatim: You are right @kroky. But I don't get a way to use a fixed index. When I take a look at https://github.com/zbateson/mail-mime-parser getAllAttachmentParts function it returns a simple array with keys starting from 0. While attachments in cypht start from 2. That why I did $msg_id - 2 for them to correspond.

kroky: Did you try a multipart message with several text parts (more than 2)?

josaphatim: Now found the pattern. https://github.com/zbateson/mail-mime-parser attachments combines embedded images and file attachments. It uses 0-based index. So I created a function to get corresponding index from the given message part. Also I added a confirmation before deleting attachment.

kroky: Yes, that's much better, thanks!

PR #630, opened 2022-09-30
ability to specify sieve client externally, so integrating apps can use a different factory

kroky:
Abstract the way sieve client is constructed, so we can use a different client in Tiki-Cypht integration - we need an internal custom one that saves filters in Tiki, so we can process them later in Tiki - this is in the case when IMAP server doesn't support sieve or we need extra functionality not supported in sieve.
Context: https://gitlab.com/tikiwiki/tiki/-/merge_requests/1837


kroky:
@jasonmunro we are soon branching Tiki 25 and would love to merge https://gitlab.com/tikiwiki/tiki/-/merge_requests/1837 which is ready but depends on this PR. Can you please review and merge if possible? Thanks!

marclaporte:
This is how we will leverage: https://doc.tiki.org/Email-filters

PR #634, opened 2022-10-04
Added visual cue on reload click

josaphatim:
## Pullrequest
Added a spinner to help users know that folders are reloading.

### Issues
- relates https://avan.tech/item82898


jasonmunro:
Manually tested this and I don't see any spinner on folder reload. I put a sleep in to make sure it was not happening too fast but even then there does not appear to be any change in behavior for me.

josaphatim:
> Manually tested this and I don't see any spinner on folder reload. I put a sleep in to make sure it was not happening too fast but even then there does not appear to be any change in behavior for me.

The error may be the deletion of the / at the beginning of the image path. For me it does not display if I add it. I have to review this. Thanks.

jasonmunro:
> > Manually tested this and I don't see any spinner on folder reload. I put a sleep in to make sure it was not happening too fast but even then there does not appear to be any change in behavior for me.
> 
> The error may be the deletion of the / at the beginning of the image path. For me it does not display if I add it. I have to review this. Thanks.

In PHP proper URLs are built with:

```WEB_ROOT.'modules/core/assets/...```

you could create an lnline js function that returns the WEB_ROOT value and use it to build the correct URL prefix. Might be useful other places in the js code as well.  We already have a place we output these types of values https://github.com/jasonmunro/cypht/blob/master/modules/core/output_modules.php#L537

**Comment by josaphatim** [comment]
> > > Manually tested this and I don't see any spinner on folder reload. I put a sleep in to make sure it was not happening too fast but even then there does not appear to be any change in behavior for me.
> > 
> > 
> > The error may be the deletion of the / at the beginning of the image path. For me it does not display if I add it. I have to review this. Thanks.
> 
> In PHP proper URLs are built with:
> 
> `WEB_ROOT.'modules/core/assets/...`
> 
> you could create an lnline js function that returns the WEB_ROOT value and use it to build the correct URL prefix. Might be useful other places in the js code as well. We already have a place we output these types of values https://github.com/jasonmunro/cypht/blob/master/modules/core/output_modules.php#L537

Th

> > > Manually tested this and I don't see any spinner on folder reload. I put a sleep in to make sure it was not happening too fast but even then there does not appear to be any change in behavior for me.
> > 
> > 
> > The error may be the deletion of the / at the beginning of the image path. For me it does not display if I add it. I have to review this. Thanks.
> 
> In PHP proper URLs are built with:
> 
> `WEB_ROOT.'modules/core/assets/...`
> 
> you could create an lnline js function that returns the WEB_ROOT value and use it to build the correct URL prefix. Might be useful other places in the js code as well. We already have a place we output these types of values https://github.com/jasonmunro/cypht/blob/master/modules/core/output_modules.php#L537

Done! Please check.

**Comment by jasonmunro** [comment]
looks good thanks for doing that. I noticed the spinner is only active for a fixed amount of time, and if the folder list does not load in those .5 seconds the spinner goes away which is kind of confusing. You can assign function events to the end of an AJAX call to make this a bit nicer. The following patch against your PR does this:

```diff --git a/modules/core/site.js b/modules/core/site.js
index 8db42e4d..4adf27ad 100644
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -1205,10 +1205,10 @@ var Hm_Folders = {
         $('.update_message_list').on("click", function(e) {
             var text = e.target.innerHTML;
             e.target.innerHTML = '<img src="'+hm_web_root_path()+'modules/core/assets/images/spinner.gif" />';
-            setTimeout(function() {
-                Hm_Folders.update_folder_list();
+            Hm_Folders.update_folder_list();
+            Hm_Ajax.add_callback_hook('hm_reload_folders', function() {
                 e.target.innerHTML = text;
-            }, 500);
+            });
             return false;
         });
         $('.hide_folders').on("click", function() { return Hm_Folders.hide_folder_list(); });
```
The only other thing I noticed is the spinner image background is slightly visible. Maybe a CSS rule is interfering that we can override?

josaphatim:
> looks good thanks for doing that. I noticed the spinner is only active for a fixed amount of time, and if the folder list does not load in those .5 seconds the spinner goes away which is kind of confusing. You can assign function events to the end of an AJAX call to make this a bit nicer. The following patch against your PR does this:
> 
> ```diff
> index 8db42e4d..4adf27ad 100644
> --- a/modules/core/site.js
> +++ b/modules/core/site.js
> @@ -1205,10 +1205,10 @@ var Hm_Folders = {
>          $('.update_message_list').on("click", function(e) {
>              var text = e.target.innerHTML;
>              e.target.innerHTML = '';
> -            setTimeout(function() {
> -                Hm_Folders.update_folder_list();
> +            Hm_Folders.update_folder_list();
> +            Hm_Ajax.add_callback_hook('hm_reload_folders', function() {
>                  e.target.innerHTML = text;
> -            }, 500);
> +            });
>              return false;
>          });
>          $('.hide_folders').on("click", function() { return Hm_Folders.hide_folder_list(); });
> ```
> 
> The only other thing I noticed is the spinner image background is slightly visible. Maybe a CSS rule is interfering that we can override?

For image as it has that background perhaps i must change it or make the spinner with css.

josaphatim:
> looks good thanks for doing that. I noticed the spinner is only active for a fixed amount of time, and if the folder list does not load in those .5 seconds the spinner goes away which is kind of confusing. You can assign function events to the end of an AJAX call to make this a bit nicer. The following patch against your PR does this:
> 
> ```diff
> index 8db42e4d..4adf27ad 100644
> --- a/modules/core/site.js
> +++ b/modules/core/site.js
> @@ -1205,10 +1205,10 @@ var Hm_Folders = {
>          $('.update_message_list').on("click", function(e) {
>              var text = e.target.innerHTML;
>              e.target.innerHTML = '';
> -            setTimeout(function() {
> -                Hm_Folders.update_folder_list();
> +            Hm_Folders.update_folder_list();
> +            Hm_Ajax.add_callback_hook('hm_reload_folders', function() {
>                  e.target.innerHTML = text;
> -            }, 500);
> +            });
>              return false;
>          });
>          $('.hide_folders').on("click", function() { return Hm_Folders.hide_folder_list(); });
> ```
> 
> The only other thing I noticed is the spinner image background is slightly visible. Maybe a CSS rule is interfering that we can override?

I pushed a new commit please check. Thanks.

PR #639, opened 2022-10-15
Fixed bugs related to smtp compose form

josaphatim:
Prevented reply to react like reply to all and also when an email is in cc instead of replying to who sent the mail we were responding to the recipient of the mail in from field.
Also prevented user to respond to the email he sent.

- relates https://avan.tech/item81605


marclaporte:
Should solve https://github.com/jasonmunro/cypht/issues/384

PR #642, opened 2022-10-17
Go to next email in both reply/reply and archive

josaphatim:
Actually Archive, Delete, Reply and archive actions redirects to next email. This Merge request does the same to reply action.

- fixes https://github.com/jasonmunro/cypht/issues/435


marclaporte:
I am using this now and it is saving a lot of time clicking to clean out an inbox.

Thank you @josaphatim and @jasonmunro

PR #645, opened 2022-10-24
Fixed missing data in info page

josaphatim:
Removed code that prevented showing full data in developper info page


kroky:
Fix looks good!

PR #647, opened 2022-10-24
Added Send and archive also in reply all

josaphatim:
Added missing button **reply and archive** when replying to all

Relates: https://avan.tech/item84646

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -1126,7 +1126,7 @@ class Hm_Output_compose_form_content extends Hm_Output_Module {
             smtp_server_dropdown($this->module_output(), $this, $recip, $selected_id).
             '<input class="smtp_send" type="submit" value="'.$this->trans('Send').'" name="smtp_send" '.$send_disabled.'/>';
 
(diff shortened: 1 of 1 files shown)


modules/smtp/modules.php
```diff
@@ -1126,7 +1126,7 @@ protected function output() {
             smtp_server_dropdown($this->module_output(), $this, $recip, $selected_id).
             '<input class="smtp_send" type="submit" value="'.$this->trans('Send').'" name="smtp_send" '.$send_disabled.'/>';
 
-        if ($this->get('list_path') && $reply_type == 'reply') {
+        if ($this->get('list_path')) {
```
kroky, line 1129: Besides reply all, I think we have reply types as forward and maybe others. Are we sure we want this button everywhere?

josaphatim: Hello @kroky. I pushed a new commit please check.

marclaporte:
I tested this, and it works for me. I really like this feature! More details: https://github.com/jasonmunro/cypht/pull/477

PR #649, opened 2022-10-30
Icons - theme compatibility - use Bootstrap icons instead of svg ones

johnsantosDev:
## Pullrequest
PROBLEM & LINK TO TASK DESCRIPTION
image

https://avan.tech/item41850-Cypht-in-Tiki-Change-icon-colors-by-inserting-a-fill-attribute-into-the-SVGs
The mission is to make Cypht icons look acceptable with all the Tiki themes, using the method suggested by Jason (or propose better).
After checking about this issue, I found that the better way is to allow using the existing bootstrap icons instead of SVG ones, why? because they're compatible with all all themes. 
I am making it optional so that user can decide to either keep using svg icons or switch to bootstrap icons.

```
; Bootstrap Icons
; ---------------
; Use Bootrap icons instead of SVG ones on menu icons
default_icons_to_bootstrap_icons=y
```

### Issues
Icons color impactibility with the existing themes

### Checklist
With this change, COMPOSER was updated, here's the version `"twbs/bootstrap-icons": "^1.9"`
 

### How2Test
You need to run composer update in order to make tests
For your test, update the file hm3.ini by adding the line `default_icons_to_bootstrap_icons=y`

 

### Todo

- [X] None


marclaporte:
For the record: @johnsantosDev is investigating how to do this without adding an extra dependency.

Issue #650, opened 2022-10-31
500 on sending mail


capocasa:
## 🐛 Bugreport

A '500 server error' page shows up on sending mail on my installation.

It's a debian-bullseyse system running Apache and PHP 7.4.30. It was first reported on gitter by Calaad. This is from his report, my error was identical.

```
[Fri Jun 24 11:10:21.900376 2022] [php7:notice] [pid 62092] [client XXX.XXX.XXX.XXX:55770] PHP Notice:  Trying to access array offset on value of type null in /usr/local/share/cypht/modules/smtp/modules.php on line 716
[Fri Jun 24 11:10:21.900406 2022] [php7:notice] [pid 62092] [client XXX.XXX.XXX.XXX:55770] PHP Notice:  Undefined index: draft in /usr/local/share/cypht/modules/smtp/modules.php on line 716
[Fri Jun 24 11:10:21.900850 2022] [php7:error] [pid 62092] [client XXX.XXX.XXX.XXX:55770] PHP Fatal error:  Uncaught Error: Call to a member function select_mailbox() on bool in /usr/local/share/cypht/modules/smtp/modules.php:1435\nStack trace:\n#0 /usr/local/share/cypht/modules/smtp/modules.php(716): delete_draft()\n#1 /usr/local/share/cypht/lib/modules_exec.php(175): Hm_Handler_process_compose_form_submit->process()\n#2 /usr/local/share/cypht/lib/modules_exec.php(153): Hm_Module_Exec->run_handler_module()\n#3 /usr/local/share/cypht/lib/dispatch.php(220): Hm_Module_Exec->run_handler_modules()\n#4 /usr/local/share/cypht/lib/dispatch.php(190): Hm_Dispatch->process_request()\n#5 /usr/local/share/cypht/site/index.php(50): Hm_Dispatch->__construct()\n#6 {main}\n  thrown in /usr/local/share/cypht/modules/smtp/modules.php on line 1435

```

### Version & Environment
Rev: 68239a70befe06bb7cddb180a6a858c4f989259b

OS:  Debian Bullseye PHP 7.4.30 on Apache

### Steps to reproduce
1. Install cypht according to instructions
2. Set up an IMAP server and configure cypht to have IMAP login
3. Send a mail.
4. Observe 500

capocasa:
[hm3.ini.txt](https://github.com/jasonmunro/cypht/files/9902311/hm3.ini.txt)

This is the used configuration

marclaporte:
[...]

PR #651, opened 2022-10-31
Dirty fix for 500 on outgoing mail

capocasa:
## Pullrequest
There is a hard-to-reproduce bug that was first reported on gitter and now by me here on github, #650 

This is not a clean fix in that it does not addres root causes, but it allows me to use the software right now apparently without issue and might give additional information.

### Issues
- dirty fix for #650

### Checklist
- [X] None

### How2Test
Reproduce bug #650 and watch it go away with this patch.

### Todo
- [ ] Reproduce: Spin up a debian bullseye instance somewhere, install apache, php and cypht and install cypht using the install script. 
- [ ] Find root cause and fix properly

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -1033,7 +1033,7 @@ class Hm_Output_compose_form_content extends Hm_Output_Module {
                 $cc = $imap_draft['Cc'];
             }
             if (array_key_exists('From', $imap_draft)) {
-                $from = $imap_draft['From'];
+                $from = $imap_draft['From'];select_mailbox
             }
             $draft_id = $msg_uid;
         }
@@ -1528,6 +1528,7 @@ function get_primary_recipient($profiles, $headers, $smtp_servers, $is_draft=Fal
 if (!hm_exists('delete_draft')) {
 function delete_draft($id, $cache, $imap_server_id, $folder) {
     $imap = Hm_IMAP_List::connect($imap_server_id);
+    if ( ! $imap ) return false;
     if ($imap->select_mailbox($folder)) {
         if ($imap->message_action('DELETE', array($id))) {
             $imap->message_action('EXPUNGE', array($id));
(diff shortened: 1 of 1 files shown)


marclaporte:
I need validation from @jasonmunro or @henrique-borba

modules/smtp/modules.php
```diff
@@ -1573,7 +1574,7 @@ function rrmdir($dir) {
                     if (filetype($dir . "/" . $object) == "dir") {
                         rrmdir($dir . "/" . $object); 
                     } else {
-                        unlink($dir . "/" . $object);
+                        unlink($dir . "/" . $object);select_mailbox
```
kroky, line 1577: Probably accidentally pasted `select_mailbox` at the end of this line - it will generate a syntax error, can you remove please?

modules/smtp/modules.php
```diff
@@ -1528,6 +1528,7 @@ function get_primary_recipient($profiles, $headers, $smtp_servers, $is_draft=Fal
 if (!hm_exists('delete_draft')) {
 function delete_draft($id, $cache, $imap_server_id, $folder) {
     $imap = Hm_IMAP_List::connect($imap_server_id);
+    if ( ! $imap ) return false;
```
kroky, line 1531: This is a result of incorrectly setup imap server which is used in a profile but cannot be authenticated (wrong password maybe). At any rate, I agree we should check if imap is authenticated before trying to select a mailbox. Can you use this, instead:
```
if (! imap_authed($imap)) {
    return false;
}
```

marclaporte:
@capocasa We'll soon release Cypht 1.4.0, and it would be ideal to get this fix in.

Thanks!

marclaporte:
@capocasa We need you :-)

capocasa:
I'm afraid I don't have a setup ready at the moment- perhaps try to reproduce on a vanilla debian install?

PR #657, opened 2022-11-29
Added possibility to use custom cache class

josaphatim:
This merge request helps embedded cypht to be integrated with already implemented  cache.

Relates: - https://avan.tech/item46649


marclaporte:
More info:

As Cypht is integrated in Tiki, we want a way to use Tiki's caching system. So for example, clear cache in Tiki also clears cache in Cypht.

Ref: https://gitlab.com/tikiwiki/tiki/-/merge_requests/2040

PR #659, opened 2022-12-10
Added snooze functionnality to imap

josaphatim:
With this merge request users can waiten emails by picking a desired date, or with some predefined helpers (later on the day, tomorrow, next weekend, next week on monday, or the first of next month). Snoozed messages go to **Snoozed** folder. All snoozed emails (with X-Snoozed header) and all emails in snoozed folder can be unsnoozed manually. Automatic unsnoozed is for now done by a script that runs every minute to unsnooze email which dates are reached.

It fixes https://github.com/jasonmunro/cypht/issues/572
And relates https://avan.tech/item59012

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -262,7 +262,7 @@ class Hm_Handler_imap_process_move extends Hm_Handler_Module {
                 }
             }
             elseif (count($moved) == 0) {
-                Hm_Msgs::add('ERRUnable to move/copy selected messages');
+                Hm_Msgs::add('ERR Unable to move/copy selected messages');
             }
             if ($form['imap_move_action'] == 'move' && $form['imap_move_page'] == 'message') {
                 $msgs = Hm_Msgs::get();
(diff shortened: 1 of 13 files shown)


modules/imap/handler_modules.php
```diff
@@ -907,6 +907,86 @@ public function process() {
...
+                }
+            }
+        }
+        $this->out('snoozed_messages', $snoozed_messages);
+        if ($snoozed_messages == count($ids)) {
+            $msg = 'Messages snoozed';
+        } elseif ($snoozed_messages > 0) {
+            $msg = 'Some messages have been snoozed';
+        } else {
+            $msg = 'ERRFailed to snooze selected messages';
```
dumblob, line 948: `ERRFailed...` -> `ERR Failed...`

dumblob:
I did not test this but from the source there seem to be no obvious mistakes.

I wonder how this behaves if one wants to edit a snoozed/scheduled email? Especially I am curious what happens if the snooze fires/triggers (in the background) during editing together with a combination of:

1. pressing plain "send"
2. pressing "snooze"
3. pressing "unsnooze"

All these tricky situations must be visually clearly communicated (e.g. a life countdown) in the UI.

Could you take a look at these and maybe others I missed?

josaphatim:
> I did not test this but from the source there seem to be no obvious mistakes.
> 
> I wonder how this behaves if one wants to edit a snoozed/scheduled email? Especially I am curious what happens if the snooze fires/triggers (in the background) during editing together with a combination of:
> 
> 1. pressing plain "send"
> 2. pressing "snooze"
> 3. pressing "unsnooze"
> 
> All these tricky situations must be visually clearly communicated (e.g. a life countdown) in the UI.
> 
> Could you take a look at these and maybe others I missed?

Well noted. I will work on. Thanks

josaphatim:
Changes made to merge request

modules/imap/handler_modules.php
kroky: It is actually correct without a space. That's how it is done throughout Cypht and Hm_Output_msgs strips the first 3 chars in case they are 'ERR' to show a formatted error message. Adding a space will add it to the message when displaying in the frontend.

modules/imap/handler_modules.php
```diff
@@ -262,7 +262,7 @@ public function process() {
                 }
             }
             elseif (count($moved) == 0) {
-                Hm_Msgs::add('ERRUnable to move/copy selected messages');
+                Hm_Msgs::add('ERR Unable to move/copy selected messages');
```
kroky, line 265: No need for this change.

modules/imap/handler_modules.php
```diff
@@ -924,6 +924,86 @@ public function process() {
...
+        list($success, $form) = $this->process_form(array('imap_snooze_ids', 'imap_snooze_until'));
+        if (!$success) {
+            return;
+        }
+        $snoozed_messages = 0;
+        $snooze_tag = null;
+        if ($form['imap_snooze_until'] != 'unsnooze') {
+            $at = date('D, d M Y H:i:s O');
+            $until = get_snooze_date($form['imap_snooze_until']);
+            $snooze_tag = "X-Snoozed: at $at;\n \tuntil $until";
```
kroky, line 945: Why two line headers? This might cause issues in some MTAs or MUAs. Just use a single line - you don't need the new line here to be able to parse it later.

modules/imap/handler_modules.php
```diff
@@ -924,6 +924,86 @@ public function process() {
...
+     * This should use cron
+     */
+    public function process() {
+        $servers = Hm_IMAP_List::dump();
+        foreach (array_keys($servers) as $server_id) {
+            $cache = Hm_IMAP_List::get_cache($this->cache, $server_id);
+            $imap = Hm_IMAP_List::connect($server_id, $cache);
+            if (imap_authed($imap)) {
+                $folder = 'Snoozed';
+                $ret = $imap->get_mailbox_page($folder, 'DATE', false, 'ALL');
```
kroky, line 990: There might be no such folder 'Snoozed'. Shouldn't you check first before getting the messages inside?

Another problem less likely to happen but still possible - user already created Snoozed folder for same or different purpose - some messages inside might not have the X-Snoozed header. You should check that.

modules/imap/handler_modules.php
```diff
@@ -924,6 +924,86 @@ public function process() {
...
+            $msg = 'Messages snoozed';
+        } elseif ($snoozed_messages > 0) {
+            $msg = 'Some messages have been snoozed';
+        } else {
+            $msg = 'ERRFailed to snooze selected messages';
+        }
+        Hm_Msgs::add($msg);
+        $msgs = Hm_Msgs::get();
+        Hm_Msgs::flush();
+        $this->session->secure_cookie($this->request, 'hm_msgs', base64_encode(json_encode($msgs)));
```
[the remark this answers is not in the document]
josaphatim: This is done wherever there is a redirect to another page. Without doing that messages are lost and user can't see them.

josaphatim:
@kroky I made changes.

modules/imap/handler_modules.php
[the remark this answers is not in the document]
josaphatim: @kroky please check.

PR #661, opened 2022-12-13
fix for some attachments whose keys are exchanged

josaphatim:
Hm_IMAP_Struct::single_format that defines Field order of a single non-text message part sometimes message struct comes with a different order as seen in the screenshot and lead to wrong behaviors (Corrupts PDF generation in this particular case). Here the encoding becoming description, size becoming the encoding.

Screenshot 2022-12-13 at 15 09 43

Relates: https://avan.tech/item61844

### DIFF
--- a/modules/imap/hm-imap-bodystructure.php
+++ b/modules/imap/hm-imap-bodystructure.php
@@ -308,11 +308,13 @@ class Hm_IMAP_Struct {
      */
     private function id_single_part($vals) {
         $res = array();
+        $single_format = false;
         if (isset($vals[0]) && strtolower($vals[0]) == 'text') {
             $flds = $this->text_format;
         }
         else {
             $flds = $this->single_format;
+            $single_format = true;
         }
         foreach($flds as $name => $pos) {
             if (isset($vals[$pos])) {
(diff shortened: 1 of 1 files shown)


modules/imap/functions.php
```diff
@@ -1144,3 +1145,17 @@ function get_personal_ns($imap) {
     );
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('check_message_part')) {
+function check_message_part($part) {
```
kroky, line 1152: Please fix this issue in hm-imap-bodystructure file where this array is generated instead of swapping parameters later. It feels odd to do this here.

We probably have a bug in set_value or attribute_set methods in that Hm_IMAP_Struct class but we need to make sure it returns proper structure with fields that are not swapped.

josaphatim: Screenshot 2023-07-06 at 22 04 15

After debugging I realized that the bug comes from the message itself. Please check.

modules/imap/hm-imap-bodystructure.php
```diff
@@ -274,6 +274,15 @@ private function parse_multi_part_subs($vals) {
             }
             else {
                 $subs = array_merge($subs, $this->id_parts(array($val)));
+                foreach ($subs as &$sub) {
+                    if (!$sub['size'] && is_numeric($sub['encoding'])) {
```
kroky, line 278: id_parts seem to return an array with numeric indexes. I think you should change this code in id_single_part where single_format list of fields point to the wrong positions in $vals array. Please add the exception there either by reordering the indexes in single_format before using them or reordering the vals array. Please also add a comment describing why are we doing this change - it seems like an edge case or improperly formatted message.

josaphatim: @kroky  changes made

marclaporte:
@kroky Are we good now? Tks

PR #669, opened 2023-02-14
Added a warn when user closes tab or browser with unsaved changes to …

josaphatim:
If someone closes browser, or tab and has unsaved changes to settings this attempts to warn him.

Fixes: - https://github.com/jasonmunro/cypht/issues/523


henrique-borba (approved):

modules/core/site.js
```diff
@@ -1783,7 +1783,18 @@ $(function() {
         $('.reset_default_value_select').on("click", reset_default_value_select);
         $('.reset_default_value_input').on("click", reset_default_value_input);
     }
-    
+
+    $(window).on('beforeunload', function(event) {
+        if ($('#unsaved_changes').val() > 0) {
```
kroky, line 1788: Unsaved changes input is updated to 1 only in specific cases in imap/smtp/pop3 modules and a few other special places. It is not updated when user manually changes settings in site settings or profiles, for example. If we count on that setting, we will need to extend all places where settings/forms are updated but not submitted - that would be quite an effort.
Why not use a simpler idea - global JS event handler for form changes and global dirty flag? If flag is raised and there is unsaved data, warn user...

modules/core/site.js
```diff
@@ -1783,7 +1783,18 @@ $(function() {
...
+            event.preventDefault();
+            return event.returnValue = 'Unsaved changes will be lost! Re-enter your password to save and exit';
+        }
+    });
+
+    $('a').on('click', function() {
```
kroky, line 1794: Form submissions via buttons should also stop checking for unsaved changes. In fact, clicking links from the menu is also treated as leaving the page and should warn for unsaved changes, IMHO. Only when submitting the form with changes, we should skip warning.

modules/core/site.js
josaphatim: @kroky I don't get a better idea on how to handle this. As there are some forms submissions (like the compose form and other) that does not lead to something that user can be saved in user settings.

kroky: Having a compose form that is half-filled and leaving the page will make user loose the content. I don't think it matters if it is a compose form, a new folder field, a new smtp-server form half-filled or a settings form. I'd vote for a new setting to trigger this behavior system-wide on and off - warn user for unsaved changes.

modules/core/site.js
josaphatim: @marclaporte, @kroky I just have a question on this:
I other webmail like gmail all we do is automatically saved. Many times in cypht I loose servers and profiles I created because I forget to save. Doesn't it make sense to save automatically ?

marclaporte: @benoitg brought this topic up with me recently :-)

marclaporte: I am fine with auto-save, or warning when-data-can-be-lost-when-leaving-a-section/page

marclaporte: I agree it should be everywhere because loosing data is not cool.

kroky: Warning on dirty forms before leaving page would be a nice addition, I agree. Autosave might be problematic in certain areas - e.g. compose form or forward form.

marclaporte: https://www.npmjs.com/search?q=dirty%20form

third_party/jquery.are-you-sure.js
marclaporte, line 1: It's great to re-use code but we need to know the source, license, etc.

josaphatim:
@kroky I have made changes can you check please ?

modules/core/site.js
```diff
@@ -1802,7 +1802,11 @@ $(function() {
         $('.reset_default_value_select').on("click", reset_default_value_select);
         $('.reset_default_value_input').on("click", reset_default_value_input);
     }
-    
+
+    if (hm_check_dirty_flag()) {
+        $('.add_server, .change_pass form, .add_contact_form, .edit_profile form, .user_settings form, .add_hl_form form, .compose_form').areYouSure();
```
kroky, line 1807: Don't we want to do it on an any form on the site? Maybe except the search forms? Otherwise, we risk omitting forms in modules. I doubt this is the complete list of forms we have.

kroky: I think that's the only issue I find. All the rest looks good!

josaphatim: Thanks @kroky. Fixed !

third_party/jquery.are-you-sure.js
```diff
@@ -0,0 +1,177 @@
+$.fn.areYouSure = function(options) {
```
marclaporte, line 1: Still missing headers/license

PR #677, opened 2023-02-23
Added sieve configuration in hm3 ini

josaphatim:
Fixes: https://github.com/jasonmunro/cypht/issues/671#issuecomment-1440362471

This PR helps enable sieve configuration when IMAP connection is added via the hm3.ini file

### DIFF
--- a/hm3.sample.ini
+++ b/hm3.sample.ini
@@ -94,6 +94,9 @@ imap_auth_port=143
 ; enable this. This is only for TLS enabled sockets (typically on port 993).
 imap_auth_tls=
 
+; The hostname and IP address and port sieve is listening on. Example: mail.abc.net:4190
+imap_auth_sieve_conf=
+
 
 ; POP3 Authentication
 ; -------------------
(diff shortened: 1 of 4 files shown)


knightsg (commented):
@josaphatim tested this and the name of the config should actually be imap_auth_sieve_conf_host - just a heads up that the name in hm3.sample.ini is incorrect

hm3.sample.ini
```diff
@@ -94,6 +94,9 @@ imap_auth_port=143
 ; enable this. This is only for TLS enabled sockets (typically on port 993).
 imap_auth_tls=
 
+; The hostname and IP address and port sieve is listening on. Example: mail.abc.net:4190
```
marclaporte, line 97: Please use example.org as per https://www.rfc-editor.org/rfc/rfc6761.html

josaphatim: Done here https://github.com/jasonmunro/cypht/pull/678

PR #681, opened 2023-02-24
Added all drafts

josaphatim:
Relates: https://avan.tech/item68171

This pull request adds the possibility to have a combined view of drafts folder.


marclaporte:
Wow, that is more code than I expected.

josaphatim:
> Wow, that is more code than I expected.
Certainly. I had to add a draft section in the parameter page. This is what increases the number of lines of code

marclaporte:
It is what it is :-)

PR #685, opened 2023-03-02
Fixed strpos warning in smtp module when there is no to

josaphatim:
When clicking on compose menu, there is no to recipient, php is throwing an warning with strpos php function. This PR fixes that


marclaporte:
We use https://gitlab.com/glitchtip for error tracking, and here is the info

marclaporte:
More context: https://dev.tiki.org/Using-GlitchTip-as-part-of-the-Tiki-development-process

PR #701, opened 2023-04-05
Added ssl/tls support for sieve

josaphatim:
Adding tls:// prefix to sieve configuration host will enable explicit STARTTLS
Adding ssl:// prefix to sieve configuration host will enable implicit SSL

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1402,3 +1402,22 @@ function snooze_dropdown($output, $unsnooze = false) {
     return $txt;
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_sieve_config_host')) {
+function parse_sieve_config_host($host) {
+    $url = parse_url($host);
+    if (! isset($url['host']) || ! isset($url['scheme'])) {
+        throw new Exception('Invalid host provided');
+    }
+    $host = $url['host'];
+    $tls = false;
+    $port = isset($url['port'])? $url['port']: '4190';
+    if ($url['scheme'] == 'ssl') {
+        $host = 'ssl://'.$host;
+    } else {
+        $tls = true;
+    }
+    return [$host, $port, $tls];
+}}
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1408,16 +1408,14 @@ function snooze_dropdown($output, $unsnooze = false) {
 if (!hm_exists('parse_sieve_config_host')) {
 function parse_sieve_config_host($host) {
     $url = parse_url($host);
-    if (! isset($url['host']) || ! isset($url['scheme'])) {
-        throw new Exception('Invalid host provided');
-    }
-    $host = $url['host'];
+    $host = $url['host'] ?? $url['path'];
+    $port = $url['port'] ?? '4190';
+    $scheme = $url['scheme'] ?? 'tcp://';
     $tls = false;
-    $port = isset($url['port'])? $url['port']: '4190';
-    if ($url['scheme'] == 'ssl') {
-        $host = 'ssl://'.$host;
-    } else {
+    if ($scheme != 'ssl') {
         $tls = true;
+    } else {
+        $host = $scheme.$host;
     }
     return [$host, $port, $tls];
 }}
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1411,11 +1411,7 @@ function parse_sieve_config_host($host) {
     $host = $url['host'] ?? $url['path'];
     $port = $url['port'] ?? '4190';
     $scheme = $url['scheme'] ?? 'tcp://';
-    $tls = false;
-    if ($scheme != 'ssl') {
-        $tls = true;
-    } else {
-        $host = $scheme.$host;
-    }
+    $tls = $scheme === 'tls';
+    // $host = '$scheme://'.$host;
     return [$host, $port, $tls];
 }}
--- a/modules/imap_folders/modules.php
+++ b/modules/imap_folders/modules.php
@@ -211,10 +211,10 @@ class Hm_Handler_process_folder_rename extends Hm_Handler_Module {
                         $linked_mailboxes = get_sieve_linked_mailbox($imap_account, $this);
                         if ($linked_mailboxes && in_array($old_folder, $linked_mailboxes)) {
                             require_once VENDOR_PATH.'autoload.php';
-                            $sieve_options = explode(':', $imap_account['sieve_config_host']);
+                            list($sieve_host, $sieve_port, $sieve_tls) = parse_sieve_config_host($imap_account['sieve_config_host']);
                             try {
-                                $client = new \PhpSieveManager\ManageSieve\Client($sieve_options[0], $sieve_options[1]);
-                                $client->connect($imap_account['user'], $imap_account['pass'], false, "", "PLAIN");
+                                $client = new \PhpSieveManager\ManageSieve\Client($sieve_host, $sieve_port);
+                                $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
                                 $script_names = array_filter(
                                     $linked_mailboxes,
                                     function ($value) use($old_folder) { 
@@ -578,10 +578,10 @@ if (!hm_exists('get_sieve_linked_mailbox')) {
             return;
         }
         require_once VENDOR_PATH.'autoload.php';
-        $sieve_options = explode(':', $imap_account['sieve_config_host']);
-        $client = new \PhpSieveManager\ManageSieve\Client($sieve_options[0], $sieve_options[1]);
+        list($sieve_host, $sieve_port, $sieve_tls) = parse_sieve_config_host($imap_account['sieve_config_host']);
+        $client = new \PhpSieveManager\ManageSieve\Client($sieve_host, $sieve_port);
         try {
-            $client->connect($imap_account['user'], $imap_account['pass'], false, "", "PLAIN");
+            $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
             $scripts = $client->listScripts();
             $folders = [];
             foreach ($scripts as $s) {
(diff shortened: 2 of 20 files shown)


modules/imap/functions.php
```diff
@@ -1144,3 +1144,21 @@ function get_personal_ns($imap) {
     );
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_sieve_config_host')) {
+function parse_sieve_config_host($host) {
+    $url = parse_url($host);
+    $host = isset($url['host'])? $url['host']: $url['path'];
```
kroky, line 1153: $url might be false if host is "seriously bad" according to manual. We should handle the case when we can't parse the host.

kroky: It is best to throw an exception if host can't be parsed or is rather invalid (e.g. empty host/path)...

modules/sievefilters/modules.php
```diff
@@ -1545,9 +1545,9 @@ class Hm_Sieve_Client_Factory {
     public function init($user_config = null, $imap_account = null)
     {
         if ($imap_account && ! empty($imap_account['sieve_config_host'])) {
-            $sieve_options = explode(':', $imap_account['sieve_config_host']);
-            $client = new PhpSieveManager\ManageSieve\Client($sieve_options[0], $sieve_options[1]);
-            $client->connect($imap_account['user'], $imap_account['pass'], false, "", "PLAIN");
+            list($sieve_hostname, $sieve_port, $sieve_tls) = parse_sieve_config_host($imap_account['sieve_config_host']);
+            $client = new PhpSieveManager\ManageSieve\Client($sieve_hostname, $sieve_port);
+            $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
```
kroky, line 1550: Add try/catch block here.

josaphatim: @kroky changes have been made. Thanks.

josaphatim: I noticed that the tls option in $client->connect function was not taking in account in henrique-borba/sieve-manager. I will do a merge request for it.

modules/imap/functions.php
```diff
@@ -1144,3 +1144,22 @@ function get_personal_ns($imap) {
     );
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_sieve_config_host')) {
+function parse_sieve_config_host($host) {
+    $url = parse_url($host);
+    if (! $url || ! isset($url['host']) || ! isset($url['scheme']) ) {
```
kroky, line 1153: If the prefix scheme is optional, just require host here and actual parsing here (i.e. not false).

modules/sievefilters/modules.php
```diff
@@ -1545,10 +1545,14 @@ class Hm_Sieve_Client_Factory {
...
-            $client = new PhpSieveManager\ManageSieve\Client($sieve_options[0], $sieve_options[1]);
-            $client->connect($imap_account['user'], $imap_account['pass'], false, "", "PLAIN");
-            return $client;
+            try {
+                list($sieve_hostname, $sieve_port, $sieve_tls) = parse_sieve_config_host($imap_account['sieve_config_host']);
+                $client = new PhpSieveManager\ManageSieve\Client($sieve_hostname, $sieve_port);
+                $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
+                return $client;
+            } catch (Exception $e) {
+                return null;
```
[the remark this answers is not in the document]
josaphatim: @kroky now good.

modules/imap/functions.php
```diff
@@ -1217,3 +1217,23 @@ function get_request_params($request) {
...
     return [$server_id, $uid, $folder, $msg_id];
 }}
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_sieve_config_host')) {
+function parse_sieve_config_host($host) {
+    $url = parse_url($host);
+    if (! isset($url['host']) || ! isset($url['scheme'])) {
```
kroky, line 1227: Does this work when $host is IP address or simple hostname without scheme? It should.

josaphatim: @kroky with ip address it works but without scheme it is much difficult to validate if it is a valid url

josaphatim: @kroky I also fixed issue when Hm_Sieve_Client_Factory::init returns null

modules/sievefilters/modules.php
```diff
@@ -1170,11 +1191,10 @@ protected function output() {
 }
 
 function get_blocked_senders_array($mailbox, $site_config, $user_config) {
-    try {
-        $factory = get_sieve_client_factory($site_config);
-        $client = $factory->init($user_config, $mailbox);
-    } catch (Exception $e) {
-        return [];
+    $factory = get_sieve_client_factory($site_config);
+    $client = $factory->init($user_config, $mailbox);
```
kroky, line 1195: Same thing here with exception handling...

modules/sievefilters/modules.php
```diff
@@ -1458,18 +1478,17 @@ function get_classic_filter_modal_content()
...
-            $client = $factory->init($user_config, $mailbox);
-            $scripts = [];
-            foreach ($client->listScripts() as $script) {
-                if (strstr($script, 'cypht')) {
-                    $scripts[] = $script;
-                }
-            }
-        } catch (PhpSieveManager\Exceptions\SocketException $e) {
+        $factory = get_sieve_client_factory($site_config);
+        $client = $factory->init($user_config, $mailbox);
```
kroky, line 1482: Exception handling...

modules/sievefilters/modules.php
```diff
@@ -182,11 +193,10 @@ public function process() {
...
 }
 
-function get_blocked_senders($mailbox, $mailbox_id, $icon_svg, $icon_block_domain_svg, $site_config, $user_config, $module) {
-    try {
-        $factory = get_sieve_client_factory($site_config);
-        $client = $factory->init($user_config, $mailbox);
-    } catch (Exception $e) {
+function get_blocked_senders($mailbox, $mailbox_id, $icon_svg, $icon_block_domain_svg, $site_config, $user_config) {
+    $factory = get_sieve_client_factory($site_config);
+    $client = $factory->init($user_config, $mailbox);
```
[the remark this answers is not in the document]
kroky: Alright, I saw you handle it in the sieve client factory itself.

modules/sievefilters/modules.php
```diff
@@ -1687,12 +1706,17 @@ class Hm_Sieve_Client_Factory {
...
-            return $client;
-        } else {
-            return null;
+            try {
+                list($sieve_hostname, $sieve_port, $sieve_tls) = parse_sieve_config_host($imap_account['sieve_config_host']);
+                $client = new PhpSieveManager\ManageSieve\Client($sieve_hostname, $sieve_port);
+                $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
+                return $client;
+            } catch (PhpSieveManager\Exceptions\SocketException $e) {
+                // An error occured
```
kroky, line 1715: Please forward the exception message to the user via error messaging.

kroky: We should at least log the exception message...

modules/sievefilters/modules.php
```diff
@@ -1458,18 +1478,17 @@ function get_classic_filter_modal_content()
...
-                }
-            }
-        } catch (PhpSieveManager\Exceptions\SocketException $e) {
+        $factory = get_sieve_client_factory($site_config);
+        $client = $factory->init($user_config, $mailbox);
+        if (!$client) {
             return '';
         }
+        $scripts = [];
+        foreach ($client->listScripts() as $script) {
```
kroky, line 1487: Methods on $client can still throw exceptions - e.g. listScripts, calls sendCommand which calls readResponse which can throw a SocketException. We should either refactor php-sieve-manager or handle these exceptions here - otherwise, Cypht will break when unhandled exceptions occur.

modules/imap/functions.php
kroky: When user supplied host = 192.168.0.1 and port = something - this check will fail as parse_url returns only path for IP addresses. I don't think users supply scheme when they supply IP addresses - looks weird to me. Maybe prepend tcp:// scheme if there is no scheme?

josaphatim: The host and the port are both provided in a single string. Then I can prepend tcp:// if there is no scheme.

modules/sievefilters/modules.php
josaphatim: @kroky Now Client exceptions are caught

modules/imap/functions.php
josaphatim: Passing the protocol to socket_connect is leading to an. But I noticed that tls option of Client::connect in sieve manager is not used. I will fix this.

kroky: @josaphatim is this ready now?

josaphatim: @kroky I think only tls will be used. Not ssl

modules/imap_folders/modules.php
```diff
@@ -212,30 +212,34 @@ public function process() {
                         if ($linked_mailboxes && in_array($old_folder, $linked_mailboxes)) {
                             require_once VENDOR_PATH.'autoload.php';
                             $sieve_options = explode(':', $imap_account['sieve_config_host']);
```
kroky, line 214: Shouldn't we parse this configuration with the new parse_sieve_config_host function?

modules/imap_folders/modules.php
```diff
@@ -576,19 +580,24 @@ function get_sieve_linked_mailbox ($imap_account, $module) {
         require_once VENDOR_PATH.'autoload.php';
         $sieve_options = explode(':', $imap_account['sieve_config_host']);
```
kroky, line 581: Shouldn't we parse this configuration with the new parse_sieve_config_host function?

kroky:
I think you need to update Hm_Sieve_Client_Factory and also the Tiki integration which uses a port of Hm_Sieve_Client_Factory to use the new parsing method.

modules/imap/functions.php
kroky: I mean this https://github.com/cypht-org/php-sieve-manager/blob/master/src/ManageSieve/Client.php#L551 still doesn't use the passed tls parameter. Why bother set and pass it then?

josaphatim:
> Hm_Sieve_Client_Factory

Hello @kroky I made changes. Thanks.

kroky:
I don't see changes in php-sieve-manager library - how did we fix the tls parameter passing to client->connect that has no effect?

marclaporte:
@kroky @josaphatim Is this suitable to backport to branch 1.4.x, or we just wait for 1.5.0?

kroky:
It should be fine to be backported.

PR #702, opened 2023-04-18
Added warning message when curl is not eenabled

josaphatim:
Relates: https://github.com/jasonmunro/cypht/issues/690

Added a nicer error message when cURL extension is not enabled.


lib/framework.php
```diff
@@ -139,7 +139,11 @@ public static function c_setopt($handle, $name, $value) {
          * @return resource|false
          */
         public static function c_init() {
-            return curl_init();
+            if (extension_loaded('curl')) {
+                return curl_init();
+            } else {
+                Hm_Msgs::add('ERR Please enable the cURL extension.');
```
kroky, line 145: Note that everywhere in Cypht, ERR is prepended to the error message without a space since core/output_modules Hm_Output_msgs class strips just the first 3 chars of the message before displaying it.

josaphatim: @kroky  Changes made.

PR #703, opened 2023-04-18
Added MDN/SDN to compose form

josaphatim:
Fixes https://github.com/jasonmunro/cypht/issues/687

Add MDN / SDN checkboxes to compose form.

### DIFF
--- a/modules/smtp/hm-smtp.php
+++ b/modules/smtp/hm-smtp.php
@@ -522,17 +522,24 @@ class Hm_SMTP {
     }
 
     /* Send a message */
-    function send_message($from, $recipients, $message) {
+    function send_message($from, $recipients, $message, $from_params = '', $recipients_params = '') {
         $this->clean($from);
-        $command = 'MAIL FROM:<'.$from.'>';
+        if ($from_params) {
+            $from_params = ' ' . $from_params;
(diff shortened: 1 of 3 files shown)


kroky:
This PR addresses the request part only (checkboxes on compose form) but Cypht is also a MUA that should handle the recipient part of the story as well (or should it?).
MTA can optionally respond with MDN or might leave that to MUA. See https://datatracker.ietf.org/doc/html/rfc8098#autoid-10 for format of the MDN that should be sent and also https://datatracker.ietf.org/doc/html/rfc8098#autoid-26 - we have a lot to implement in Cypht if we want to support the full timeline of events.
Is the PR only supposed to allow requesting these MDNs but not respond to them? I think @jasonmunro disagrees with integrating them in Cypht, at least with the recipient part of the story.

josaphatim:
@kroky I made changes please check

kroky:
OK, I think the code is good to be merged for the part that requests a MDN. It seems we didn't have an agreement to implement recipient part of the MDN (sending MDNs from cypht), so leaving that for later.
@josaphatim can you please resolve the conflicts and we can merge.

josaphatim:
@kroky now read

PR #704, opened 2023-05-06
Added all junk

josaphatim:
Added all spam. Also added Junk as special folder.


modules/imap/functions.php
```diff
@@ -17,16 +17,16 @@
  * @return array
  */
 if (!hm_exists('imap_sent_sources')) {
-function imap_sent_sources($callback, $mod) {
+function imap_sent_sources($callback, $mod, $folder = 'sent') {
```
henrique-borba, line 20: If the function is going to be used for any special directory, better use the name `imap_sources`

modules/imap/handler_modules.php
```diff
@@ -1408,8 +1411,9 @@ public function process() {
             if ($callback != 'imap_background_unread_content') {
                 $this->out('move_copy_controls', true);
             }
-            if ($path == 'sent') {
-                foreach (imap_sent_sources($callback, $this) as $vals) {
+            if ($path == 'sent' || $path == 'junk') {
+                $folder = $path == 'junk'? 'junk': 'sent';
+                foreach (imap_sent_sources($callback, $this, $folder) as $vals) {
```
henrique-borba, line 1416: Don´t forget to change the call from `imap_sent_sources` to `imap_sources` here

henrique-borba (approved):

kroky:
@josaphatim the PR looks good to me. Please address Henrique's comments and we can merge.

kroky:
Btw, #681 and #697 repeat a lot of the code in this PR... shouldn't we just combine 3 of them into one MR and reuse whatever functions we can? Having the same code copy-pasted in so many places is not desirable. E.g. Hm_Handler_imap_trash and related methods - each one does exactly the same thing, just receives the folder name as parameter. We can combine into one method and one ajax page to list these. Pretty sure the JS code can be combined as well.

PR #717, opened 2023-06-12
Updated the Italian translation

Magneticdud:
## Pullrequest
I updated the Italian strings for localization

I checked with a regex comparing to the german translation and i didn't see any error

### Issues
- [X] None


marclaporte:
Thank you!

PR #721, opened 2023-06-15
Sieve: added more options to block/ban page

josaphatim:
This merge request adds more options to block/ban page

It helps users to know who is blocked offering the possibility to block
- The whole domain
- Specific email

Also it helps choose the action to be executed
- Default action: This is specified in the block_list page where users choose the default action (Discard/Reject with a default message)
- Discard
- Move to Banned folder
- Reject with default message
- Reject with specific message

Relates: https://avan.tech/item92586

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -1766,6 +1766,9 @@ class Hm_Handler_imap_message_content extends Hm_Handler_Module {
             $this->out('msg_server_id', $form['imap_server_id']);
             $this->out('msg_folder', $form['folder']);
             $this->out('msg_list_path', 'imap_'.$form['imap_server_id'].'_'.$form['folder']);
+            $this->out('site_config', $this->config);
+            $this->out('user_config', $this->user_config);
+            $this->out('imap_accounts', $this->user_config->get('imap_servers'), array());
             $this->out('show_pagination_links', $this->user_config->get('pagination_links_setting', true));
             $part = false;
             $prefetch = false;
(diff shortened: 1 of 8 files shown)


marclaporte:
Will require this: https://github.com/henrique-borba/php-sieve-manager/pull/2

marclaporte:
About terminology: We are using many terms, like
* Banned
* Block
* Reject
* Bounce
* Discard
* ....
And some of these are synonyms in our context.


I think in the code and visible to end users, we should avoid synonyms and converge on a terminology.

We should stick to upstream standard, unless there is a reason to diverge. Ex.: It's not good/clear.

So let's make the list of synonyms, and after, we can pick the best one.
* Block, ban
* Reject, bounce
* Discard, delete, remove
* Any others?

kroky:
Looks good to me. @josaphatim are you still working on something here?

josaphatim:
> Looks good to me. @josaphatim are you still working on something here?

I just need to change terminology as @marclaporte suggested.

josaphatim:
> About terminology: We are using many terms, like
> 
> * Banned
> * Block
> * Reject
> * Bounce
> * Discard
> * ....
>   And some of these are synonyms in our context.
> 
> I think in the code and visible to end users, we should avoid synonyms and converge on a terminology.
> 
> We should stick to upstream standard, unless there is a reason to diverge. Ex.: It's not good/clear.
> 
> So let's make the list of synonyms, and after, we can pick the best one.
> 
> * Block, ban
> * Reject, bounce
> * Discard, delete, remove
> * Any others?

@marclaporte I think now everything is okay

marclaporte:
Terminology looks good to me.

PR #730, opened 2023-07-10
Fixed nux development updates

josaphatim:
when the maximum amount of time allowed to make the connection to the server is set to 2 results are returned.


josaphatim:
> Doesn't make much sense to be so restrictive - can you increase to at least 5 seconds please? Also, there are conflicts to be resolved...

@kroky now good

PR #731, opened 2023-07-12
Added possibility to see and edit sender behavior

josaphatim:
In block list page, I added option to view current behavior and edit it per blocked email/domain.

### DIFF
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -196,12 +196,18 @@ function get_blocked_senders($mailbox, $mailbox_id, $icon_svg, $icon_block_domai
     $current_script = $client->getScript('blocked_senders');
     $blocked_list_actions = [];
     if ($current_script != '') {
-        $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[1]);
-        $blocked_list = json_decode(base64_decode($base64_obj));
-        $base64_obj_actions = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[2]);
-        $blocked_list_actions = json_decode(base64_decode($base64_obj_actions), true);
+        $script_split = preg_split('#\r?\n#', $current_script, 0);
+        if (!isset($script_split[1])) {
+            return '';
+        }
+        $base64_obj = str_replace("# ", "", $script_split[1]);
+        $blocked_list = json_decode(base64_decode($base64_obj));        
         if (!$blocked_list) {
             return '';
+        }
+        if (isset($script_split[2])) {
+            $base64_obj_actions = str_replace("# ", "", $script_split[2]);
+            $blocked_list_actions = json_decode(base64_decode($base64_obj_actions), true);
         }
         foreach ($blocked_list as $blocked_sender) {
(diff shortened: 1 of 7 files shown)


modules/sievefilters/modules.php
```diff
@@ -194,9 +194,12 @@ function get_blocked_senders($mailbox, $mailbox_id, $icon_svg, $icon_block_domai
         return '';
     }
     $current_script = $client->getScript('blocked_senders');
+    $blocked_list_actions = [];
     if ($current_script != '') {
         $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[1]);
         $blocked_list = json_decode(base64_decode($base64_obj));
+        $base64_obj_actions = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[2]);
```
kroky, line 201: There is a possibility that $current_script is not in the format you expect, so preg_split()[] will fail. Also, this line is repeated two times. Can you make it more robust by splitting first, then checking if the format is OK (split result contains 2 items) and then proceed to use them?

modules/sievefilters/site.js
```diff
@@ -222,6 +232,44 @@ $(function () {
...
+            let parent = $(this).closest('tr');
+            let elem = parent.find('#block_action');
+            let sender = $(this).closest('tr').children().html();
+            let scope = sender.startsWith('*@') ? 'domain': 'sender';
+
+            $('.dropdown').toggle();
+            Hm_Ajax.request(
+                [
+                    {'name': 'hm_ajax_hook', 'value': 'ajax_sieve_block_unblock'},
+                    {'name': 'imap_server_id', 'value': $(this).closest('.dropdown').prev().attr('mailbox_id')},
```
kroky, line 246: Can we add the mailbox_id as part of the form, so we don't have to traverse the DOM to find out the ID? This traversal is fragile - as soon as we change the DOM for some other reason, this code will break.

modules/sievefilters/site.js
```diff
@@ -222,6 +232,44 @@ $(function () {
             );
         });
 
+        $(document).on('click', '#edit_blocked_behavior', function(e) {
+            e.preventDefault();
+            let parent = $(this).closest('tr');
+            let elem = parent.find('#block_action');
+            let sender = $(this).closest('tr').children().html();
```
kroky, line 239: Isn't the sender in the first child element (first td of the row)?

modules/sievefilters/modules.php
josaphatim: Current script has always this format. Sometimes line 2 is empty.
`
# CYPHT CONFIG HEADER - DON'T REMOVE
# PyIqQCpnioqwerbWFbC5jb20iLCIqQG1hcUoiN6m9YXBvcnRlLmNvbSJd
# eyIqQG1hcmNsYXBvcnRlLmNvbSI6eyJhY3Rpb24iOiJyZWplY3RfZGVmYX

# Requirements
require ["regex","reject"];

if header :contains "From" ["*@*xyz.com"] {
	discard;
	stop;
}
`

josaphatim:
Hello @kroky
I worked on above issues and also fixed the drop down issue.

PR #735, opened 2023-07-31
Set simple UI as the default in Cypht

henochit:
### Set simpler UI as the default in Cypht
- [X] Simple UI is disabled by default
- [X] Make download options simpler to understand

### How2Test

- [X] Reset config
- [X] In general config check if "Show simple message part structure when reading a message" is checked by default
- [X] See the preview of a mail with attachments


marclaporte:
@henochit This branch has conflicts that must be resolved

modules/imap/output_modules.php
```diff
@@ -901,6 +901,7 @@ protected function output() {
         $reset = '';
         if (array_key_exists('simple_msg_parts', $settings) && $settings['simple_msg_parts']) {
             $checked = ' checked="checked"';
+        } else {
```
kroky, line 904: This else means that the reset checkbox below will be without label - seems like a bug to me. What is the purpose? If the checkbox shouldn't be visible at all, then you should hide the checkbox as well.

modules/imap/setup.php
```diff
@@ -333,6 +335,7 @@
         'msg_part_icons' => FILTER_VALIDATE_BOOLEAN,
         'simple_msg_parts' => FILTER_VALIDATE_BOOLEAN,
         'pagination_links' => FILTER_VALIDATE_BOOLEAN,
+        'imap_per_page_setting' => FILTER_VALIDATE_BOOLEAN,
```
kroky, line 338: This doesn't seem related to the present PR. Are you fixing another bug here?

modules/imap/output_modules.php
kroky: Ok, I see!

modules/imap/functions.php
```diff
@@ -579,26 +579,56 @@ function format_msg_part_section($struct, $output_mod, $part, $dl_link, $at_link
...
+        $res = format_attachment($struct,  $output_mod, $part, $dl_link);
     }
     return $res;
 }}
 
+function format_attachment ($struct,  $output_mod, $part, $dl_args) {
+    $res = '';
+
+    foreach ($struct as $id => $vals) {
+        if(is_array($vals) && $vals['type'] != 'multipart' && isset($vals['file_attributes']) && !empty($vals['file_attributes'])) {
```
kroky, line 611: isset($vals['type']) - same like the block above in not simple view if statement - I am afraid, otherwise, $vals can be an array but missing type key will issue a warning.

henochit: Done

Issue #732, opened 2023-07-14
updating cypth and windows 1257 error on php 8.2


mcbmcb0:
## 💬 Question
I've updated to php 8.2.7 (from 7.3.3) and now see the following:
`PHP Fatal error:  Uncaught ValueError: mb_convert_encoding(): Argument #3 ($from_encoding) contains invalid encoding "windows-1257" in /var/www/cypht/modules/core/message_functions.php`
1. strangely, i don't even know if i have the current version of cypth as can't see a current version displayed nor find info on updating.... is it a reinstall or does it self update?
2. if cypht is up to date... any clues on that `mb_convert_encoding()` error message? 
thanks!

mcbmcb0:
further, i noticed that the function creating the error is unchanged in the new cypht 1.4, and that this issue has appeared in nextcloud and roundcube as well - eg [https://github.com/roundcube/roundcubemail/issues/1004](https://github.com/roundcube/roundcubemail/issues/1004). apparently: 
> as of PHP 8.0.0, a ValueError is thrown if the value of to_encoding or from_encoding is an invalid encoding

adapting the roundcube patch i hacked this into message_functions.php:

```
$fld = substr($fld, (strpos($fld, '?') + 1)); // EXISTING line 403
      
$alt_charset = ['UTF-7'=>"UTF7-IMAP", 'WINDOWS-1257'=> "ISO-8859-13"]; 
if(array_key_exists(strtoupper($charset), $alt_charset)){// MB
   $charset = $alt_charset[strtoupper($charset)];
 }

if (strtoupper($encoding) == 'B') { // existing line 404
...
```

this appears to work. the array can obviously expand to other similar problems (eg windows 1250 and 1265 are reported as problems by some).
i tried using iconv for these character sets but had lots of '`Wrong encoding`' errors so didn't bother.

this is temporary and not elegant but may help someone until a proper fix is found.

marclaporte:
Thank you @mcbmcb0 

@josaphatim and I will investigate along with your other PHP8 report.

marclaporte:
FYI, Cypht is bundled in Tiki so it will have a lot of extra eyeballs in this context:
- https://doc.tiki.org/PHP8
- https://doc.tiki.org/Cypht

[...]

PR #738, opened 2023-08-02
Handled unsupported charsets on mb_convert_encoding

josaphatim:
- fixes (https://github.com/cypht-org/cypht/issues/732)

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -403,15 +403,25 @@ function decode_fld($string) {
             $encoding = $fld[0];
             $fld = substr($fld, (strpos($fld, '?') + 1));
             if (strtoupper($encoding) == 'B') {
-                $fld = mb_convert_encoding(base64_decode($fld), 'UTF-8', $charset);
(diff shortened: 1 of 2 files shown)


modules/core/message_functions.php
```diff
@@ -403,17 +403,27 @@ function decode_fld($string) {
...
     }
     return trim($string);
 }}
 
+if (!hm_exists('convert_to_utf8')) {
+function convert_to_utf8($data, $from_encoding) {
+    try {
+        $data = mb_convert_encoding($data, 'UTF-8', $from_encoding);
+    } catch (ValueError $e) {
+        $data = iconv($from_encoding, 'UTF-8', $data);
```
kroky, line 422: I see comments that iconv handles more encodings than mb_convert_encoding, so it is good to use it as a fallback. However, we should also require it as an extension in our composer.json - ext-iconv might not be available?

josaphatim: @kroky ext-iconv added in composer.json

PR #743, opened 2023-08-11
 [NEW] Error message if there is a space at the beginning or end of username or password

NasserNgandu:
Display a message to the user during login if the username or password contains a space at the beginning or at the end. Here is the link to the task: https://avan.tech/item10670


kk6mrp:
I can understand spaces in usernames, but why the restriction for spaces in the password?

marclaporte:
Users copy-paste usernames/passwords and a space creeps in causing login failure, blocked accounts, etc.

Unless we think there is going to be a password with a space, it should be trimmed, or at least a warning given. (ex.: "You have a leading or trailing space")

https://ux.stackexchange.com/questions/75686/why-do-some-websites-including-google-trim-whitespaces-in-passwords
https://ux.stackexchange.com/questions/8010/should-i-trim-spaces-in-passwords
http://stackoverflow.com/questions/7240869/should-i-trim-spaces-in-a-password-field
http://tonyshowoff.com/articles/should-you-trim-passwords/
https://github.com/syncthing/syncthing/issues/3935

kroky:
Looks good to me. Is the login prompt only affected place? What about imap/etc. server definitions where usernames and passwords are usually pasted? A warning seems a nice solution because user can still ignore and proceed. Maybe think about a warning style that's not so annoying as js alert - e.g. Cypht-style error messages at the top that user can see but doesn't need to click something to close...

marclaporte:
@kroky Agreed IMAP/etc. accounts should also have the warning.

Agreed as well on not too annoying alerting...

NasserNgandu:
> @kroky Agreed IMAP/etc. accounts should also have the warning.
> 
> Agreed as well on not too annoying alerting...

Alright, I'm working on it.

marclaporte:
There is a mismatch between the description of the MR and the code.

NasserNgandu:
> There is a mismatch between the description of the MR and the code.

Yes there was a conflict here is another merge request : https://github.com/cypht-org/cypht/pull/780.

Issue #734, opened 2023-07-21
PHP8 - Fatal error with some imap attachments?


mcbmcb0:
## 💬 Question

I've put this as a questions tho I think its a bug due to php making count types compulsory. but I don't know how to reliably reproduce it:
```
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, bool given in /var/www/cypht/modules/imap/functions.php:914
Stack trace:
#0 /var/www/cypht/modules/imap/functions.php(476): get_imap_part_name()
#1 /var/www/cypht/modules/imap/functions.php(410): get_part_desc()
#2 /var/www/cypht/modules/imap/functions.php(534): format_msg_part_row()
#3 /var/www/cypht/modules/imap/functions.php(540): format_msg_part_section()
#4 /var/www/cypht/modules/imap/output_modules.php(152): format_msg_part_section()
#5 /var/www/cypht/lib/module.php(526): Hm_Output_filter_message_struct->output()
#6 /var/www/cypht/lib/modules_exec.php(96): Hm_Output_Module->output_content()
#7 /var/www/cypht/lib/modules_exec.php(83): Hm_Module_Exec->process_result()
#8 /var/www/cypht/lib/modules_exec.php(43): Hm_Module_Exec->run_output_module()
#9 /var/www/cypht/lib/dispatch.php(235): Hm_Module_Exec->run_output_modules()
#10 /var/www/cypht/lib/dispatch.php(190): Hm_Dispatch->process_request()
#11 /var/www/cypht/site/index.php(50): Hm_Dispatch->__construct()
#12 {main}\n  thrown in /var/www/cypht/modules/imap/functions.php on line 914
```

the offending code is:
`for ($i=0;$i<count($struct['disposition']['attachment']);$i++) {`
I assume there are no attachments  tho they are expected. This line has not been updated in the latest current release. I've hacked a quick fix until someone can replicate this and/or patch it:
`for ($i=0;$i<count( is_countable($struct['disposition']['attachment']??null) ? $struct['disposition']['attachment']: [] );$i++) { `

thanks

marclaporte:
@josaphatim What do you think?

josaphatim:
> @josaphatim What do you think?

@mcbmcb0, @marclaporte  I couldn't really reproduce that by as error is clear I'll proposed a merge request

marclaporte:
@mcbmcb0 Please confirm the issue is solved in master.

[...]

PR #759, opened 2023-09-09
Fixed invalid argument passed to foreach

josaphatim:
Relates: https://github.com/cypht-org/cypht/issues/734#issuecomment-1711102910

Assign return of array push to $values lead to invalid passing int to foreach.


mcbmcb0:
from #734, this issue now resolved. thanks

PR #765, opened 2023-09-13
Make cypht search shows latest results when searching in a mail box

mmarcwabo:
## Pullrequest
Make cypht search show latest results

Make default sort mode on search by internal date from the newest to the latest

### Issues
- [X] None

### Checklist

- [X] None

### How2Test
- [X] Make a search from any page and notice that found messages at the top are the recent ones 

### Todo
- [ ] Make sure sort by date choosen by the user works as well


kroky:
Shouldn't we sort by the default field user is seeing messages sorted? Might be internal date, might be date of arrival.

modules/imap/hm-imap.php
```diff
@@ -1090,7 +1090,13 @@ public function search($target='ALL', $uids=false, $terms=array(), $esearch=arra
             if ($esearch_enabled) {
                 $res = $esearch_res;
             }
-            return $this->cache_return_val($res, $cache_command);
+            $found_messages_uids = $this->cache_return_val($res, $cache_command);
```
kroky, line 1093: Are we sure that we get all search results here? If they are paginated, sorting won't work afterwards.
Also, isn't there an IMAP command to search AND sort in one go?

mmarcwabo:
> Shouldn't we sort by the default field user is seeing messages sorted? Might be internal date, might be date of arrival
I have found that internal_date and date are the same. The goal here were to make sure by default newers messages are on top. Some piece of work I am adding is to make sure that sort follows users choices / preferences (Ascending or Descending).

modules/imap/hm-imap.php
mmarcwabo: > Are we sure that we get all search results here? If they are paginated, sorting won't work afterwards. Also, isn't there an IMAP command to search AND sort in one go?

According to my researches, yes: we get all of them, since the IMAP response (`$status`) is "OK" there.
Pagination is being processed when formatting as I observed.

Search and sort in one go with IMAP command: I'll look into this possibility and share my findings, so that you can advise.

kroky:
There are certainly 2 different dates that user can choose for default sort order in imap module - Date of the msg (date sent) and date of arrival (which is probably the internal date). I think we should be in sync here with the setting.

marclaporte:
@mmarcwabo Please take this over the finish line :-)

TheDigitalOrchard:
Just adding my 2¢, Cypht might have an opportunity to do truly smart searches... by doing a "relevance" search and sorting with the most relevant results at the top.

I haven't yet used Cypht in the real world, so this is just a thought off the top of my head.

kroky:
Closing this in favor of https://github.com/cypht-org/cypht/pull/1120 - turns out we needed both front and backend fixes to make sorting work in imap folders.

PR #777, opened 2023-09-30
workflows: add test ci

wangxiaoerYah:
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/.github/tests/hm3.ini
+++ b/.github/tests/hm3.ini
@@ -1,0 +1,552 @@
+
+; -----------------------------------------------------------------------------
+; General settings
+; -----------------------------------------------------------------------------
+
+; Session Type
+; ------------
+; Sessions are how the server maintains your logged in state. Cypht supports
+; standard PHP sessions, as well as it's own completely independent session
+; system that uses memcached or a database for storage. In order to use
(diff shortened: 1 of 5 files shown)


marclaporte:
Thank you @wangxiaoerYah !

@josaphatim will look into it.

kroky:
Looks good to me. Thanks @wangxiaoerYah! 

PHP 8.2 and newer phpunit versions need some fixes in the tests,  we will do that later when we prepare next release of Cypht. For now, maybe you can keep the ci pipeline with php 8.2 commented?

We don't build an image of Cypht, it is released by tagging specific version after we plan a release. If you need to reference a nighly version, just use latest master or a specific commit in the git history of master if you target specific code.

kroky:
Btw, one nice addition here if you happen to work again on the selenium tests integration is copying the hm3.sample.ini to .github/tests and modifying the variables you want (I think only the db connection variables were modified from the default ones). This way, we will safeguard against future changes of hm3.sample.ini.

marclaporte:
Thank you @wangxiaoerYah and @kroky

PR #780, opened 2023-10-02
[NEW] Error message if there is a space at the beginning or end of username or password(login, nux, smtp, imap)

NasserNgandu:
Display a message if the username or password contains a space at the beginning or at the end at login, nux, smtp, imap. Here is the link to the task: https://avan.tech/item10670


modules/smtp/site.js
```diff
@@ -97,6 +97,20 @@ var smtp_delete_action = function(event) {
     );
 };
 
+var smpt_error_if_there_is_leading_or_trailing_space = function(element) {
```
marclaporte, line 100: smpt -> smtp

NasserNgandu: Okay @marclaporte, it's fixed

marclaporte:
Given the task URL above is not public, I will copy some relevant info from it.

Task: Make sure copy pasting a username or password (in a login prompt) with a leading or trailing space doesn't cause a problem

This happens, and leads to support issues. Unless we think there is going to be a password with a space, it should be trimmed, or at least a warning given. (ex.: "You have a leading or trailing space")

Related links:
- https://ux.stackexchange.com/questions/75686/why-do-some-websites-including-google-trim-whitespaces-in-passwords
- https://ux.stackexchange.com/questions/8010/should-i-trim-spaces-in-passwords
- http://stackoverflow.com/questions/7240869/should-i-trim-spaces-in-a-password-field
- http://tonyshowoff.com/articles/should-you-trim-passwords/

modules/smtp/site.js
```diff
@@ -97,6 +97,20 @@ var smtp_delete_action = function(event) {
     );
 };
 
+var error_if_there_is_leading_or_trailing_space = function(element) {
+    if (element.val() !== element.val().trim()) {
+        if (element.attr('name').slice(-4) === 'user') {
```
josaphatim, line 102: @NasserNgandu I think we cannot rely on user in the field name. In nux module username has nux_email as name same as for password.

modules/smtp/site.js
```diff
@@ -184,6 +198,9 @@ if (hm_page_name() === 'servers') {
     $('.save_smtp_connection').on('click', smtp_save_action);
     $('.forget_smtp_connection').on('click', smtp_forget_action);
     $('.delete_smtp_connection').on('click', smtp_delete_action);
+    $('.server_content').on('change', '.credentials, .smtp_password, .nux_password', function() {
```
josaphatim, line 201: What if we do for on paste event ?

modules/core/output_modules.php
```diff
@@ -164,7 +164,28 @@ class Hm_Output_login_end extends Hm_Output_Module {
      * Closes the login form
      */
     protected function output() {
-        return '</form>';
+        $js = '<script type="text/javascript">
```
josaphatim, line 167: I don't get well this. What is the purpose ? Also we have Jquery in the project why not use it here ?

modules/smtp/site.js
josaphatim: According to me this function could receive the text and the type of the input (username or password). Then should check if the text contains leading/trailing space and according to the type build the error message. Then each module which has a username/password (smtp, imap, nux) should add events on their own inputs and call this methods for validation.

Issue #757, opened 2023-09-06
If an email reply will go to a different address than the sender, make it obvious


marclaporte:
![2023-09-06_095256](https://github.com/cypht-org/cypht/assets/1004261/fa28638e-0fc4-44d6-85b2-ccd939d14e94)

Yannick243:
Hello @marclaporte,
can this be closed ?

PR #781, opened 2023-10-02
Display the 'Reply-To' address if it is different from the sender's address

Yannick243:
This PR displays the 'Reply-To' address if it is different from the sender's address.
Related issue: [https://github.com/cypht-org/cypht/issues/757](https://github.com/cypht-org/cypht/issues/757)

Screenshot 2023-10-02 at 14 33 20

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1747,30 +1747,30 @@
     "packages-dev": [
         {
             "name": "doctrine/instantiator",
-            "version": "1.5.0",
+            "version": "2.0.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/instantiator.git",
-                "reference": "0a0fa9780f5d4e507415a065172d26a98d02047b"
+                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/0a0fa9780f5d4e507415a065172d26a98d02047b",
-                "reference": "0a0fa9780f5d4e507415a065172d26a98d02047b",
+                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
+                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
                 "shasum": ""
             },
             "require": {
-                "php": "^7.1 || ^8.0"
+                "php": "^8.1"
             },
             "require-dev": {
-                "doctrine/coding-standard": "^9 || ^11",
+                "doctrine/coding-standard": "^11",
                 "ext-pdo": "*",
                 "ext-phar": "*",
-                "phpbench/phpbench": "^0.16 || ^1",
-                "phpstan/phpstan": "^1.4",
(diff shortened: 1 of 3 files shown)


modules/imap/output_modules.php
```diff
@@ -202,6 +202,13 @@ protected function output() {
                             } catch (Exception $e) {}
                             $txt .= '<tr class="header_'.$fld.'"><th>'.$this->trans($name).'</th><td>'.$this->html_safe($value).'</td></tr>';
                         }
+                        elseif ($fld == 'reply-to') {
+                            if (str_replace(['<','>'], '', $headers['From']) !== $value) {
```
josaphatim, line 206: @Yannick243 according to https://datatracker.ietf.org/doc/html/rfc4021#section-2.1.4 the Reply-To header field can hold multiple emails. Testing it this way will no work then.

Also there is a function in modules/imap/functions.php, process_address_fld which helps to us get label and email from a given header field as the Reply-To field can have format `Some Name <abc@cdf.com>, axy@domain.fr`

composer.lock
```diff
@@ -1186,30 +1186,30 @@
...
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/0a0fa9780f5d4e507415a065172d26a98d02047b",
-                "reference": "0a0fa9780f5d4e507415a065172d26a98d02047b",
+                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
+                "reference": "c6222283fa3f4ac679f8b9ced9a4e23f163e80d0",
                 "shasum": ""
             },
             "require": {
-                "php": "^7.1 || ^8.0"
+                "php": "^8.1"
```
kroky, line 1202: Cypht should work with PHP 7.x as well. I think we need to add platform.php variable in composer.json to avoid upgrades with php8 dropping support for php 7

modules/imap/output_modules.php
```diff
@@ -203,10 +203,16 @@ protected function output() {
...
+                            $from = addr_parse($headers['From']);
+                            
+                            $replyEmails = array_map(function ($addr) {
+                                return $addr['email'];
+                            }, process_address_fld($headers['Reply-To']));
+
+                            if (count($replyEmails) === 1 && ($replyEmails[0] === $from['email'])) {
                                 $txt .= '<tr style="display: none;" class="long_header"><th>'.$this->html_safe($name).'</th><td>'.$this->html_safe($value).'</td></tr>';
+                            } else {
+                                $txt .= '<tr class="header_'.$fld.'"><th>'.$this->trans($name).'</th><td>&lt;'.$this->html_safe(join(',', $replyEmails)).'&gt;</td></tr>';
```
kroky, line 215: Having multiple addresses in angle brackets is against RFC: https://www.rfc-editor.org/rfc/rfc5322.html#section-3.4

composer.lock
Yannick243: I've set the minimum PHP version to 7.3 because the current installed PHPUnit (^9.3.0) requires that version.

modules/imap/output_modules.php
Yannick243: The angle brackets have been removed.
Thanks.

PR #789, opened 2023-10-10
fix imap server deletion sanitizing server IDs

kroky:
These might be alphanumerics and not strictly ints.


marclaporte:
@kroky Good for a baclkport to 1.4.x?

PR #809, opened 2023-10-26
Output email menu content: for just one email account All and Everything removed

Shadow243:
If I only have a single account, I won't require access to 'Everything,' 'All,' or the corresponding box or container. These options are more relevant when dealing with multiple accounts.

### DIFF
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -1299,11 +1299,11 @@ class Hm_Output_main_menu_content extends Hm_Output_Module {
     protected function output() {
         $res = '';
         $email = false;
-        $single = $this->get('single_server_mode');
         if (array_key_exists('email_folders', merge_folder_list_details($this->get('folder_sources', array())))) {
             $email = true;
         }
-        if (!$single) {
+        $total_accounts = count($this->get('imap_servers', array())) + count($this->get('feeds', array()));
+        if ($total_accounts > 2) {
             $res .= '<li class="menu_combined_inbox"><a class="unread_link" href="?page=message_list&amp;list_path=combined_inbox">';
             if (!$this->get('hide_folder_icons')) {
                 $res .= '<i class="bi bi-box2-fill fs-5 me-2"></i>';
@@ -1404,7 +1404,7 @@ class Hm_Output_email_menu_content extends Hm_Output_Module {
                 $res .= '<div style="display: none;" ';
             }
             $res .= 'class="'.$this->html_safe($src).'"><ul class="folders">';
-            if ($name == 'Email' && !$single) {
+            if ($name == 'Email' && count($this->get('imap_servers', array()))  > 1) {
                 $res .= '<li class="menu_email"><a class="unread_link" href="?page=message_list&amp;list_path=email">';
                 if (!$this->get('hide_folder_icons')) {
                     $res .= '<i class="bi bi-globe-americas fs-5 me-2"></i>';
--- a/modules/feeds/modules.php
+++ b/modules/feeds/modules.php
@@ -790,11 +790,14 @@ class Hm_Output_filter_feed_folders extends Hm_Output_Module {
         $res = '';
         $folders = $this->get('feed_folders', array());
         if (is_array($folders) && !empty($folders)) {
-            $res .= '<li class="menu_feeds"><a class="unread_link" href="?page=message_list&amp;list_path=feeds">';
-            if (!$this->get('hide_folder_icons')) {
-                $res .= '<i class="bi bi-rss-fill fs-5 me-2"></i>';
+            if(count($this->get('feeds', array()))  > 1) {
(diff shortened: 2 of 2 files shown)


josaphatim:
> [Related task](https://avan.tech/item64896)

@Shadow243 Please add the description of the is being solved and If possible the link of the issue being solved.

Shadow243:
@josaphatim Requested Description added: [Descrption](https://github.com/cypht-org/cypht/pull/809#issue-1964325521)

marclaporte:
I wonder: What about other types of content beyond email, like RSS feeds?

modules/core/output_modules.php
```diff
@@ -1225,11 +1225,10 @@ class Hm_Output_main_menu_content extends Hm_Output_Module {
     protected function output() {
         $res = '';
         $email = false;
-        $single = $this->get('single_server_mode');
         if (array_key_exists('email_folders', merge_folder_list_details($this->get('folder_sources', array())))) {
             $email = true;
         }
-        if (!$single) {
+        if (count($this->get('imap_servers', array()))  > 1) {
```
kroky, line 1231: Please check carefully what the All and Everything pages actually list. I think they aggregate more than the imap servers, so we need to be careful with the condition here.

josaphatim: @Shadow243 I think jmap_servers, and feeds too as they also appear in everything should also be checked

Shadow243: Oh copy that @josaphatim, let me check it out..

Shadow243: @marclaporte @kroky 
It seems that EveryThing indeed contains several elements including feeds, emails etc...

This condition already applied therefore does not cover this scenario, however for the All menu under Email it works. which amounts to asking for clarification: will the task only cover All or should we also adapt Everything (apply a verification of all the elements concerned)??

kroky: I'd suggest you check if all/everything actually displays more than one source - e.g. 2 imap accounts, 1 imap account and 1 rss feed, 1 imap account and 1 jmap account, etc. If it turns out there is only one source configured to display in All or Everything, then you can hide those links.

Shadow243: Everything displays all configured sources. e.g. if we have 2 imap, 3 jmap, 1 feeds all datas will be visible in Everything.

However All under each menu(Email, Feeds) displays all sources each one based on parent. e.g: All under email displays All Email sources and All under Feeds displays all Feeds sources.

This is what confuses me a little, we will hide All(under Email and Feed) and let Everything as the main menu be displayed or also hide Everything(Main menu) if feeds and imap_servers(jmap_servers) only have one source configured each

According to my understanding, the first analysis is the correct one. An other idea ?

Shadow243: In imap_servers array we have also jmap servers.

Somes changes added.

kroky: We should be consistent. If there is only one email source, All menu item under email is hidden. Same for feeds and any other type of entity. As for 'Everything' - it will also be hidden if there is only one source of any type. In all other cases, it must be shown.

Shadow243: That's what I did yesterday.

modules/core/output_modules.php
```diff
@@ -1404,7 +1404,7 @@ protected function output() {
                 $res .= '<div style="display: none;" ';
             }
             $res .= 'class="'.$this->html_safe($src).'"><ul class="folders">';
-            if ($name == 'Email' && !$single) {
+            if ($name == 'Email' && count($this->get('imap_servers', array()))  > 1) {
```
Shadow243, line 1407: This is for All under Email: checking if there more than one source.

modules/core/output_modules.php
```diff
@@ -1299,11 +1299,11 @@ class Hm_Output_main_menu_content extends Hm_Output_Module {
     protected function output() {
         $res = '';
         $email = false;
-        $single = $this->get('single_server_mode');
         if (array_key_exists('email_folders', merge_folder_list_details($this->get('folder_sources', array())))) {
             $email = true;
         }
-        if (!$single) {
+        $total_accounts = count($this->get('imap_servers', array())) + count($this->get('feeds', array()));
```
Shadow243, line 1305: This is for Everything Menu: check both(Email and Feeds)

kroky: What about jmap_servers and why it is > 2 - if you have 2, everything will not be displayed which seems wrong.

modules/core/output_modules.php
kroky: Ok, just a minor comment and we are good.

modules/core/output_modules.php
Shadow243: I checked jmap servers are not in: jmap_servers but imap_servers

kroky: Ah, yes, forgot about that.

Issue #794, opened 2023-10-18
Plain text shows as attachment


ulfgebhardt:
## 🐛 Bugreport

![image](https://github.com/cypht-org/cypht/assets/1238238/d1ae0b93-1e55-4c38-aa80-2b65f58ad062)

When sending plain text emails to oneself the text is shown as attachment while it is just the message itself.
This is a regression compared to before (2021).

This is how those emails looked before

![image](https://github.com/cypht-org/cypht/assets/1238238/95eb1690-6e79-44f5-8198-7f783c11fb7c)
(this was sent for a cypht version of 2021)

### Version & Environment

Rev: https://github.com/cypht-org/cypht/commit/a5285756aff89bf1ee85cffa7fea2a6fd9b132ce

OS:  [chrome-latest] - more details where this runs if needed

### Steps to reproduce
1. Send plain text email to oneself
2. receive email attachment symbol
3. wonder where the attachment is
4. ...
5. Profit

marclaporte:
Thank you @ulfgebhardt for taking the time to compare with a 2021 version to confirm the change of behavior.

marclaporte:
Could this be related to this? https://github.com/cypht-org/cypht/pull/735

kroky:
I think it is related to https://github.com/cypht-org/cypht/pull/661

ulfgebhardt:
This is the commit I have checked out where everything is still fine. The change must have been introduced after this:
https://github.com/cypht-org/cypht/commit/02ddf0d10e4fc304e8e07dad8700bf33a739071d

marclaporte:
@ulfgebhardt  Can you confirm that this fixed it? https://github.com/cypht-org/cypht/pull/815

PR #815, opened 2023-10-30
Fixed Plain text showing as attachment

josaphatim:
Fixes https://github.com/cypht-org/cypht/issues/794

Using php explode function on empty string creates an array with 1 key. Which makes the attachment array to have at least one key.


ulfgebhardt (approved):
With this branch I get the old view when writing new EMails - old emails are still the same - I assume this is correct.

Old EMail with this branch:
image

New EMail with this branch:
image

Hence I'd say its fixed.

Issue #813, opened 2023-10-30
Allow database authentication using `pg_service.conf`


ivanov17:
## 🚀 Feature

Hello! Unfortunately, it is currently not possible to authenticate to a PostgreSQL database using the credentials stored in the [pg_service.conf](https://www.postgresql.org/docs/current/libpq-pgservice.html) file. However, this is a more secure method than storing credentials in the application configuration, even if it is not located at the root of the web server. 

As far as I can see, the application currently requires storing database credentials in the application configuration.

https://github.com/cypht-org/cypht/blob/3579a081b544179bce26a7cacf0fce51ad2f9dbf/lib/db.php#L18

This leads to the fact that despite a successful IMAP login, I can't to login to application. 

```php
    [0] => Using Hm_DB_Session with Hm_Auth_IMAP
    [1] => Using DB user configuration
    [2] => Already registered module for message re-attempted: imap_show_message
    [3] => Already registered module for message re-attempted: imap_message_list_type
    [4] => Already registered module for message re-attempted: imap_server_ids
    [5] => Using sapi: cli-server
    [6] => Request type: HTTP
    [7] => Request path: /
    [8] => TLS request: 0
    [9] => Mobile request: 0
    [10] => Page ID: home
    [11] => Using Hm_Cache for cache
    [12] => CACHE backend using: noop
    [13] => Setting cookie: name: hm_id, lifetime: 0, path: /, domain: mail.anarhist.net, secure: , html_only 1
    [14] => Missing configuration setting for db_user
    [15] => Missing configuration setting for db_pass
    [16] => Missing configuration setting for db_name
    [17] => Missing configuration setting for db_host
    [18] => SQLSTATE[08006] [7] could not match 1 host names to 2 hostaddr values
    [19] => Setting cookie: name: hm_msgs, lifetime: 0, path: /, domain: mail.anarhist.net, secure: , html_only 1
    [20] => Redirecting to /
    [21] => PHP version 8.2.11
    [22] => Zend version 4.2.11
    [23] => Peak Memory: 2048
    [24] => PID: 834
    [25] => Included files: 89
```

At the same time, on the Cypht login page I see the message `Invalid username or password`, which is incorrect because the IMAP server logs indicate the opposite.

### Design, Layout, Architecture

I think making the `db_user`, `db_pass`, `db_name` and `db_host` parameters optional would be sufficient. Or maybe the application should try to connect to the database without using any parameters, and only if there is no connection should it try to use the credentials from the configuration.

[...]

PR #823, opened 2023-11-10
Switch cypht config hm3.* to dotenv

Shadow243:
### Task Description: Integrate Symfony Dotenv Package into Cypht Project

**Details:**
In response to [GitHub Issue #813](https://github.com/cypht-org/cypht/issues/813), the Symfony Dotenv package has been added to the Cypht project. This integration is part of an effort to standardize the handling of environment variables across different projects, including Cypht, Tiki, and Tiki Manager.

**Next Steps:**
1. Confirm the successful integration of the Symfony Dotenv package into the Cypht project.
2. Verify that environment variables are appropriately managed and utilized within the project.
3. Consider extending this standardized approach to other related projects like Tiki for better alignment.


Shadow243:
should i perhaps call 

```
$environment->load();//load env files
```
 in index.php ? after this:
 ```
 /* get configuration */
$config = new Hm_Site_Config_File(CONFIG_FILE);

/* setup ini settings */
if (!$config->get('disable_ini_settings')) {
    require APP_PATH.'lib/ini_set.php';
}
```

**Comment by kroky** [comment]
The https://github.com/cypht-org/cypht/issues/813 requests the ability to load config variables through env variables. We don't need dotenv package or .env file to do this. We just need to look at the ENV array or getenv function first before falling back to hm3.ini ones loaded in site config.
I don't recommend we do a partial switch to dotenv. We should either replace hm3.ini and all related ini file functionality with dotenv (which also includes unit tests, Tiki integration, config generation and probably other aspecs of the app) and be sure to keep the existing verbose comments in that config file OR keep the old ini file with db credentials and allow some of the variables to be overridden by the env. First approach is a big refactor while second one only addresses the issue request.

**Comment by kroky** [comment]
Wow, quite a refactor! I like the approach but can be streamline 2 issues here:
1. The repeated calls to merge_config_files with params passing the config folder location - no need to repeat all this. I think we should simply parse the config folder location anywhere Hm_Config_... is initialized or used  - in fact, it can be part of this class initialization process.
2. All module .ini files should follow the same syntax as we use now for app config - either .php config files or .env files.

**Comment by Shadow243** [comment]
This commit addresses two key issues in the current codebase to enhance the configuration handling process. The proposed changes aim to improve code organization, eliminate redundancy, and promote consistency in the configuration structure.

## Changes

### Issue 1: Duplicated Config Folder Location

The repeated calls to `merge_config_files` with manually passed configuration folder locations are removed. Instead, the refactor suggests parsing the config folder location within the initialization process of the `Hm_Config_...` class. This approach centralizes the configuration folder handling logic, improving code readability and maintainability.

### Issue 2: Standardizing Module .ini Files

All module `.ini` files are refactored to follow the same syntax used for app configuration. This involves adopting either `.php` config files or `.env` files for module configurations, ensuring a consistent structure throughout the application.

### Constructor Parameter for PHPUnit Tests

In the `Hm_Site_Config_File` class, a constructor parameter has been added:

```php
public function __construct($all_configs = []) {
    $this->load(empty($all_configs) ? merge_config_files(APP_PATH.'config') : $all_configs, false);
}
```
This parameter allows for handling PHPUnit tests that use custom config files other than those found in config/*.php. It provides flexibility for testing scenarios where alternative configurations are necessary.

lib/framework.php
```diff
@@ -8,6 +8,7 @@
 
 define('VERSION', .1);
 
+require_once VENDOR_PATH.'composer/autoload_classmap.php';
```
kroky, line 11: Individual modules require VENDOR_PATH/autoload.php when needed. If you need it here in order to use dotenv package, please require the autoload and not the autoload_classmap file.

kroky: Also, I think you should remove the require blocks for vendor/autoload in modules.

lib/framework.php
```diff
@@ -38,6 +40,11 @@
     require VENDOR_PATH.'paragonie/random_compat/lib/random.php';
 }
 
+/* load env files */
+require_once VENDOR_PATH.'symfony/dotenv/Dotenv.php';
```
kroky, line 44: If you require the vendor/autoload file, you won't need to require individual vendor package classes before using them.

scripts/config_gen.php
```diff
@@ -115,11 +115,11 @@ function build_config() {
     }
 
     /* get the site settings */
-    $settings = parse_ini_file(APP_PATH.'hm3.ini');
+    $settings = merge_config_files(APP_PATH.'config');
 
     if (is_array($settings) && !empty($settings)) {
         $settings['version'] = VERSION;
-        $settings = parse_module_ini_files($settings);
+        // $settings = parse_module_ini_files($settings);
```
kroky, line 122: Please remove the commented code.

scripts/config_gen.php
```diff
@@ -304,11 +301,6 @@ function write_config_file($settings, $filters) {
     Hm_Output_Modules::try_queued_modules();
     Hm_Output_Modules::process_all_page_queue();
     Hm_Output_Modules::try_queued_modules();
-    $settings['handler_modules'] = Hm_Handler_Modules::dump();
-    $settings['output_modules'] = Hm_Output_Modules::dump();
-    $settings['input_filters'] = $filters;
```
kroky, line 309: Previous hm3.rc json file had these 3 arrays - handler_modules, output_modules and input_filters generated automatically from config_gen. Now we have dynamic config, how do we generate these 3 arrays? I see lib/modules_exec.php setup_production_modules method uses the 3 arrays from config but I am not sure we have them there with the new setup. Maybe generate another file in config folder with these 3 arrays when doing config_gen?

tests/phpunit/config.php
```diff
@@ -171,7 +171,7 @@ public function setUp(): void {
      * @runInSeparateProcess
      */
     public function test_get_modules() {
-        $config = new Hm_Site_Config_File('./data/siteconfig.rc');
```
kroky, line 174: I think you should also remove that file if it is no longer used.

lib/framework.php
Shadow243: Most definitely, the line removed

scripts/config_gen.php
Shadow243: commented code removed

scripts/config_gen.php
Shadow243: a dynamic file dynamic.php file will now be created when doing config_gen.

tests/phpunit/config.php
Shadow243: Done.

scripts/config_gen.php
kroky: Alright. However, I'd prefer it to be in the .gitignore list, so it doesn't end up in the repo.

lib/framework.php
Shadow243: Copy that @kroky,

They have been removed in the latest commit.

kroky:
Looks good to me code-wise. Can maybe @josaphatim or somebody from the team test it a bit more - enable/disable some modules, regenerate config, change config/.env settings, etc.

Also, @Shadow243 once this is merged, the first thing we should do with Tiki integration master branch is get the latest Cypht and update the integration to use the latest .env setup.

.env.dist
```diff
@@ -0,0 +1,178 @@
...
+
+SESSION_TYPE=PHP
+AUTH_TYPE=DB
+LDAP_AUTH_SERVER=localhost
+LDAP_AUTH_PORT=389
+LDAP_AUTH_TLS=
+LDAP_AUTH_BASE_DN=example,dc=com
+
+IMAP_AUTH_NAME=Gandi
+IMAP_AUTH_SERVER=mail.gandi.net
```
josaphatim, line 20: @Shadow243 please refer to hm3.sample.ini for defaults values.

.env.dist
```diff
@@ -0,0 +1,178 @@
...
+IMAP_AUTH_SIEVE_CONF_HOST=tls://mail.gandi.net:4190
+
+DEFAULT_SMTP_NAME=
+DEFAULT_SMTP_SERVER=
+DEFAULT_SMTP_PORT=
+DEFAULT_SMTP_TLS=
+DEFAULT_SMTP_NO_AUTH=
+
+USER_CONFIG_TYPE=file
+USER_SETTINGS_DIR=/Users/shadow243/Documents/var/lib/hm3/users
```
josaphatim, line 32: Do it for all settings in this file please.

josaphatim:
@kroky I tested this and It is working. I enabled/disabled modules and also ran php unit tests.

@Shadow243 Just one more thing, there was a hm3.sample.ini file and hm3.ini that we used for our own configurations and present in .gitignore file. But in this new architecture there is only .env.dist. I think we should have a .env.example and then change the INSTALL file according to this.

Issue #216, opened 2017-08-09
Add drag/drop within the same account


Yamakasi:
Would it be an idea to add de possibillity to drag/drop messages within the same account between folders ?

marclaporte:
I hope we can support drag and drop messages even between accounts

marclaporte:
@jasonmunro Are you OK if we use https://github.com/SortableJS/Sortable ?  This is our usual tool in Tiki.

Thanks!

dumblob:
Or maybe one of the following:

https://github.com/bevacqua/dragula
https://github.com/mindplay-dk/zero-drag

andonov85:
It looks like moving data between lists, so [](https://github.com/SortableJS/Sortable) looks appropriate to me.

marclaporte:
https://github.com/mindplay-dk/zero-drag/commits/master

Last commit 2018. IMO, that disqualifies it.

marclaporte:
https://github.com/SortableJS/Sortable/blob/master/Sortable.js is indeed bigger than https://github.com/bevacqua/dragula/blob/master/dist/dragula.js

Last commit in 2020-09  (15 months is not a disqualifier but a concern)
https://github.com/bevacqua/dragula/commits/master


Another topic: should we use minified version?

- https://github.com/SortableJS/Sortable/blob/master/Sortable.min.js (43 KB)
- https://github.com/bevacqua/dragula/blob/master/dist/dragula.min.js (13.6 KB)

marclaporte:
Here is what Sortable says: (not an objective point of view)
https://vimeo.com/311584137

marclaporte:
Using the data from:

- https://github.com/SortableJS/Sortable/
- https://github.com/bevacqua/dragula/
- https://www.openhub.net/p/SortableJS
- https://www.openhub.net/p/Dragula
- https://www.openhub.net/p/_compare?project_0=SortableJS&project_1=Dragula 

Both are very popular in terms of watches, forks and stars.

SortableJS: pull requests: 14 open vs 456 closed
Dragula: pull requests: 51 open vs 134 closed

SortableJS: issues: 338 open vs 1262 closed
Dragula: issues: 97 open vs 406 closed

In general, for pull requests and issue management, SortableJS has more activity and is better maintained.

SortableJS has had 99 committers vs 33 for Dragula.

One more factor: we have a front-end developer (@andonov85) who knows SortableJS well, and can support it.

So IMO SortableJS wins here, even if heavier.

[...]

PR #834, opened 2023-12-03
[NEW] Add drag-and-drop feature to move messages between folders

jacob-js:
## Pullrequest
Introduce drag-and-drop of messages between folders
### Issues

- [X] https://github.com/cypht-org/cypht/issues/216

### Demo
[drag-and-drop-demo.webm](https://github.com/cypht-org/cypht/assets/68013195/511710bc-a131-455f-907c-c27ddddb6b96)


modules/core/site.js
```diff
@@ -1857,3 +1857,82 @@ function listControlsMenu() {
...
+        if (row.querySelector('.checkbox_cell input[type=checkbox]:checked')) {
+            selectedRows.push(row);
+        }
+    });
+    
+    if (selectedRows.length == 0) {
+        selectedRows.push(movingElement);
+    }
+
+    const movingIds = selectedRows.map(row => row.className.split(' ')[0]);
```
josaphatim, line 1891: There is an issue with this line. Try to drap and release the click you will see the issue.
Uncaught TypeError: Cannot read properties of undefined (reading 'className')

josaphatim:
@jacob-js I noticed that dragging to **manage folders** link also starts the ajax request which shouldn't. Also the row moved is deleted but I think the current page should be reloaded instead.

marclaporte:
@jacob-js Does it work with a touch interface (tablet or phone)

jacob-js:
> @jacob-js Does it work with a touch interface (tablet or phone)

While it ideally should, the app interface currently lacks the capability to display both the message list and folders simultaneously on mobile and tablet devices. Consequently, simulating the drag-and-drop functionality on these devices isn't feasible.

jacob-js:
> @jacob-js I noticed that dragging to **manage folders** link also starts the ajax request which shouldn't. Also the row moved is deleted but I think the current page should be reloaded instead.

@josaphatim

1. I'm not sure if I got you clearly, are you suggesting that the messages should move between folders without triggering any AJAX request? If so, how could you do that?
2. Yes of course the row moved is deleted in the interface, I certainly agree that we could reload the page and fetch the entire list of messages, but this seems to me inappropriate because we just need to reflect the changes on the list without disrupting the user from what they are doing.

modules/core/site.js
jacob-js: It seems to be impossible for me to reproduce this error. Any specific indication?

jacob-js:
> @jacob-js I noticed that dragging to **manage folders** link also starts the ajax request which shouldn't. Also the row moved is deleted but I think the current page should be reloaded instead.

@josaphatim 

I got it, I added a check to allow only valid folders.

jacob-js:
> > @jacob-js Does it work with a touch interface (tablet or phone)
> 
> While it ideally should, the app interface currently lacks the capability to display both the message list and folders simultaneously on mobile and tablet devices. Consequently, simulating the drag-and-drop functionality on these devices isn't feasible.

@marclaporte here is what I'm talking about in visual.
image
image

marclaporte:
> > > @jacob-js Does it work with a touch interface (tablet or phone)
> > 
> > 
> > While it ideally should, the app interface currently lacks the capability to display both the message list and folders simultaneously on mobile and tablet devices. Consequently, simulating the drag-and-drop functionality on these devices isn't feasible.
> 
> @marclaporte here is what I'm talking about in visual.


Yes, I understand. Some laptops have a touch screen, but it's not so common.

josaphatim:
Now the whole is working as expected, no error messages in the console.
But can you fix this test please: test_page_js_debug ?

Issue #684, opened 2023-02-26
Warn about potential missing attachment when sending an email


marclaporte:
Something like this: https://www.claws-mail.org/plugin.php?plugin=attachwarner

marclaporte:
AI: Seems like overkill to add a dependency for this.

A simple example: https://github.com/roundcube/roundcubemail/blob/master/plugins/attachment_reminder/attachment_reminder.php

We can easily get translators to identify the main terms for currently supported languages.

I like the idea of machine learning and AI for more advanced use cases. Related: https://dev.tiki.org/Machine-Learning-for-Email

Shadow243:
Pull Request https://github.com/cypht-org/cypht/pull/839 has been submitted, introducing the feature to warn users about potential missing attachments when composing emails.

PR #839, opened 2023-12-11
Fix Warn about potential missing attachment when sending an email

Shadow243:
### Description
Introduce a feature to warn users about potential missing attachments when composing and sending an email. This includes a configuration setting to toggle the feature, JavaScript implementation for keyword checking, modal notification, and an option to force send.

### Implementation Steps
1. **Configuration Setting**
   - [ ] Introduce `enable_attachment_reminder` configuration value.

2. **JavaScript Implementation**
   - [ ] Implement keyword checking in the email body.
   - [ ] Store keywords in the `translated_keywords` variable.

3. **Modal Notification**
   - [ ] Display a modal notification when keywords suggest a missing attachment.
   - [ ] Include information about the potential missing attachment.

4. **Force Send Option**
   - [ ] Provide a "Send Anyway" button within the modal.

5. **Testing**
   - [ ] Test the feature in various scenarios.
   - [ ] Ensure the modal appears only when necessary.

### Additional Information
- **Issue:** #684

### DIFF
--- a/language/ru.php
+++ b/language/ru.php
@@ -579,6 +579,8 @@ return array(
     'Sieve server capabilities' => false,
     'Connection To Sieve Server Failed' => false,
     'Warn for unsaved changes' => false,
+    'We couldn\'t find the attachment you referred to. Please confirm if you attached it or provide the details again.' => false,
+    'attachment,file,attach,attached,attaching,enclosed,CV,cover letter' => false
 ); 
 
 ?>
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -595,9 +595,9 @@ class Hm_Output_js_data extends Hm_Output_Module {
             'window.hm_current_lang = ' . json_encode($this->lang) . ';' .
             'window.hm_translations = ' . json_encode($this->all_trans()) . ';' .
             'var hm_trans = function(key, lang = window.hm_current_lang) {' .
-            '    const translation = window.translations[lang][key];' .
-            '    if (translation !== undefined && translation !== false) {' .
-            '        return translation;' .
+            '    const langTranslations = window.translations && window.translations[lang];' .
+            '    if (langTranslations && langTranslations[key] !== undefined && langTranslations[key] !== false) {' .
+            '        return langTranslations[key];' .
             '    }' .
             '    return key;' .
             '};' .
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -592,6 +592,15 @@ class Hm_Output_js_data extends Hm_Output_Module {
     protected function output() {
         $res = '<script type="text/javascript">'.
             'var globals = {};'.
+            'window.hm_current_lang = ' . json_encode($this->lang) . ';' .
+            'window.hm_translations = ' . json_encode($this->all_trans()) . ';' .
+            'var hm_trans = function(key, lang = window.hm_current_lang) {' .
+            '    const translation = window.translations[lang][key];' .
+            '    if (translation !== undefined && translation !== false) {' .
+            '        return translation;' .
+            '    }' .
+            '    return key;' .
+            '};' .
             'var hm_empty_folder = function() { return "'.$this->trans('So alone').'"; };'.
             'var hm_mobile = function() { return '.($this->get('is_mobile') ? '1' : '0').'; };'.
             'var hm_debug = function() { return "'.(DEBUG_MODE ? '1' : '0').'"; };'.
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -646,8 +646,11 @@ if (!hm_exists('get_mime_type')) {
 class Hm_Handler_process_compose_form_submit extends Hm_Handler_Module {
     public function process() {       
         /* not sending */
-        if (!array_key_exists('compose_smtp_id', $this->request->post)) {
+        if (!array_key_exists('smtp_send', $this->request->post)) {
             $this->out('enable_attachment_reminder', $this->user_config->get('enable_attachment_reminder_setting', false));
+            return;
+        }
+        if (!array_key_exists('compose_smtp_id', $this->request->post)) {
             return;
         }
 
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -646,7 +646,8 @@ if (!hm_exists('get_mime_type')) {
 class Hm_Handler_process_compose_form_submit extends Hm_Handler_Module {
     public function process() {       
         /* not sending */
-        if (!array_key_exists("compose_smtp_id", $this->request->post)) {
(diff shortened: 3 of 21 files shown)


modules/smtp/site.js
```diff
@@ -382,14 +382,38 @@ var is_valid_recipient = function(recipient) {
     return recipient.match(valid_regex);
 };
 
-$(function() {    
+var show_cypht_reminder_dialog = function () {
```
josaphatim, line 385: @Shadow243 can you just use confirm method instead of creating a modal here ? Confirm is used elsewhere in Cypht.

language/ru.php
```diff
@@ -577,7 +577,11 @@
     'Archive to the original folder' => false,
     'Move To Blocked Folder' => false,
     'Sieve server capabilities' => false,
-    'Connection To Sieve Server Failed' => false
+    'Connection To Sieve Server Failed' => false,
+    'Missing attachment?' => 'Забыли вложение?',
+    'Did you forget to attach a file?' => 'Вы не забыли прикрепить файл?',
```
josaphatim, line 582: This translation is not used anywhere

language/ru.php
```diff
@@ -577,7 +577,11 @@
...
+    'Remind about forgotten attachments' => 'Напоминать о забытых вложениях',
```
josaphatim, line 583: This too

modules/smtp/site.js
Shadow243: @josaphatim  confirm usage added.

language/ru.php
Shadow243: You are right @josaphatim , I still wandering if there is no way to use trans in site.js, I would prefer to keep:
``` 
'We couldn\'t find the attachment you referred to. Please confirm if you attached it or provide the details again.' => false 
```
in lang file instead.

language/ru.php
Shadow243: Removed

modules/smtp/modules.php
```diff
@@ -668,7 +669,7 @@ public function process() {
             'draft_subject' => $form['compose_subject'],
             'draft_smtp' => $smtp_id
         );
-
+
```
kroky, line 672: Whitespace...

modules/smtp/setup.php
```diff
@@ -135,6 +138,7 @@
         'profile_value' => array(FILTER_SANITIZE_FULL_SPECIAL_CHARS, false),
         'msg_sent_and_archived' => array(FILTER_VALIDATE_BOOLEAN, false),
         'sent_msg_id' => array(FILTER_VALIDATE_BOOLEAN, false),
+        // 'enable_attachment_reminder' => array(FILTER_VALIDATE_BOOLEAN, false),
```
kroky, line 141: Why comment?

modules/smtp/site.js
```diff
@@ -406,20 +444,45 @@ $(function() {
...
-                $('.compose_next_email_data').val(next_message);
+        $('.compose_form').on('submit', function(e) {
+            const compose_body_value = document.getElementById('compose_body').value;
+            const force_send = document.getElementById('force_send')?.value;
+
+            var reminder_value = $('.compose_form').data('reminder');
+
+            if (reminder_value === 1) {
+                const uploaded_files = $("input[name='uploaded_files[]']").map(function () { return $(this).val(); }).get();
+                const translated_keywords = 'attachment,file,attach,attached,attaching,enclosed,CV,cover letter' + //en
```
kroky, line 455: It would be better to make this list translatable and put into translation arrays for each language.

language/ru.php
kroky: Yes, let's make it possible to use trans function in JS context as well. This would entail:
- output module (in core module probably) to send existing language translations as array to the DOM (through a script tag and json output)
- having a global or window js variable with translations will make writing a trans function in JS quite easy
- use that trans everywhere you need to translate a string in JS context.

This will be useful for other js files that have static strings and need translation.

modules/smtp/modules.php
```diff
@@ -1810,6 +1840,15 @@ function get_outbound_msg_detail($post, $draft, $body_type) {
     }
     return array($body, $cc, $bcc, $in_reply_to, $draft);
 }}
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('attachment_reminder_dialog')) {
```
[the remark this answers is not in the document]
Shadow243: Removed.

modules/smtp/modules.php
Shadow243: Removed.

modules/smtp/site.js
Shadow243: Copy that @kroky, Refactor Done.

language/ru.php
Shadow243:
```
window.transations 
```
and 
```
window.current_lang
```
added with trans() function available in js

Shadow243:
We have a conflict with [https://github.com/cypht-org/cypht/pull/868](https://github.com/cypht-org/cypht/pull/868), I'l have to review what changes from Jacob PR.

But all requested updates from @josaphatim & @kroky are Done.

modules/core/output_modules.php
```diff
@@ -544,6 +543,15 @@ class Hm_Output_js_data extends Hm_Output_Module {
     protected function output() {
         $res = '<script type="text/javascript">'.
             'var globals = {};'.
+            'window.current_lang = ' . json_encode($this->lang) . ';' .
+            'window.translations = ' . json_encode($this->all_trans()) . ';' .
+            'var trans = function(key, lang = window.current_lang) {' .
```
kroky, line 548: Can we continue with the naming convention hm_? E.g. hm_trans, hm_translations, hm_current_lang?

modules/core/output_modules.php
```diff
@@ -544,6 +543,15 @@ class Hm_Output_js_data extends Hm_Output_Module {
     protected function output() {
         $res = '<script type="text/javascript">'.
             'var globals = {};'.
+            'window.current_lang = ' . json_encode($this->lang) . ';' .
+            'window.translations = ' . json_encode($this->all_trans()) . ';' .
```
kroky, line 547: No need to pass all the translations from all the languages here. Just pass the current language translations to the JS context.

Shadow243: Copy that @kroky , Considering users can write emails in any language, even different from the default, and our need to detect attachments in testing across languages, how would you suggest optimizing the process without losing this flexibility?

kroky: Yes, that's a good point but will probably make the check quite slower if we keep it like that. Hmm, I don't see another option, however. OK, let's keep it with all languages for now.

modules/core/output_modules.php
Shadow243: Done too.

PR #840, opened 2023-12-12
Added missing special folders types

josaphatim:
Some special folder types where missing.

### DIFF
--- a/modules/imap_folders/modules.php
+++ b/modules/imap_folders/modules.php
@@ -138,15 +138,25 @@ class Hm_Handler_process_accept_special_folders extends Hm_Handler_Module {
                 Hm_Msgs('ERRUnable to connect to the selected IMAP server');
                 return;
             }
-
             $specials = $this->user_config->get('special_imap_folders', array());
-            if ($exposed = $imap->get_special_use_mailboxes()) {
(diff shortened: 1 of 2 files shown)


modules/imap_folders/modules.php
```diff
@@ -138,15 +138,15 @@ public function process() {
                 Hm_Msgs('ERRUnable to connect to the selected IMAP server');
                 return;
             }
-
             $specials = $this->user_config->get('special_imap_folders', array());
-            if ($exposed = $imap->get_special_use_mailboxes()) {
-                $specials[$form['imap_server_id']] = array('sent' => $exposed['sent'], 'draft' => '', 'trash' => $exposed['trash'], 'archive' => '', 'junk' => '');
-            } else if ($form['imap_service_name'] == 'gandi') {
-                $specials[$form['imap_server_id']] = array('sent' => 'Sent', 'draft' => 'Drafts', 'trash' => 'Trash', 'archive' => 'Archive', 'junk' => 'Junk');
```
kroky, line 146: Seems like we had a special fix for gandi - maybe their expose list is empty and we need to hardcode the array? This PR removes the functionality - I wonder if it is on purpose?

josaphatim: @kroky right. Their expose list is empty. I restored this functionnality

Issue #660, opened 2022-12-13
Add new contacts source


Liia-Dulher:
Is there any way to add contacts from my database? I would like to use it as new source or replace local emails with it. Also I can not find where local contacts are stored and how they are loaded.

kambereBr:
Hello @Liia-Dulher, here's a pending Pull Request (https://github.com/cypht-org/cypht/pull/845) that introduces CSV import functionality to the existing local contacts module. Before finalizing it, I'd love to get your feedback and suggestions. The pull request is still a draft, so your insights at this stage would be valuable.

To give you a better idea of how it works, I've attached a demo GIF showcasing the implemented feature in action. Have a look when you get a chance

marclaporte:
"Persistent data stored between logins is encrypted with a key derived from your clear text password, which is obviously not stored anywhere"
Source: https://www.cypht.org/security.html

marclaporte:
@Liia-Dulher coming soon: https://github.com/cypht-org/cypht/pull/904

kambereBr:
@Liia-Dulher the Pull Request https://github.com/cypht-org/cypht/pull/904, is already merged.

PR #845, opened 2023-12-16
Add CSV import/export functionality to the contacts module

kambereBr:
### Pull Request

This pull request introduces CSV import/export functionality to the existing contacts module. It enables users to import/export contact information in CSV format, enhancing the contact management capabilities 

### Issues
https://github.com/cypht-org/cypht/issues/660

### Demo

cypht_import_export

### DIFF
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -635,7 +635,7 @@ class Hm_Output_js_data extends Hm_Output_Module {
             '        return langTranslations[key];'.
             '    }'.
             '    return key;'.
-            '};'.
+            '};';
         $res .= '</script>';
         return $res;
     }


modules/contacts/modules.php
```diff
@@ -106,7 +131,17 @@ protected function output() {
  */
 class Hm_Output_contacts_content_start extends Hm_Output_Module {
     protected function output() {
-        return '<div class="contacts_content"><div class="content_title">'.$this->trans('Contacts').'</div>';
+        $contact_source_list = $this->get('contact_sources', array());
+        $actions = '<div class="src_title">'.$this->html_safe('Export Contacts as CSV').'</div>';
+        $actions .= '<div class="list_src"><a href="?page=export_contact&amp;contact_source=all">'.$this->html_safe('All Contacts').'</a></div>';
```
kroky, line 136: No need to html_safe a static string. However, you need to translate it.

modules/contacts/modules.php
```diff
@@ -106,7 +131,17 @@ protected function output() {
...
+        foreach ($contact_source_list as $value) {
+            $actions .= '<div class="list_src"><a href="?page=export_contact&amp;contact_source='.$this->html_safe($value).'">'.$this->html_safe($this->html_safe($value).' Contacts').'</a></div>';
```
kroky, line 138: html_safe only needed on the value variable. Please translate the static strings.

modules/local_contacts/modules.php
```diff
@@ -173,3 +229,27 @@ protected function output() {
...
+
+/**
+ * @subpackage import_local_contacts/output
+ */
+class Hm_Output_import_contacts_form extends Hm_Output_Module {
+    protected function output() {
+
+        $form_class = 'contact_form';
+        $button = '<input class="add_contact_submit" type="submit" name="add_contact" value="'.$this->trans('Add').'" />';
+        $notice = 'Please ensure your CSV header file follows the format: display_name, email_address, phone_number';
```
kroky, line 241: Probably better idea is to provide a sample csv file. You can put a static one in a web-accessible URL. Something as small as:
```
display_name, email_address, phone_number
Thomas Tester, test@example.org, 1234567890
```
and then have a link to download a sample csv file.

modules/local_contacts/modules.php
```diff
@@ -173,3 +229,27 @@ protected function output() {
...
+        $title = $this->trans('Add from CSV');
```
kroky, line 242: Import from CSV file

modules/local_contacts/modules.php
```diff
@@ -68,6 +68,62 @@ public function process() {
             $this->session->record_unsaved('Contact Added');
             Hm_Msgs::add('Contact Added');
         }
+
+        list($success, $form) = $this->process_form(array('contact_source', 'add_contact'));
```
kroky, line 72: Form needs to be processed only once and code above already does it. However, I think the functionality itself here is worth adding another handler class and method to process the form. Please separate it from the original handler method for adding a single contact.

modules/local_contacts/modules.php
```diff
@@ -68,6 +68,62 @@ public function process() {
...
+                    if (!$contactUpdated) {
+                        $contacts->add_contact($details);
+                        $message .= 'Contact Added: '.$email.'\n';
+                    }
+                }
+                fclose($csv);
+                $this->user_config->set('contacts', $contacts->export());
+                $this->session->record_unsaved('Contact Added');
+                if ($message) {
+                    Hm_Msgs::add($message);
```
kroky, line 123: Have you tested importing 100 or 1000 contacts? I think this message might not fit in the hm messages elegantly in that case. Maybe it is worth adding a separate output module to display the results of the import operation - one line per the input file with a success mark/message (e.g. created or updated) or error message if the line didn't import.

modules/contacts/modules.php
```diff
@@ -84,6 +84,31 @@ public function process() {
...
+class Hm_Handler_process_export_contacts extends Hm_Handler_Module {
+    public function process() {
+        if (array_key_exists('contact_source', $this->request->get)) {
+            $source = $this->request->get['contact_source'];
+            $contact_list = $this->user_config->get('contacts', array());
+            if ($source != 'all') {
+                $contact_list = array_filter($contact_list, function($v) { return $v['source'] == $this->request->get['contact_source']; });
+            }
+
+            header('Content-Type: text/csv');
```
kroky, line 99: Let's use the Cypht function Hm_Functions::header

modules/contacts/modules.php
```diff
@@ -84,6 +84,31 @@ public function process() {
...
+            header('Content-Disposition: attachment; filename="'.$source.'_contacts.csv"');
+            $output = fopen('php://output', 'w');
+            fputcsv($output, array('display_name', 'email_address', 'phone_number'));
+            foreach ($contact_list as $contact) {
+                fputcsv($output, array($contact['display_name'], $contact['email_address'], $contact['phone_number']));
+            }
+            fclose($output);
+            exit;
```
kroky, line 107: I think we should be more creative here and disallow direct output from handler modules and stopping the whole application in the middle of execution. It violates the purpose of these modules and also prevents us from properly debugging in the future. It also skips sometimes other important application code from executing if you die/exit early.

If you check lib/dispatch.php you will see that normal requests are dispatched via Hm_Dispatch->process_request which calls render_output at the end. It assembles a Hm_Output_HTTP class which renders the headers and then sends the headers to the browser. This is exactly what we want. So, instead of doing it in the handler, you can generate the csv as string (or store as array of values), pass that to output modules, have a special output module that formats the array as csv and renders that as a single output content. Then, the Hm_Output_HTTP class will handle the rest. So, it is important that the new output module render only the csv string, pass the needed headers and prevent any other output from being displayed. We might need to tweak Hm_Output_Module classes to allow a particular child module to reset the output and replace with a particular string that will be the only output sent to the browser. This will be useful when we want to stream contents of files, download other types of content from within cypht itself as some future tasks.

modules/local_contacts/modules.php
```diff
@@ -173,3 +238,31 @@ protected function output() {
...
+ * @subpackage import_local_contacts/output
+ */
+class Hm_Output_import_contacts_form extends Hm_Output_Module {
+    protected function output() {
+        $form_class = 'contact_form';
+        $button = '<input class="add_contact_submit" type="submit" name="import_contact" value="'.$this->trans('Add').'" />';
+        $notice = 'Please ensure your CSV header file follows the format: display_name, email_address, phone_number';
+        $title = $this->trans('Import from CSV file');
+
+        $sample_csv_data = "display_name,email_address,phone_number\nThomas Tester,test@example.org,1234567890";
```
kroky, line 252: Please add this file as a static file directly in the git repo. No need to create it dynamically here.

kambereBr:
> > Dear @kroky, thanks for your guidance. I correct it according to your recommendations
> 
> We have 2 big outstanding tasks in this ticket, so please let me know when you complete them to review again. Thanks!

Great, thank you

kambereBr:
Sorry! Closed this PR mistakenly, trying to fix the issue

kambereBr:
The PR was mistakenly closed

kambereBr:
🔄 **Reopening Notice**: Apologies for any confusion caused by the mistakenly closed of this pull request. The closure was unintended and resulted from some inadvertent manipulation. We've reopened a new pull request https://github.com/cypht-org/cypht/pull/904 to continue the progress.

PR #854, opened 2023-12-28
Ref: Replace inline msg controls with dropdown

henochit:
### Issues
Related to : [https://avan.tech/item69459](https://avan.tech/item69459-Cypht-Flag-Unflag-Read-Unread-uses-a-lot-of-space-especially-on-mobile)

### How2Test
- [X] (use mobile) go to the page message_list
- [X]  Select a mail or Switch to select mode
- [X]  There will be a drop-down list of actions

### DIFF
--- a/modules/core/message_list_functions.php
+++ b/modules/core/message_list_functions.php
@@ -352,18 +352,18 @@ if (!hm_exists('message_controls')) {
 function message_controls($output_mod) {
     $res = '<a class="toggle_link" href="#"><i class="bi bi-check-square-fill"></i></a>'.
         '<div class="msg_controls fs-6 d-none gap-1 align-items-center">
-            <div class="dropdown">
+            <div class="dropdown on_mobile">
               <a class="btn btn-outline-success btn-sm dropdown-toggle" href="#" role="button" id="actionsDropdownMenuLink" data-bs-toggle="dropdown" aria-expanded="false">
                 Actions
               </a>
             
               <ul class="dropdown-menu on_mobile" aria-labelledby="actionsDropdownMenuLink">'.
-                '<a class="dropdown-item msg_read core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="read">'.$output_mod->trans('Read').'</a>'.
-                '<a class="dropdown-item msg_unread core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="unread">'.$output_mod->trans('Unread').'</a>'.
-                '<a class="dropdown-item msg_flag core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="flag">'.$output_mod->trans('Flag').'</a>'.
-                '<a class="dropdown-item msg_unflag core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="unflag">'.$output_mod->trans('Unflag').'</a>'.
-                '<a class="dropdown-item msg_delete core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="delete">'.$output_mod->trans('Delete').'</a>'.
-                '<a class="dropdown-item msg_archive core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="archive">'.$output_mod->trans('Archive').'</a>
+                '<li><a class="dropdown-item msg_read core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="read">'.$output_mod->trans('Read').'</a></li>'.
+                '<li><a class="dropdown-item msg_unread core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="unread">'.$output_mod->trans('Unread').'</a></li>'.
+                '<li><a class="dropdown-item msg_flag core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="flag">'.$output_mod->trans('Flag').'</a></li>'.
+                '<li><a class="dropdown-item msg_unflag core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="unflag">'.$output_mod->trans('Unflag').'</a></li>'.
+                '<li><a class="dropdown-item msg_delete core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="delete">'.$output_mod->trans('Delete').'</a></li>'.
+                '<li><a class="dropdown-item msg_archive core_msg_control btn btn-sm btn-light text-black-50" href="#" data-action="archive">'.$output_mod->trans('Archive').'</a></li>
               </ul>
             </div>'.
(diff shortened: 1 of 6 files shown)


lib/output.php
```diff
@@ -135,6 +135,7 @@ public static function __callStatic(string $method, array $parameters)
...
+    public static $filter = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAB4AAAAeCAYAAAA7MK6iAAAACXBIWXMAAAsTAAALEwEAmpwYAAAArElEQVR4nO2SwQlCMRAFcxFbEOxAG/BiLV72vYVAerCM/9vxgx3YgWAL4kUR9CKIJpsg6g7kFhjyMiE4fwXJgeS50Rk+Jd68evXoeqmWEMA2pTR+a/IY4wTAvoL0oKrTrP8WkQWAo0F8EpFllvQOyZVBzGABQF8g7UzSktiyYqoVW1FMFWIrj8kYmy0m55FnU4fW0MX0qX85LgA7AGtVnTUX4yYTkXlzmfPVXACDguuOuS6tYAAAAABJRU5ErkJggg==';
```
josaphatim, line 138: @henochit please use svg instead of png ?

modules/core/message_list_functions.php
```diff
@@ -351,13 +351,29 @@ function icon_callback($vals, $style, $output_mod) {
...
-        '<a class="msg_unflag core_msg_control" href="#" data-action="unflag">'.$output_mod->trans('Unflag').'</a>'.
-        '<a class="msg_delete core_msg_control" href="#" data-action="delete">'.$output_mod->trans('Delete').'</a>'.
-        '<a class="msg_archive core_msg_control" href="#" data-action="archive">'.$output_mod->trans('Archive').'</a>';
+        '<div class="msg_controls">
+            <div class="dynamic_dropdown on_mobile">
+              <button class="dynamic_dropdown-toggle" aria-haspopup="true" aria-expanded="false">
+                Actions
+                <span class="arrow"></span>
+              </button>
+              <ul class="dynamic_dropdown-menu" aria-hidden="true" aria-labelledby="dropdownMenuButton">
```
josaphatim, line 360: Please take a look at class .cypht-dropdown and modify it's js so that we we have just one code of displaying modals.

modules/core/message_list_functions.php
```diff
@@ -351,13 +351,29 @@ function icon_callback($vals, $style, $output_mod) {
...
+                '<li><a class="msg_unread core_msg_control" href="#" data-action="unread">'.$output_mod->trans('Unreade').'</a></li>'.
+                '<li><a class="msg_flag core_msg_control" href="#" data-action="flag">'.$output_mod->trans('Flag').'</a></li>'.
+                '<li><a class="msg_unflag core_msg_control" href="#" data-action="unflag">'.$output_mod->trans('Unflag').'</a></li>'.
+                '<li><a class="msg_delete core_msg_control" href="#" data-action="delete">'.$output_mod->trans('Delete').'</a></li>'.
+                '<li><a class="msg_archive core_msg_control" href="#" data-action="archive">'.$output_mod->trans('Archive').'</a></li>
+              </ul>
+            </div>';
+
+     $res .= '<a class="msg_read core_msg_control no_mobile" href="#" data-action="read">'.$output_mod->trans('Read').'</a>'.
+            '<a class="msg_unread core_msg_control no_mobile" href="#" data-action="unread">'.$output_mod->trans('Unreade').'</a>'.
```
josaphatim, line 372: Unreade -> Unread. Same as on line 363

modules/core/message_list_functions.php
```diff
@@ -351,13 +351,29 @@ function icon_callback($vals, $style, $output_mod) {
...
-        '<a class="msg_unread core_msg_control" href="#" data-action="unread">'.$output_mod->trans('Unread').'</a>'.
-        '<a class="msg_flag core_msg_control" href="#" data-action="flag">'.$output_mod->trans('Flag').'</a>'.
-        '<a class="msg_unflag core_msg_control" href="#" data-action="unflag">'.$output_mod->trans('Unflag').'</a>'.
-        '<a class="msg_delete core_msg_control" href="#" data-action="delete">'.$output_mod->trans('Delete').'</a>'.
-        '<a class="msg_archive core_msg_control" href="#" data-action="archive">'.$output_mod->trans('Archive').'</a>';
+        '<div class="msg_controls">
+            <div class="dynamic_dropdown on_mobile">
+              <button class="dynamic_dropdown-toggle" aria-haspopup="true" aria-expanded="false">
+                Actions
+                <span class="arrow"></span>
```
josaphatim, line 358: This doesn't show the arrow on my side

lib/output.php
henochit: Okay, What is the benefit ?

modules/core/message_list_functions.php
henochit: Sure ? can you test again please
Capture d’écran 2024-01-03 à 11 49 46

modules/core/message_list_functions.php
henochit: @josaphatim, I just saw .cypht-dropdown, I realize that it is used in one place and seems to be incomplete compared to what I have set up. different dropdown are used in different places (ex: filter, etc??).

What I suggest to you is that I create a task to standardize this. I think you will know that I left a TODO in the code to signify this.

modules/core/site.js
```diff
@@ -1861,3 +1861,39 @@ function listControlsMenu() {
     $('#list_controls_menu').toggleClass('show')
     $('.list_sources').hide();
 }
+
+// Function to toggle dynamic dropdown menu
+// TODO: Replace all dropdown with this (ref: Msg controls dropdown on mobile)
+function toggleDropdown(event) {
```
josaphatim, line 1867: @henochit can you take advantage of cash.min.js to simplify JS

modules/core/message_list_functions.php
josaphatim: Still not able to see the arrow

josaphatim:
@henochit also fix failing tests

modules/core/message_list_functions.php
josaphatim: I use a different theme from the default theme, that's why I couldn't see the arrow. When I switch to the default theme I see it. Can you try other themes and fix the error?

modules/core/message_list_functions.php
kroky: Just a note that cypht-dropdown was introduced in order to not interfere with other dropdown elements used in different frameworks in integrating clients (e.g. Tiki). If we can use a single dropdown class like cypht-dropdown for all dropdowns in Cypht that will be best. You can add an additional class to this specific dropdown if you need additional functionality.

josaphatim:
@henochit Something is wrong. Please try to test with both mobile and non mobile mode.

henochit:
@josaphatim, could you please provide more details? After testing, the only issue I found is with the 'Copy & Move' dropdown. Do you notice any other issues?

PR #864, opened 2024-01-12
[FIX] allow drag and drop of recipients with accents in the name

jacob-js:
## Pull Request

This pull request addresses the issue encountered while trying to reorder recipients between the fields `to`, `cc`, and `bcc`, which was not functioning. Additionally, it resolves the problem where a bubble was not being formed for a contact with an accent in the name.

### Issues

- [X] https://avan.tech/item98845

### Checklist

- [X] None

### How2Test

- [ ] Create some contacts in your interface (with or without accents in the name)
- [ ] Navigate to `compose`
- [ ] Type the contact in the field `to`, `cc, or 'bcc`. Once the contact is found and inserted in the field
- [ ] You can drag-and-drop it to move to another field.

### Todo

- [x] None


modules/smtp/site.js
```diff
@@ -378,7 +378,7 @@ var copy_text_to_clipboard = function(e) {
 }
 
 var is_valid_recipient = function(recipient) {
-    var valid_regex = /^[\w ]*[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/;
+    var valid_regex = /^[\p{L}' ]*[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/u;
```
kroky, line 381: Why can't an email address start with a number? I am not sure what the unicode letter requirement in the beginning is used for? Maybe we need a more standard email checking regex?

modules/contacts/modules.php
```diff
@@ -176,7 +176,7 @@ protected function output() {
                     '<img alt="'.$this->trans('Send To').'" width="16" height="16" src="'.Hm_Image_Sources::$person.'" /></a> '.
                     '</d><td>'.$this->html_safe($contact->value('type')).'<td><span class="contact_src">'.
                     ($contact->value('source') == 'local' ? '' : $this->html_safe($contact->value('source'))).'</span>'.
-                    '</td><td>'.$this->html_safe($name).'</td>'.
+                    '</td><td>'.$this->html_safe_decoded($name).'</td>'.
```
kroky, line 179: If we decode html entities, won't this bring security issues listing the names here in html context? I think it defeats the purpose of using html_safe in the first place.

lib/module.php
```diff
@@ -99,6 +99,19 @@ public function html_safe($string, $special_only=false) {
...
+     * Decode sanitized input string
+     * @param string $string text to decode
+     * @param bool $special_only only use htmlspecialchars not htmlentities
+     * @return string sanitized value
+     */
+    public function html_safe_decoded($string, $special_only=false) {
+        if ($special_only) {
+            return htmlspecialchars_decode($this->html_safe($string, true), ENT_QUOTES | ENT_SUBSTITUTE);
+        }
+        return html_entity_decode($this->html_safe($string), ENT_QUOTES | ENT_SUBSTITUTE);
```
kroky, line 112: Isn't html_entity_decode the opposite of htmlentities which is what html_safe does? I don't get the idea of calling both methods on the same string one after the other.

modules/smtp/site.js
kroky: The name can't start with a number as well?

jacob-js: I'm not aware of any valid person's name starting with a number, but if it's possible for such a name to be accepted, let's consider that possibility. By the way, the use of the Unicode character allows for the inclusion of letters from various languages worldwide, such as in the name 'André'.

kroky: Even if it doesn't start with a letter itself, a name like Peter Parker 2nd <peter@parker.org> will not pass your new regex. I agree using unicode letters is a good thing but we should expand the range of matches in my opinion.

lib/module.php
kroky: Maybe you need to provide proper encoding to htmlentities on the first place? Reverting the encoding really doesn't make sense to me. We should either encode and fix the encoding issue or not encode at all and deal with the security issue.

jacob-js: image
I'm intentionally performing this action after disabling the encoding. However, unexpectedly, these characters, as depicted in the picture, are being interpreted as plain text, despite being valid HTML tags. This outcome aligns with our intentions, correct? I don't perceive any security concerns associated with this approach (not encoding).

kroky: What is the result of applying only html_safe? Which string do you have problems with? Can you paste an example here?

jacob-js: For example the string `André` when applying html_safe, we have the following returned ` Andr&eacute;`

kroky: htmlspecialchars doesn't convert the `é` character and is still ensuring good level of security. You can use it by calling html_safe with setting the second parameter to true instead of decoding an already encoded string.
Re: security considerations, see the difference of the output of the following 2 code samples:
```
$ php
<?=html_entity_decode(htmlentities('André <script>alert("hi")</script>'))?>

André <script>alert("hi")</script>
```

and

```
$ php
<?=htmlspecialchars('André <script>alert("hi")</script>')?>   

André &lt;script&gt;alert(&quot;hi&quot;)&lt;/script&gt;
```

We prefer the second one for obvious reasons.

jacob-js: The issue lies in the fact that when a character is encoded by htmlenitites(), the browser still knows how to interpret it. However, in our scenario, the problem arises because we are encoding the characters twice. The code snippet below, found within the filters, is intended to ensure the safety of the received input. However, by applying an additional layer of htmlentities(), we inadvertently compromise the integrity of our string:

```php
    'allowed_post' => array(
        'contact_email' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'contact_name' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'contact_phone' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'contact_id' => FILTER_VALIDATE_INT,
        'contact_value' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'edit_contact' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'add_contact' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'contact_source' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
        'contact_type' => FILTER_SANITIZE_FULL_SPECIAL_CHARS
    ),
```

kroky: Then, there is no need to call html_safe on an already filtered input - it is better to encode once than encode twice and decode once.

kroky: Btw, what you say is not totally correct as the sample code we are discussing calls html_safe on a contact name that comes from a contact source possibly stored externally (carddav, etc.) which we can't rely on being encoded via our POST filters...

jacob-js: Right, if we are not sure, though it's a possibility, then we should first try to decode, then encode after ourselves. But, for local contacts, we are sure that they are already encoded.

kroky: Sorry but that's incorrect. Remote sources assume data is unsafe, so we need to encode when displaying. Local storage has filters applied when inserting but still can't rely if data is encoded as it is not stored encoded - html entity encoding is meant only for displaying the data to the end-user in html context. When we store data in json context, there is no need to html-encode it. In fact, it would be harmful to encode it because other sources might read the data and JSON encoding never suggests that underlying data is also html-encoded.
I continue to dislike double encode/decode solution and propose we use simple html_safe with second paramater = true to actually use htmlspecialchars instead of htmlentities function which will solve the accent characters issue.

jacob-js: Unfortunately what you're suggesting is not solving the issue with the accent, because still, when we use htmlspecialchars the encoding is performed twice.

kroky: You mean something is calling html_safe twice? Where does the second encoding come from?

jacob-js: The first one comes from the filter sanitizer when the contact is created.

kroky: Ah, I think we are finally on the same page... hmm, but then we have the same issue throughout Cypht. We have hundreds of variables that are sanitized on the input and later html_safe is called before displaying them. This double-encoding will be an issue in search_terms, nux_name, profile_name and many other places. Sanitizing html entities on data input doesn't make sense to me. First of all, we are not sure if the target data storage will be used in html context at all. Then, html encoding is something done on the output, not input. Thus, I suggest we fix this by using:

```
filter_var($a, FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES);
```

This leaves accents, quotes, etc., so the string is untouched but still sanitize on input some invalid characters or html code. We need to make sure our output uses html_safe function to properly encode html special chars but that should mostly be the case throughout Cypht.

The best solution here should be replacing all FILTER_SANITIZE_FULL_SPECIAL_CHARS with FILTER_SANITIZE_STRING and also apply the NO_ENCODE_QUOTES flag. If you don't apply that flag, you can see what happens - try to search for a string containing a quote - the search field ends up with html-encoded string since it is double-encoded.

jacob-js: Apparently `FILTER_SANITIZE_STRING` has been deprecated since PHP 8.0.0 compelling us to find another alternative. Maybe we should just use `FILTER_SANITIZE_SPECIAL_CHARS` and add the flag `NO_ENCODE_QUOTES`?

kroky: How about FILTER_DEFAULT which is an alias of FILTER_UNSAFE_RAW?

jacob-js: That one can work well for our case, because it doesn't change anything.

kroky:
@jacob-js thank you for the changes. Can you resolve the conflicts as we can merge?

jacob-js:
> @jacob-js thank you for the changes. Can you resolve the conflicts as we can merge?

Resolved.

josaphatim:
@kroky, @jacob-js

Now  this test is failing test_filter_output
Not sure but I think that FILTER_DEFAULT doesn't filter output

josaphatim:
> FILTER_SANITIZE_FULL_SPECIAL_CHARS should be replaced with FILTER_DEFAULT in tests as well. The idea is that we don't escape html entities in input data but do it on output in html context.

test_filter_output filter uses FILTER_DEFAULT which makes the test to fail. FILTER_SANITIZE_FULL_SPECIAL_CHARS makes it to succeed but It does not make sense to use it.

kroky:
We need to change the tests as well - we are no longer filtering input for html special chars when storing the input.

PR #867, opened 2024-01-14
[ENH] allow contacts within the angle bracket format to become bubbles

jacob-js:
## Pull Request

Allow contact with the angle bracket format to become bubble in order to allow its move between fields `cc`, `bcc`, and `to`.

### How To Test

- [X] Navigate to compose page
- [X] Copy-paste the following email addresses in the field `to` and hit enter.
```
ABC <abc@example.org>, DEF <def@example.org>, XYZ <xyz@example.org>, ABC <abc@example.com>, DEF <def@example.com>, XYZ <xyz@example.com>
```
- [X] All of them should generate bubbles

### Todo

- [X] None


modules/smtp/site.js
```diff
@@ -327,13 +327,14 @@ var text_to_bubbles = function(input) {
...
 
-    if ($(input).val() && contact_id) {
+    if ($(input).val()) {
         var recipients = $(input).val().split(/,|;/);
         var invalid_recipients = '';
 
         for (var i = 0; i < recipients.length; i++) {
             if (is_valid_recipient(recipients[i])) {
-                append_bubble(recipients[i].trim(), input, contact_id, contact_type, contact_source);
+                const value = recipients[i].trim().replace(/[<>]/g, '')
```
kroky, line 336: Why do we remove the brackets before adding the bubble? `Name <email>` is a valid format of recipients according to the RFC.

jacob-js: Yes, it is, but when we display the bubble we don't show those brackets. The format that we display is `Name email`.

kroky: I don't think we should strip the brackets in that case. It is standard display in email clients to display the brackets, so unless there is a compelling reason not to display them, can you return them back?

jacob-js: Alright, let's go for it.

Issue #866, opened 2024-01-13
Sending an email with no body causes fatal PHP error


PaulTGG:
## 🐛 Bugreport

The fatal error leaves users with a blank screen.  Example error:

[Fri Jan 12 16:41:04.080582 2024] [proxy_fcgi:error] [pid 26227] [client *deleted*] AH01071: Got error 'PHP message: PHP Fatal error:  Uncaught ValueError: DOMDocument::loadHTML(): Argument #1 ($source) must not be empty in /usr/local/share/cypht/modules/core/message_functions.php:438\nStack trace:\n#0 /usr/local/share/cypht/modules/core/message_functions.php(438): DOMDocument->loadHTML()\n#1 /usr/local/share/cypht/modules/core/message_functions.php(41): HTMLToText->__construct()\n#2 /usr/local/share/cypht/modules/smtp/hm-mime-message.php(244): convert_html_to_text()\n#3 /usr/local/share/cypht/modules/smtp/hm-mime-message.php(92): Hm_MIME_Msg->prep_message_body()\n#4 /usr/local/share/cypht/modules/smtp/modules.php(730): Hm_MIME_Msg->get_mime_msg()\n#5 /usr/local/share/cypht/lib/modules_exec.php(177): Hm_Handler_process_compose_form_submit->process()\n#6 /usr/local/share/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()\n#7 /usr/local/share/cypht/lib/dispatch.php(221): Hm_Module_Exec->run_handler_modules()\n#8 /usr/local/share/cypht/lib/dispatch.php(191): Hm_Dispatch->process_request()\n#9 /...'

@marclaporte suggested the user be prompted if they try to send an email with a blank body or subject.  A good idea, and it would also be nice if it was an option that could be toggled.

### Version & Environment

Rev: Not sure, but it was the master on January 5.

OS:  Debian 12.4.0

### Steps to reproduce
1. Try sending an email with an empty body.
2. Done.

PR #828, opened 2023-11-15
Move to Bootstrap 5

alvinBM:
## Pullrequest
Move to bootstrap 5

## Issues
- #762

PR #868, opened 2024-01-15
[NEW] add a modal dialog to warn the user about sending an email without a body

jacob-js:
## Description

Add a warning prompt that's shown when the user is about to send an email with an empty body.

### Issues

- [X] #866

### DIFF
--- a/modules/core/site.css
+++ b/modules/core/site.css
@@ -330,3 +330,104 @@ div.unseen, .unseen .subject { font-weight: 700; }
 .drag_target {
     background-color: #888 !important;
 }
+
+.cypht-modal {
+    display: block;
+    position: fixed;
+    z-index: 1;
+    left: 0;
+    top: 0;
+    width: 100%;
+    height: 100%;
+}
+
+.cypht-modal-bg {
+    background-color: rgba(0,0,0,0.4);
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 100%;
+    height: 100%;
+}
+  
+.cypht-modal-content {
+    background-color: #fefefe;
+    margin: 15% auto;
+    padding: 20px;
+    border: 1px solid #888;
+    width: 50%;
+    display: flex;
+    flex-direction: column;
+    gap: 20px;
+    position: relative;
+}
+
+.cypht-modal-content-close {
+    color: #aaa;
+    float: right;
+    font-size: 28px;
+    font-weight: bold;
+    align-self: flex-end;
+    position: absolute;
+}
+
+.cypht-modal-content-close:hover,
+.cypht-modal-content-close:focus {
+    color: black;
+    text-decoration: none;
+    cursor: pointer;
+}
+
+.cypht-modal-header {
+    font-size: 1.5rem;
+    font-weight: 600;
+}
+
+.cypht-modal-footer {
+    display: flex;
+    justify-content: flex-end;
+    gap: 10px;
+}
+
+.cypht-modal-footer button {
+    padding: 10px 20px;
+    cursor: pointer;
+}
+
+.cypht-modal-content-close:hover,
+.cypht-modal-content-close:focus {
+    color: black;
+    text-decoration: none;
+    cursor: pointer;
+}
+
+.cypht-modal-header {
+    font-size: 1.5rem;
+    font-weight: 600;
+}
+
+.cypht-modal-footer {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: flex-end;
+    gap: 10px;
+}
+
+.cypht-modal-footer button {
+    padding: 10px 20px;
+    cursor: pointer;
+    height: max-content;
+    width: max-content;
+}
+
+.mobile .cypht-modal-content {
+    width: 75%;
+}
+
+.mobile .cypht-modal-footer {
+    justify-content: flex-start;
+}
+
+.mobile .cypht-modal-footer button {
+    padding: 5px 10px;
+}
--- a/modules/smtp/site.js
+++ b/modules/smtp/site.js
@@ -406,20 +406,86 @@ $(function() {
         $('.delete_draft').on("click", function() { smtp_delete_draft($(this).data('id')); });
         $('.smtp_save').on("click", function() { save_compose_state(false, true); });
         $('.smtp_send_archive').on("click", function() { send_archive(false, true); });
-        $('.compose_form').on('submit', function() {
-            var msg_uid = hm_msg_uid();
-            var detail = Hm_Utils.parse_folder_path(hm_list_path(), 'imap');
-            var class_name = 'imap_'+detail.server_id+'_'+msg_uid+'_'+detail.folder;
-            var key = 'imap_'+Hm_Utils.get_url_page_number()+'_'+hm_list_path();
-            var next_message = Hm_Message_List.prev_next_links(key, class_name)[1];
-            if (next_message) {
-                $('.compose_next_email_data').val(next_message);
+        $('.compose_form').on('submit', function(e) {
+            e.preventDefault();
+            const body = $('.compose_body').val().trim();
+            const subject = $('.compose_subject').val().trim();
+
+            let modalContentHeadline = '';
+            let dontWanValueInStorage = '';
+
+            // If the subject is empty, we should warn the user
+            if (!subject) {
+                dontWanValueInStorage = 'dont_warn_empty_subject';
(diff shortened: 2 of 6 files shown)


PaulTGG:
I found that when the setting is disabled, I was shown a warning, and prevented from sending an email with an empty body.  However, when the setting is enabled, trying to send an email with an empty body still crashes php.  Here's the last error message.

`[Tue Jan 16 10:59:32.513723 2024] [proxy_fcgi:error] [pid 57369] [client 172.16.32.226:60912] AH01071: Got error 't()\n#3 /usr/local/share/testcypht/modules/smtp/hm-mime-message.php(92): Hm_MIME_Msg->prep_message_body()\n#4 /usr/local/share/testcypht/modules/smtp/modules.php(736): Hm_MIME_Msg->get_mime_msg()\n#5 /usr/local/share/testcypht/lib/modules_exec.php(177): Hm_Handler_process_compose_form_submit->process()\n#6 /usr/local/share/testcypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()\n#7 /usr/local/share/testcypht/lib/dispatch.php(221): Hm_Module_Exec->run_handler_modules()\n#8 /usr/local/share/testcypht/lib/dispatch.php(1...'`

marclaporte:
I don't want one more setting. Just make it that there is warning to the user.

PaulTGG:
IMHO, giving users the option is never a bad thing (especially when you can just "set it and forget it"), although I might just be saying that because of my use-case.  Either way, the php crash is still happening in the background.

Jan 16, 2024 6:07:36 p.m. Marc Laporte ***@***.***>:

marclaporte:
Options are more code to maintain. It's one more thing to think about when testing or reporting issues.

I can't think of a real-world use case where someone would regularly and intentionally send emails without a subject or body.

PaulTGG:
> I can't think of a real-world use case where someone would regularly and intentionally send emails without a subject or body.

You clearly don't work with some of my co-workers... lol  Lots of subject-only emails flying around!

Could a workaround be to pre-populate the email composer with the signature line?  Then as long as the profile has a signature, there would always be something in the email body.

marclaporte:
Idea: The warning could be:

* Cancel sending (default)
* Send anyway
* Send anyway and don't warn me in the future

The third option would add a cookie to the user's browser.

jacob-js:
> I found that when the setting is disabled, I was shown a warning, and prevented from sending an email with an empty body. However, when the setting is enabled, trying to send an email with an empty body still crashes php. Here's the last error message.
> 
> `[Tue Jan 16 10:59:32.513723 2024] [proxy_fcgi:error] [pid 57369] [client 172.16.32.226:60912] AH01071: Got error 't()\n#3 /usr/local/share/testcypht/modules/smtp/hm-mime-message.php(92): Hm_MIME_Msg->prep_message_body()\n#4 /usr/local/share/testcypht/modules/smtp/modules.php(736): Hm_MIME_Msg->get_mime_msg()\n#5 /usr/local/share/testcypht/lib/modules_exec.php(177): Hm_Handler_process_compose_form_submit->process()\n#6 /usr/local/share/testcypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()\n#7 /usr/local/share/testcypht/lib/dispatch.php(221): Hm_Module_Exec->run_handler_modules()\n#8 /usr/local/share/testcypht/lib/dispatch.php(1...'`

@PaulTGG which PHP version is your environment using? I did not get any such error when trying this on my side. Could it be related to your PHP version?

jacob-js:
> Idea: The warning could be:
> 
> * Cancel sending (default)
> * Send anyway
> * Send anyway and don't warn me in the future
> 
> The third option would add a cookie to the user's browser.

Good.

PaulTGG:
> @PaulTGG which PHP version is your environment using? I did not get any such error when trying this on my side. Could it be related to your PHP version?

@jacob-js I'm using PHP 8.2.7.  It's the latest version that's available in the Debian repositories.  (Using Debian 12.4.)

jacob-js:
> > @PaulTGG which PHP version is your environment using? I did not get any such error when trying this on my side. Could it be related to your PHP version?
> 
> @jacob-js I'm using PHP 8.2.7. It's the latest version that's available in the Debian repositories. (Using Debian 12.4.)

Can you please try switching to version 7.4?

PaulTGG:
> Can you please try switching to version 7.4?

I wasn't able to complete the install using PHP 7.4.  It's not in the repository of Debian 12, and I tried Debian 11, but composer gave the following error:  

`require.composer is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separated by -, . or _. The complete name should match "^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$".`

Hopefully the PHP version isn't the issue, since 7.4 was end-of-life a little over a year ago!

modules/smtp/site.js
```diff
@@ -419,7 +421,33 @@ $(function() {
...
-            $('.smtp_send').on("click", function() { return false; }); 
+            $('.smtp_send').on("click", function() { return false; });
+
+            if(Hm_Utils.get_from_local_storage('dont_warn_empty_body')) {
+                e.target.submit();
+                return;
+            }
+
+            const modalContent = `
+            <p>You're about to send an empty body</p>
```
marclaporte, line 432: There should be a message somewhere about an empty subject.

modules/core/site.css
```diff
@@ -329,4 +329,105 @@ div.unseen, .unseen .subject { font-weight: 700; }
 }
 .drag_target {
     background-color: #888 !important;
+}
+
+.modal {
```
marclaporte, line 334: All this will need to move to Bootstrap 5 soon: https://github.com/cypht-org/cypht/pull/828

modules/smtp/site.js
jacob-js: Empty subject or empty body you mean? I'm not following.

jacob-js:
> > Can you please try switching to version 7.4?
> 
> I wasn't able to complete the install using PHP 7.4. It's not in the repository of Debian 12, and I tried Debian 11, but composer gave the following error:
> 
> `require.composer is invalid, it should have a vendor name, a forward slash, and a package name. The vendor and package name can be words separated by -, . or _. The complete name should match "^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$".`
> 
> Hopefully the PHP version isn't the issue, since 7.4 was end-of-life a little over a year ago!

@PaulTGG  By now, I'm certain you will not get that exception anymore using the PHP version you have.

modules/smtp/site.js
marclaporte: Both. Cypht should warn if subject or body is empty.

jacob-js: Good. Implemented✅

PaulTGG:
> @PaulTGG By now, I'm certain you will not get that exception anymore using the PHP version you have.

Just tried it, and everything looks good, thanks!

modules/core/site.css
```diff
@@ -329,4 +329,105 @@ div.unseen, .unseen .subject { font-weight: 700; }
...
+.modal-bg {
+    background-color: rgba(0,0,0,0.4);
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 100%;
+    height: 100%;
+}
+  
+.modal-content {
```
josaphatim, line 353: @jacob-js this is a good start for refactoring and making uniform cypht modals but as you can view in this comment https://github.com/cypht-org/cypht/pull/854#discussion_r1441515544, using class names used by popular frameworks can lead to unexpected behaviors when Cypht is embedded. Can you prefix all these classes with something like **cypht-**

modules/ldap_contacts/modules.php
```diff
@@ -776,7 +776,7 @@ function fetch_ldap_contacts($config, $user_config, $contact_store, $session=fal
     $ldap_config = ldap_add_user_auth($ldap_config, $user_config->get('ldap_contacts_auth_setting', array()));
     if (count($ldap_config) > 0) {
         foreach ($ldap_config as $name => $vals) {
-            $vals['name'] = $name;
+            is_array($vals) && $vals['name'] = $name;
```
josaphatim, line 779: This is being fixed in this commit https://github.com/cypht-org/cypht/pull/869. You can remove this change.

PR #869, opened 2024-01-16
Fix ldap_contacts error running under php8.2 & remove unused and correct ldap .env variable

Shadow243:
### Problem Description:
Under the PHP 8.2 environment in Cypht, an error is encountered in the ldap_contacts module, specifically on line 320. The error is related to `array_key_exists()` where a string is given instead of an array. Notably, this error is not displayed under PHP 7.4. The issue occurs on pages such as settings and compose, leading to an error page.

### Steps to Reproduce:
1. Set up Cypht with PHP 8.2.
2. Navigate to the settings or compose page.
3. Observe the error related to `array_key_exists()` on line 320 in the ldap_contacts module.

#### Code Modification:
- Investigate and resolve the issue causing the `array_key_exists()` error in the ldap_contacts module under PHP 8.2.
- Ensure compatibility with PHP 8.2, and 8.3 without compromising functionality in PHP 7.4.

#### Cleanup:
- Remove unused environment variables from the `.env.example` file.
- Rename the LDAP_AUTH variable in `.env.example` according to `config/ldap.php` file.

### Additional Information:
- The error seems specific to the interaction between Cypht and PHP 8.2.
- No such error is observed when using PHP 7.4.
- Clean up the environment variable configuration for better maintenance.

### Environment:
- PHP 8.2
- Cypht
- Ubuntu server on Digital Ocean

### Steps to Test:
1. Apply the code modifications.
2. Confirm the error is resolved under PHP 8.2.
3. Verify that functionality remains intact under PHP 7.4.
4. Test the cleanup changes related to environment variables.

### Related Issues:
N/A

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,10 +11,6 @@ DB_SOCKET=/var/lib/mysqld/mysqld.sock
 
 SESSION_TYPE=PHP
 AUTH_TYPE=DB
-LDAP_AUTH_SERVER=localhost
-LDAP_AUTH_PORT=389
-LDAP_AUTH_TLS=
-LDAP_AUTH_BASE_DN=example,dc=com
 
 IMAP_AUTH_NAME=localhost
 IMAP_AUTH_SERVER=localhost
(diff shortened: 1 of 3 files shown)


josaphatim:
@Shadow243 

I think there are two configurations that are different. Try to read these below code below.

This config was used when auth_type was set to LDAP
https://github.com/cypht-org/cypht/blob/1.4.x/hm3.sample.ini#L58

This config file has sections, for each LDAP backend you want to support
https://github.com/cypht-org/cypht/blob/1.4.x/modules/ldap_contacts/ldap.ini#L11

Shadow243:
@josaphatim 

You are right, I updated.

> @Shadow243
> 
> I think there are two configurations that are different. Try to read code below.
> 
> This config was used when auth_type was set to LDAP https://github.com/cypht-org/cypht/blob/1.4.x/hm3.sample.ini#L58
> 
> This config file has sections, for each LDAP backend you want to support https://github.com/cypht-org/cypht/blob/1.4.x/modules/ldap_contacts/ldap.ini#L11

modules/ldap_contacts/modules.php
```diff
@@ -316,10 +316,11 @@ public function process() {
 class Hm_Handler_load_ldap_settings extends Hm_Handler_Module {
     public function process() {
         $connections = array();
-        foreach (ldap_config($this->config) as $name => $vals) {
-            if (array_key_exists('auth', $vals) && $vals['auth']) {
-                if ((!array_key_exists('user', $vals) || !$vals['user']) &&
-                    (!array_key_exists('pass', $vals) || !$vals['pass'])) {
+        $ldap_config = ldap_config($this->config);
+        foreach ($ldap_config as $name => $vals) {
+            if (array_key_exists('auth', $ldap_config) && $ldap_config['auth']) {
```
kroky, line 321: I don't think this is the right way to fix the issue. Note that previously, Cypht supported multiple sets of ldap_config arrays named differently (e.g. Personal, etc.). We need to keep the same behavior with .env migration. E.g. by default, we put the initial ldap config in Personal namespace (array key) and then allow user to add more configs if needed. Then, the loops here in ldap_contacts module shouldn't be changed.

Shadow243: Copy that @kroky, By updating ldap.php like this:
```
 'ldap' => [
        'Personal' => [....]
  ]
```
It works, I changed also Carddav.php to use the same structure.

kroky: Great, thanks!

PR #872, opened 2024-01-23
Issue with module name during development on Windows

ElvisAns:
### Issues

- [X] During development (APP_DEBUG enabled) cypht dont load correctly css and js files unless i go to /site  
image
image

### Checklist

- [X] None

### How2Test

- [X] I tested this on a windows computer, so i believe it's a platform dependent issue. To test, pull the branch an ensure the site load correctly without going to /site
image

### Todo

- [X] None

### DIFF
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -504,9 +504,9 @@ class Hm_Output_header_css extends Hm_Output_Module {
         if (DEBUG_MODE) {
             $res .= '<link href="vendor/twbs/bootstrap/dist/css/bootstrap.min.css" rel="stylesheet" type="text/css" />';
             $res .= '<link href="vendor/twbs/bootstrap-icons/font/bootstrap-icons.css" rel="stylesheet" type="text/css" />';
-            foreach (glob(APP_PATH.'modules/**', GLOB_ONLYDIR | GLOB_MARK) as $name) {
+            foreach (glob(APP_PATH.'modules'.DIRECTORY_SEPARATOR.'**', GLOB_ONLYDIR | GLOB_MARK) as $name) {
                 $rel_name = str_replace(APP_PATH, '', $name);
-                $mod = str_replace(array('modules/', '/', '\\'), '', $rel_name);
+                $mod = str_replace(array('modules', DIRECTORY_SEPARATOR), '', $rel_name);
                 if (in_array($mod, $mods, true) && is_readable(sprintf("%ssite.css", $name))) {
                     $res .= '<link href="'.sprintf("%ssite.css", $rel_name).'" media="all" rel="stylesheet" type="text/css" />';
                 }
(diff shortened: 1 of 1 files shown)


modules/core/output_modules.php
```diff
@@ -498,7 +498,7 @@ protected function output() {
         if (DEBUG_MODE) {
             foreach (glob(APP_PATH.'modules/**', GLOB_ONLYDIR | GLOB_MARK) as $name) {
                 $rel_name = str_replace(APP_PATH, '', $name);
-                $mod = str_replace(array('modules/', '/'), '', $rel_name);
+                $mod = str_replace(array('modules/', '/', '\\'), '', $rel_name);
```
kroky, line 501: Why don't you try with DIRECTORY_SEPARATOR constant when you see windows-specific behavior problems?

ElvisAns: @victor i am not sure if the DIRECTORY SEPARATOR is needed here
the idea is to replace "modules/" "/"  and "\" by an empty string "" so that we stay with only the name of the module as return value
On Windows the original string is comming with "\" but we didn't have "\" in the replacement

victor: please be careful with your @-mentions

kroky: I see you added forward slash to the array of replacements. That's the DIRECTORY_SEPARATOR constant in windows. Please use it instead of guessing or trying to replace all possible variants.

josaphatim: @kroky on windows the $rel_name has values like **modules/2fa\**. This is one of values I copied from a var_dump that I did on windows. That why I think the / and \ are necessary there.

josaphatim: > @kroky i am not sure if the DIRECTORY SEPARATOR is needed here the idea is to replace "modules/" "/" and "\" by an empty string "" so that we stay with only the name of the module as return value On Windows the original string is coming with "\" but we didn't have "\" in the replacement

@ElvisAns Windows returns values like `modules/2fa\` in $rel_name variable. With the modules/ replaced we just need to remove the DIRECTORY_SEPARATOR as for this example the remaining part will be `2fa\`

ElvisAns: Thanks @kroky & @josaphatim

kroky: Hmm, I thought it would be something like 'modules' . DIRECTORY_SEPARATOR, so you don't end up testing for forward and backward slashes.

ElvisAns: Aha!
@kroky We are back to what i said initially
The original string contain a mixture of "/" and "\" when we use DIRECTORY_SEPARATOR we will only replace "\" on windows and "/" on linux but not both at the same resulting in an issue again
image

kroky: If you end up with a mixture of these slashes, you are probably not using the directory separator correctly. I still see hard-coded forward slashes in modules/ string and in glob pattern.

ElvisAns: You are absolutely right @kroky 🙌

Issue #861, opened 2024-01-10
Option to always allow external images


PaulTGG:
## 🗣 Suggestion
This may fly in the face of security, but it would be nice if there was an option to "always show inline images."  At number of email senders use inline images (whether for simple header images, formatting purposes, aesthetic reasons, or, more importantly, for confirmation buttons, or action buttons that users have to click), so it would be nice to have an option where the user can accept the risks involved, and just let the images appear for all their mail.

marclaporte:
Brainstorming a few options:
* Never permit external images (the default)
* Permit external images for this email (an option now)
* Permit external images from known senders (a reasonable enhancement)
* Permit external images per image source (This could be a lot of work)
* Permit all external images (Opens invasion of privacy)

marclaporte:
For confirmation buttons, or action buttons that users have to click: I noticed that Roundcube deals with this better: They show where the image is, so you know where to click even if the image is not loaded. Cypht should do this as well.

PaulTGG:
I think your third option is best - something like "Always allow images from this sender."

> Brainstorming a few options:
> 
>     * Never permit external images (the default)
> 
>     * Permit external images for this email (an option now)
> 
>     * Permit external images from known senders (a reasonable enhancement)
> 
>     * Permit external images per image source (This could be a lot of work)
> 
>     * Permit all external images (Opens invasion of privacy)

marclaporte:
> For confirmation buttons, or action buttons that users have to click: I noticed that Roundcube deals with this better: They show where the image is, so you know where to click even if the image is not loaded. Cypht should do this as well.

Spin off: https://github.com/cypht-org/cypht/issues/865

marclaporte:
Please make it nicer visually and clearer for end user what is the security/privacy concern.

![2024-01-17_095324](https://github.com/cypht-org/cypht/assets/1004261/c6f49449-f790-4189-aa6a-726f656a79ee)

marclaporte:
Please make a list of each external ressource, and permit that we approve them individually, with an option to approve all if there is more than one.

PR #873, opened 2024-01-23
Option to always allow external images

jacob-js:
## Description

show the list of all external resources in an email and allow the user to approve each one individually or all at once or approve all resources from the sender.

### Issues

- [X] https://github.com/cypht-org/cypht/issues/861

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -16,16 +16,26 @@
 if (!hm_exists('format_msg_html')) {
 function format_msg_html($str, $images=false) {
     $str = str_ireplace('</body>', '', $str);
+
     $config = HTMLPurifier_Config::createDefault();
+    $config->set('HTML.DefinitionID', 'hm-message');
+    $config->set('HTML.DefinitionRev', 1);
     $config->set('Cache.DefinitionImpl', null);
     $config->set('HTML.TargetBlank', true);
     $config->set('HTML.TargetNoopener', true);
-    $config->set('HTML.Allowed', 'a[href|target]');
+
     if (!$images) {
         $config->set('URI.DisableExternalResources', true);
     }
     $config->set('URI.AllowedSchemes', array('mailto' => true, 'data' => true, 'http' => true, 'https' => true));
     $config->set('Filter.ExtractStyleBlocks.TidyImpl', true);
+
+    $def = $config->getHTMLDefinition(true);
+    $html_tags = ['img', 'script', 'iframe', 'audio', 'embed', 'source', 'track', 'video'];
+    foreach ($html_tags as $tag) {
+        $def->addAttribute($tag, 'data-src', 'Text');
+    }
+    
     try {
         $purifier = new HTMLPurifier($config);
         return $purifier->purify($str);
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -88,6 +88,7 @@ class Hm_Output_filter_message_body extends Hm_Output_Module {
     /**
      * Format html, text, or image content
      */
+
     protected function output() {
         $txt = '<div class="msg_text_inner">';
         if ($this->get('msg_text')) {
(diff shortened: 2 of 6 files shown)


kroky:
Why do we want to keep the exceptions in a set of cookies? Might be too much information for the cookie storage - imagine people allowing all external resources in many emails one after the other. Instead, can't we do it all in JS on client-side?

E.g. server-side continues to replace src attributes but also adds data-source, so JS code can access the original sources. It can also add the proper allow/allow all links but then JS handlers will actually execute, replace img tags with actual sources and allow the display if user clicks the action buttons/links.

jacob-js:
> Why do we want to keep the exceptions in a set of cookies? Might be too much information for the cookie storage - imagine people allowing all external resources in many emails one after the other. Instead, can't we do it all in JS on client-side?
> 
> E.g. server-side continues to replace src attributes but also adds data-source, so JS code can access the original sources. It can also add the proper allow/allow all links but then JS handlers will actually execute, replace img tags with actual sources and allow the display if user clicks the action buttons/links.

What you're suggesting is valid, but then, it would require the user to always perform that action of allowing external images whenever they open the same email and want to visualize images because we are not storing the state anywhere. I think the purpose is to make it easy for the user so they don't repeat the same action always, we are even planning to add another option that will allow the user to always display images from a certain sender.
What do you think @kroky?

kroky:
Can we use local storage or session storage much like other cypht code use it in that case?

jacob-js:
> Can we use local storage or session storage much like other cypht code use it in that case?

Ok, let's explore that way.

modules/imap/output_modules.php
```diff
@@ -101,17 +102,32 @@ protected function output() {
             }
             if (isset($struct['subtype']) && strtolower($struct['subtype']) == 'html') {
                 $allowed = $this->get('header_allow_images');
```
kroky, line 104: If we stop using header_allow_images and imap_allow_images, shouldn't we get rid of them in the handler methods as well? Also what is going to happen with the allow_external_image_sources config setting? If we don't respect it, then we should probably remove it from settings.

modules/imap/output_modules.php
```diff
@@ -101,17 +102,32 @@ protected function output() {
...
+                }, $externalResources[0]);
+
+                $externalResourcesText = '<div class="external_notices">';
+                
+                $externalResourcesText .= '</div>';
+                $txt .= $externalResourcesText;
+
+                $replacer = "**000**";
+                $msgText = preg_replace($externalResRegexp, $replacer, $msgText);
+                $msgTextArray = explode($replacer, $msgText);
```
kroky, line 121: Note that you have preg_replace_callback function that allows you to do a replace and execute custom php code while replacing - I think it will make this block easier to follow and also skip the need of custom placeholders like **000**.

jacob-js: I really don't understand how preg_replace_callback would help me to get an array whose elements are the values returned from the same function `preg_replace_callback`.

kroky: Something like:
```
$finalText = preg_replace_callback('/src="(https?:\/\/[^"]*)"/i', function ($matches) use ($allowed) {
    return 'data-src="' . ($allowed ? $matches[1] : '') .'" ' . 'src="" ' . 'data-message-part="'.$this->html_safe($this->get('imap_msg_part')) . '"';
}, $msgText);
```
Also note that src attributes might enclose the links in single quotes.

modules/core/message_functions.php
```diff
@@ -16,16 +16,26 @@
...
 function format_msg_html($str, $images=false) {
     $str = str_ireplace('</body>', '', $str);
+
     $config = HTMLPurifier_Config::createDefault();
+    $config->set('HTML.DefinitionID', 'hm-message');
+    $config->set('HTML.DefinitionRev', 1);
     $config->set('Cache.DefinitionImpl', null);
     $config->set('HTML.TargetBlank', true);
     $config->set('HTML.TargetNoopener', true);
-    $config->set('HTML.Allowed', 'a[href|target]');
```
kroky, line 23: Don't we want to keep links in html content? We even had recently added external links to open in new tabs. Wouldn't this change remove the links from the purified content?

modules/imap/output_modules.php
kroky: OK, I made some more investigations and think we should do:
- respect allow_external_image_sources application config setting - if that's off - Content-securty-policy disables external resource loading altogether, so it doesn't makes sense to do any of this data-src replacing and handling - it is probably best to just keep the old URI.DisableExternalResources setting or replace the src attribute and skip adding data-src. I think keep the old setting is more robust as we might have missed a use-case of the src matching.
- if that's on, we go ahead and do the replacing of data-src and other logic.
All other variables like imap_allow_images should be removed from handlers as they are not used.

modules/core/message_functions.php
jacob-js: I believe not. Links are present by default. This configuration allows only the `a` tag and does not return the entire message content when I inspected it.

modules/imap/output_modules.php
jacob-js: Alright, this is much better. I'm adjusting the regex so that it takes into consideration the links enclosed in single quotes as well.

modules/imap/output_modules.php
jacob-js: OK, then we'll have also to not display the warning messages to the user about external resources in case the setting allow_external_image_sources was off.

modules/core/message_functions.php
kroky: Alright, I am not very familiar with the purifier.

modules/imap/output_modules.php
```diff
@@ -97,17 +98,27 @@ protected function output() {
...
-                            $this->trans('Allow Images').'</a></div>';
-                    }
+                $msgText = $this->get('msg_text');
+                // Everything in the message starting with src="http:// or src="https:// or src='http:// or src='https://
+                $externalResRegexp = '/src="(https?:\/\/[^"]*)"|src=\'(https?:\/\/[^\']*)\'/i';
+                // Get all external resources
+                preg_match_all($externalResRegexp, $msgText, $externalResources);
+                $externalResources = array_map(function ($resource) {
+                    return preg_replace('/src="(.*)"/i', '$1', $resource);
+                }, $externalResources[0]);
```
kroky, line 108: Lines 105-108 should be deleted as we well, right?

PR #891, opened 2024-02-11
Added bootstrap to carddav and ldap contacts modules

josaphatim:
This MR adds bootstrap to carddav contacts and ldap contacts modules

### DIFF
--- a/modules/carddav_contacts/modules.php
+++ b/modules/carddav_contacts/modules.php
@@ -269,11 +269,11 @@ class Hm_Output_carddav_auth_settings extends Hm_Output_Module {
             }
             $res .= '<tr class="carddav_settings"><td>'.$this->html_safe($name).'</td><td>';
             $res .= '<input autocomplete="username" type="text" value="'.$user.'" name="carddav_usernames['.$this->html_safe($name).']" ';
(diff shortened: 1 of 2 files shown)


alvinBM:
hello @josaphatim 
Thank you for noticing these modules which were forgotten during the bootstrap migration. 
For the ldap_contacts module, I suggest you use the [Floating label](https://getbootstrap.com/docs/5.0/forms/floating-labels/)  on the form fields as this form is very long, to always have the name of the field visible when it is edited or when it is filled. So you could also get rid of the `.screen_reader` class on the `label` tags

josaphatim:
> hello @josaphatim Thank you for noticing these modules which were forgotten during the bootstrap migration. For the ldap_contacts module, I suggest you use the [Floating label](https://getbootstrap.com/docs/5.0/forms/floating-labels/) on the form fields as this form is very long, to always have the name of the field visible when it is edited or when it is filled. So you could also get rid of the `.screen_reader` class on the `label` tags

Thanks @alvinBM copy that !

josaphatim:
@alvinBM now ready.

alvinBM:
> @alvinBM now ready.

Thanks @josaphatim

PR #905, opened 2024-02-16
Reorganize system alert messages based on bootstrap alerts

alvinBM:


modules/sievefilters/site.js
```diff
@@ -501,13 +492,11 @@ $(function () {
 
         function save_script(imap_account) {
             if ($('.modal_sieve_script_name').val() === "") {
-                $('.sys_messages').html('<span class="err">You must provide a name for your script</span>');
-                Hm_Utils.show_sys_messages();
+                $('.sys_messages').html('<div class="alert alert-danger alert-dismissible fade show" role="alert"><i class="bi bi-exclamation-triangle me-2"></i><span class="err">You must provide a name for your script</span><button type="button" class="btn-close" data-bs-dismiss="alert" aria-label="Close"></button></div>');
```
josaphatim, line 495: Can you please refactor so there will be one class to manage alert and to build look according to the type. In that way we won't repeat code.

alvinBM: Yes @josaphatim  , you're right, I'm working on refactoring all that

alvinBM:
Hello @josaphatim , This PL is ready

josaphatim:
In my opinion, the Hm_Notices::show function to which we pass an array of messages as the first parameter should display an alert for each message instead of having a single alert for all the messages passed separated by a comma.
But also in the same function please add the type of error depending on ERR passed before the string or not.

But also the Hm_Utils::show_sys_messages method had the sole responsibility of displaying pending messages already added to the element $('.sys_messages') via elements having the .err class.

I think it could keep the same role but just think about how to keep messages waiting before displaying

modules/core/site.js
```diff
@@ -388,46 +388,44 @@ var Hm_Notices = {
...
+        var type = 'info';
         for (var i in msgs) {
             if (msgs[i].match(/^ERR/)) {
-                msg_list.push(msgs[i].substring(3));
+                message = msgs[i].substring(3);
+                type = 'danger';
             }
             else {
-                msg_list.push(msgs[i]);
+                message = msgs[i];
```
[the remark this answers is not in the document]
alvinBM: Yeah, you are right. It's fixed

PR #910, opened 2024-02-20
Fixed saving smtp servers

josaphatim:
Fixed issue in hm-smtp after migrating to uniqid saving a server wasn't working.


kroky:
Nice catch, thanks!

Issue #871, opened 2024-01-17
Add a simpler way to show source of an email (Like in Roundcube)


marclaporte:
![2024-01-17_101902](https://github.com/cypht-org/cypht/assets/1004261/adedd648-4f50-49ac-8086-a7d37385f8ea)

marclaporte:
@jacob-js I like this!

I see the source, and I have a link to download or copy to clipboard. Can we have a way to return to the email?

jacob-js:
> @jacob-js I like this!
> 
> I see the source, and I have a link to download or copy to clipboard. Can we have a way to return to the email?

@marclaporte in case you didn't notice, the source is opened in a new tab, leaving the email tab safe. I just think that in case the user wants to go back, they can close the newly opened tab via the close button provided by the browser. Reimplementing this on the page would be an unnecessary overload. Does this resonate well with you?

marclaporte:
Makes sense. Closing.

PR #918, opened 2024-02-23
[NEW] add a way to show and download the source of an email

jacob-js:
# How to test

1. Click the link `show source`

2. Get the email source in a new tab

# Issue: #871

### DIFF
--- a/modules/imap/site.js
+++ b/modules/imap/site.js
@@ -736,6 +736,11 @@ var imap_message_view_finished = function(msg_uid, detail, skip_links) {
 
         return block_unblock_sender(msg_uid, detail, scope, action, sender, reject_message);
     });
+    $('#show_message_source').on("click", function(e) {
+        e.preventDefault();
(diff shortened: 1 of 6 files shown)


modules/imap/site.js
```diff
@@ -1375,3 +1380,23 @@ if (message) {
...
 }
+
+const handleDownloadMsgSource = function() {
+    const messageSource = document.querySelector('pre.msg_source');
+    const blob = new Blob([messageSource.textContent], { type: "message/rfc822" });
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement('a');
+    a.href = url;
+    const subject = messageSource.textContent.match(/Subject: (.*)/)?.[1];
+    a.download = subject + '.eml';
```
kroky, line 1391: Can we fallback to msg uid or something similar if this regex match fails, so the filename is still something valid?

jacob-js: That's a valid point because the subject could be empty. So, let's use the msg uid as a fallback.

PR #941, opened 2024-04-04
Fixed redirection to server page when cypht is used as plugin in tiki

josaphatim:


modules/nux/site.js
```diff
@@ -52,13 +52,11 @@ var display_final_nux_step = function(res) {
...
                 {'name': 'imap_service_name', value: res.nux_service_name}],
-                function(res) {
-                    window.location.href = "?page=servers";
-                }
+                false
             );
         }
 
-        window.location.href = "?page=servers";
+        window.location.reload(true);
```
kroky, line 59: Reload might not load the same page as a link to servers page. I think we should do a better job with redirects from javascript. Please check Tiki integration code - there is a Tiki_Hm_Custom_Session::dedup_page_links called when running on a wiki page as a wiki plugin which takes care of incorrect Tiki links - Tiki uses page param for another thing so we replace with page_id. This is done for normal hrefs and links but not for javascript redirects. I see other modules the shortcuts one also have javascript redirects with ?page= attribute. I think we should add a JS function to do these redirects and overwrite it in Tiki environment, so it does the correct URLs before redirecting.

josaphatim: @kroky some changes have been made

kroky: Looks good, thanks.

Issue #940, opened 2024-04-03
composer create-project installing older version


ConnectGrid:
## 🐛 Bugreport

Running `composer require jason-munro/cypht` installs the latest version (1.4.1 at time of this writing). However, calling `composer create-project jason-munro/cypht` is installing v1.4.0, which has outdated dependencies that don't support PHP 8.3.x.

I'm not familiar with enough the `create-project` command to know how to fix this.

### Version & Environment

Rev: []

OS:  []

### Steps to reproduce
1. Run `composer create-project jason-munro/cypht`

On PHP 8.3, dependency validation fails.

marclaporte:
@christer77 has volunteered on this one.

Shadow243:
@ConnectGrid It could be due to caching.

Did you try this:`composer clear-cache` and then `composer create-project jason-munro/cypht` ??

marclaporte:
> On PHP 8.3, dependency validation fails.

What is the exact error message?

Cypht 1.4.x supports minimum PHP 5.6
Cypht 2.0.x (soon to be released) supports minimum PHP 7.4

It could be a pain to support PHP 5.6 to PHP 8.4 so we may want to put a max PHP version for Cypht 1.4

Can you try Cypht master? We did some work to support PHP 7.4 to PHP 8.4

christer77:
ERROR MESSAGE :
The error is not related to cypht but to the htmlpurifier package which was fixed to "^4.13" while this version does not support php 8.3.x, so since 2022-09-16 to fix some bugs including php 8.3.x support [htmlpurifier-releases](https://github.com/ezyang/htmlpurifier/releases)

christer77:
ERROR MESSAGE :
![cypht-jason-munro-error](https://github.com/cypht-org/cypht/assets/62720246/d9d75d8a-b4b9-497b-b01f-fe6a70c8b6bd)

marclaporte:
ok, so the next version of HTML Purifier will support PHP 8.4
https://github.com/ezyang/htmlpurifier/pull/396

christer77:
Exactly, and this since version 4.17, this means that I am returning the PR because [this](https://github.com/cypht-org/cypht/pull/943) of which is 4.16 is limited to 8.3

marclaporte:
@christer77 You need to generate a composer.lock file as well.

christer77:
It's noted @marclaporte

marclaporte:
[...]

PR #943, opened 2024-04-10
[UPD]composer.json: composer create-project installing older version

christer77:
[related issue](https://github.com/cypht-org/cypht/issues/940)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "ext-session": "*",
-        "ezyang/htmlpurifier": "^4.13",
+        "ezyang/htmlpurifier": "^4.17",
         "henrique-borba/php-sieve-manager": "^1.0.0",
         "league/commonmark": "^2.4",
         "paragonie/random_compat": "^2.0.18",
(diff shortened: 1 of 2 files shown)


Shadow243 (changes requested):
I actually see version 4.17 with bug fixes and some good reactions. Seems like they're making steady improvements.

[https://github.com/ezyang/htmlpurifier/releases](https://github.com/ezyang/htmlpurifier/releases)

marclaporte:
@Shadow243 Since composer.json has "ezyang/htmlpurifier": "^4.16", it should get 4.17, 4.18, etc. right?

marclaporte:
Yes, we are already getting: https://github.com/cypht-org/cypht/blob/master/composer.lock#L249

@christer77 I don''t understand how this MR fixes the issue

composer.json
```diff
@@ -32,7 +32,7 @@
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "ext-session": "*",
-        "ezyang/htmlpurifier": "^4.13",
+        "ezyang/htmlpurifier": "^4.16",
```
Shadow243, line 35: @christer77  Can you add composer.lock to this PR please!

composer.lock
marclaporte, line 1: ezyang/htmlpurifier doesn't change in composer.lock

PR #949, opened 2024-04-13
use CONFIG_PATH defined in FC consistently throughout

TheDigitalOrchard:
## Pullrequest

The front controller (index.php) defines a CONFIG_PATH constant, but this is not used throughout when accessing the directory. This commit corrects that so the path is set just once.

### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/lib/config.php
+++ b/lib/config.php
@@ -25,7 +25,7 @@ abstract class Hm_Config {
 
     /* flag to save config after page handling */
     public $save_on_login = false;
-    
+
     /**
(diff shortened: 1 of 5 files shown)


TheDigitalOrchard:
Note, my editor strips out trailing whitespace, so that was included in this commit unintentionally, but is a harmless change. Do you have a policy around that?

TheDigitalOrchard:
Confused. Temporarily closing.

PR #953, opened 2024-04-15
Fixed phpunit tests

josaphatim:

PR #954, opened 2024-04-15
Workflows: add test-selenium

wangxiaoerYah:
Like I said, it didn't really run. But I think this is an idea, and I hope to submit it to the main line for everyone to refer to.

This is just a workflow, not code, has no impact on the project, and is not a broken code. It is just a Workflow.


josaphatim:
> Need to merge #953, then it should work.

Already merged.

kroky:
You added dovecot and postfix setup in order to run the selenium email sending test - awesome! I like the code but we need to make it work in order to merge. Do you have any specific issues making it work?

marclaporte:
Well done @wangxiaoerYah !!!

PR #973, opened 2024-04-21
[FIX] Folders: fixing unable to set draft folder

Baraka24:
- We were unable to set draft folder as you can see from:

https://github.com/cypht-org/cypht/assets/81784141/2dc36aff-700a-4676-a70c-470adef198f2

- There was missing `style="z-index : 10"` on ul tag....adding it fixed as well see:

https://github.com/cypht-org/cypht/assets/81784141/f2b9b6d2-a526-42eb-b1b9-c55347a8c774

### DIFF
--- a/modules/imap_folders/modules.php
+++ b/modules/imap_folders/modules.php
@@ -504,7 +504,7 @@ class Hm_Output_folders_sent_dialog extends Hm_Output_Module {
                         <a href="#" class="select_sent_folder">'.$this->trans('Select Folder').'</a>:
                         <span class="selected_sent"></span>
                     </div>
-                    <ul class="folders sent_folder_select" style="z-index: 10">
+                    <ul class="folders sent_folder_select">
                         <li class="sent_title">
                             <a href="#" class="close">'.$this->trans('Cancel').'</a>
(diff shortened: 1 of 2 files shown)


modules/imap_folders/modules.php
```diff
@@ -592,7 +592,7 @@ protected function output() {
                         <a href="#" class="select_draft_folder">'.$this->trans('Select Folder').'</a>:
                         <span class="selected_draft"></span>
                     </div>
-                    <ul class="folders draft_folder_select">
+                    <ul class="folders draft_folder_select" style="z-index : 10">
```
Shadow243, line 595: It could be better if you place the style in `modules/imap_folders/site.css` for: draft_folder_select, trash_folder_select, sent_folder_select, etc., so you will not repeat it in the HTML tag.

Shadow243 (changes requested):
All good, @Baraka24

modules/imap_folders/modules.php
Baraka24: Okay well. I shall do so. Thanks @Shadow243

Issue #974, opened 2024-04-22
Declarative config for multiple accounts


jonocodes:
## 💬 Question

First off, thanks for this app. I'm very happy to have found it, as it fits the needs I am looking for on my new server.

I have about 10 IMAP accounts. I would like to set them up with a config file instead of using the UI. It looks like this can be done for only a single IMAP server via env var IMAP_AUTH_*.

Is there a recommended way to do this? Perhaps set it up once and use USER_CONFIG_TYPE=file
then copy that file to my new system?

Side note I am running this in NixOS which I only mention because declarative configs are the norm there.

marclaporte:
Thank you for the kind words.

Related: https://github.com/cypht-org/cypht/issues/470

For .env, let's ask the author @Shadow243 for some advice.

jonocodes:
Where are the imap creds stored now? I was not able to quickly find them. I saw there were some blobs in the mysql db. There?

Shadow243:
> > Thank you for the kind words.
> > Related: #470
> > For .env, let's ask the author @Shadow243 for some advice.
> 
> IMAP_AUTH_ is primarily used when AUTH_TYPE=IMAP to authenticate during the connection (on the login page) and of course to automatically add the server once connected.
> 
> Configuring multiple accounts via .env will complicate this file, but I think we can do it through a configuration file, as we did for LDAP Contact (config/ldap.php).
> 
> ```
> 'imap_accounts' => [
>     'Personal' => [
>         'server'  => '',
>         'port' => '',
>         'tls' => '',
>         'sieve' => '',
>         'username' => '',
>         'pwd' => ''
>     ],
>     'OtherName' => [
>         'server'  => '',
>         'port' => '',
>         'tls' => '',
>         'sieve' => '',
>         'username' => '',
>         'pwd' => ''
>     ],
> ]
> ```
> 
> This could allow us to anticipate the accounts (servers) to add on the server page automatically while retaining the IMAP_AUTH_ functionality with its initial function.

However, this scenario does not work in all cases. For example, we may have 10 IMAP accounts, but each account belongs to a different person. Therefore, each person must log in with their own account. By putting the accounts in the configuration file, everyone who is logged in will have access. However, this works in the case of someone managing (centralizing) his accounts in one place for himself

jonocodes:
@Shadow243 my goal is to use the unified inbox with all these accounts, so no separate logins.

marclaporte:
[...]

Issue #470, opened 2021-04-21
Export/Import of the user SMTP, IMAP etc, data 


thecuki:
## 🚀 Feature

It would be great if there is a feature that can export  servers data for one user, (then if needed edit it externally or internally with search and replace function), and then import that data to other user on the same cypht database.
The edit function is not important as long as user data can be downloaded unencrypted and manipulated on local computers, and then import back also unencrypted.

marclaporte:
@thecuki @dumblob Are you aware of a standard format for this?

I would love to export to a format which can be imported in https://www.claws-mail.org/  But others will want other tools, so maybe we can use (or propose) a standard format?

dumblob:
Before discussing formats, etc. I would like to clarify the motivation in form of listing well-defined use cases this should cover.

So far this sounds to me more like a one-off action which is usally better served by producing a nicely human-readable PDF with copyable text (not the character soup when copying from some PDFs).

marclaporte:
Use cases:
* Easily share email account access between team members (ex.: info@example.org) in Cypht or other email clients (webmail, mobile or desktop)
* Easily maintaining email access in a desktop client for offline access.
* Access to IMAP info for backups. Ex.: https://imapsync.lamiral.info/

dumblob:
Yep, these are pretty much the use cases where I would prefer a PDF (i.e. Cypht will show a very simple page with all this sensitive information revealed together with a precise time stamp and identification of the user who requested to view this page /probably the logged-in user/ and JS timer to go to Cypht previous page after a short timeout /perhaps 30 sec/ - this page can then easily be printed either on a printer or to PDF).

Basically this is only about credentials together with a very few settings (server IP/DNS, port numbers). The industry IMHO tries to not a have a standard for this as it would encourage one to use competitor's products/services.

In other words we are talking credentials management business here and not email business here :wink:. My strong preference is Bitwarden (or its open source counterpart [Vaultwarden](https://github.com/dani-garcia/vaultwarden/ )).

marclaporte:
https://wiki.mozilla.org/Thunderbird:Autoconfiguration:ConfigFileFormat

jonocodes:
I love this idea and the thunderbird config looks like a good direction since its used by several open source mail clients.

[...]

PR #976, opened 2024-04-25
Adding bulk-import accounts using csv sample file

Shadow243:
## Email Account Configuration for Cypht

This repository contains the account configuration settings for the Cypht. Unlike an approach based on directly storing configuration settings in the code (ex: config/account.php), we have opted to use a CSV file for configuring accounts.

Related Issue: [https://github.com/cypht-org/cypht/issues/974](https://github.com/cypht-org/cypht/issues/974)

### Data Isolation

By storing configuration settings in a CSV file, each user can have their own distinct configuration file. This allows for data isolation, ensuring that each user only has access to their own configured accounts and cannot access accounts configured by other users.

### Example CSV File Structure:
Screenshot 2024-04-25 at 20 59 10

### This is the display
Screenshot 2024-04-25 at 21 05 32

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,7 @@
         "paragonie/random_compat": "^2.0.18",
         "php": ">=8.1",
         "symfony/dotenv": "^4.3 || 5.4",
-        "symfony/yaml": "^5.4",
+        "symfony/yaml": "~6.4.3",
         "thomaspark/bootswatch": "^5.3",
         "twbs/bootstrap": "^5.3",
         "twbs/bootstrap-icons": "^1.11",
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
         "paragonie/random_compat": "^2.0.18",
         "php": ">=8.1",
         "symfony/dotenv": "^4.3 || 5.4",
+        "symfony/yaml": "^5.4",
         "thomaspark/bootswatch": "^5.3",
         "twbs/bootstrap": "^5.3",
         "twbs/bootstrap-icons": "^1.11",
--- a/modules/nux/assets/data/server_accounts_sample.yaml
+++ b/modules/nux/assets/data/server_accounts_sample.yaml
@@ -1,36 +1,36 @@
 Test1:
   jmap:
-    jmap_server:
+    server:
+    hide_from_combined_view: false
   user:
     username: email@example.org
     password: secret
   imap:
     port: 993
     tls: true
-    imap_server: imap.example.org
+    hide_from_combined_view: false
+    server: imap.example.org
   smtp:
-    support_smtp: true
-    smtp_server: smtp.example.org
-    smtp_port: 465
-    smtp_tls: true
+    server: smtp.example.org
+    port: 465
+    tls: true
   sieve:
-    sieve_host: tls://imap.example.org
-    sieve_port: 4190
+    host: tls://imap.example.org
+    port: 4190
 Test2:
   jmap:
-    jmap_server: 
+    server: 
   user:
     username: email@example.org
     password: secret
   imap:
     port: 993
     tls: true
-    imap_server: mail.example.org
+    server: mail.example.org
   smtp:
-    support_smtp: true
-    smtp_server: mail.example.org
-    smtp_port: 465
-    smtp_tls: true
+    server: mail.example.org
+    port: 465
+    tls: true
   sieve:
-    sieve_host: tls://mail.example.org
-    sieve_port: 4190
+    host: tls://mail.example.org
+    port: 4190
--- a/modules/nux/assets/data/server_accounts_sample.yaml
+++ b/modules/nux/assets/data/server_accounts_sample.yaml
@@ -1,36 +1,36 @@
-Migadu.com:
+Test1:
   jmap:
     jmap_server:
   user:
-    username: test@email.com
+    username: email@example.org
     password: secret
   imap:
     port: 993
     tls: true
-    imap_server: imap.migadu.com
+    imap_server: imap.example.org
   smtp:
     support_smtp: true
-    smtp_server: smtp.migadu.com
+    smtp_server: smtp.example.org
     smtp_port: 465
     smtp_tls: true
   sieve:
-    sieve_host: tls://imap.migadu.com
+    sieve_host: tls://imap.example.org
     sieve_port: 4190
-Postale.io:
+Test2:
   jmap:
     jmap_server: 
   user:
-    username: test@email.com
+    username: email@example.org
     password: secret
   imap:
     port: 993
     tls: true
-    imap_server: mail.postale.io
+    imap_server: mail.example.org
   smtp:
     support_smtp: true
-    smtp_server: mail.postale.io
+    smtp_server: mail.example.org
     smtp_port: 465
     smtp_tls: true
   sieve:
-    sieve_host: tls://mail.postale.io
+    sieve_host: tls://mail.example.org
     sieve_port: 4190
--- a/modules/nux/assets/data/server_accounts_sample.yaml
+++ b/modules/nux/assets/data/server_accounts_sample.yaml
@@ -1,0 +1,36 @@
+Migadu.com:
+  jmap:
+    jmap_server: 
+  user:
+    username: test@enterprisedepartment.com
+    password: 8xiK&z5TsQaJ!t
+  imap:
+    port: 993
+    tls: true
+    imap_server: imap.migadu.com
+  smtp:
+    support_smtp: true
+    smtp_server: smtp.migadu.com
+    smtp_port: 587
+    smtp_tls: true
+  sieve:
+    sieve_host: tls://imap.migadu.com
+    sieve_port: 4190
+Postale.io:
+  jmap:
+    jmap_server: 
+  user:
+    username: test@notrewiki.com
+    password: FuMxdfhCLD3vEuP5
+  imap:
+    port: 993
+    tls: true
+    imap_server: mail.postale.io
+  smtp:
+    support_smtp: true
+    smtp_server: mail.postale.io
+    smtp_port: 465
+    smtp_tls: true
+  sieve:
+    sieve_host: tls://mail.postale.io
+    sieve_port: 4190
--- a/modules/nux/assets/data/server_accounts_sample.yaml
+++ b/modules/nux/assets/data/server_accounts_sample.yaml
@@ -1,9 +1,9 @@
 Migadu.com:
   jmap:
-    jmap_server: 
+    jmap_server:
   user:
-    username: test@enterprisedepartment.com
-    password: 8xiK&z5TsQaJ!t
+    username: test@email.com
+    password: secret
   imap:
     port: 993
     tls: true
@@ -11,7 +11,7 @@ Migadu.com:
   smtp:
     support_smtp: true
     smtp_server: smtp.migadu.com
-    smtp_port: 587
+    smtp_port: 465
     smtp_tls: true
   sieve:
     sieve_host: tls://imap.migadu.com
@@ -20,8 +20,8 @@ Postale.io:
   jmap:
     jmap_server: 
   user:
-    username: test@notrewiki.com
-    password: FuMxdfhCLD3vEuP5
+    username: test@email.com
+    password: secret
   imap:
     port: 993
     tls: true
--- a/modules/nux/assets/data/server_accounts_sample.yaml
+++ b/modules/nux/assets/data/server_accounts_sample.yaml
@@ -1,15 +1,14 @@
-Test1:
+Mailbox 1:
+  username: email@example.org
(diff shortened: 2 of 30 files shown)


Shadow243:
I chose a CSV file instead of, for example:
```
<?php

return [
    /*
    | -------------------------------------
    | Constants used for auto configs servers and profiles
    | -------------------------------------
    |
    | These accounts are used to automatically configure the server and profile
    | When you add a new one don't forget to add env variables to .env file with the same name you are using here
    | just make sure names are unique to avoid conflicts. ex: PERSONAL_JMAP_SERVER
    | In case you are not putting env variables, you can set the value directly here
    | ex: 'server_name' => 'Migadu'
    |
    */
    'support_server_accounts' => env('SUPPORT_AUTO_ACCOUNTS_CONFIG', false),

    'server_accounts' => [
        'Personal' => [
            /*
            | When this is set to true, only server_name, email_or_username, password and host are required
            */
            'jmap_server' => env('PERSONAL_JMAP_SERVER', false),
            /*
            | This is the server name
            | ex: Migadu, Gandi, Postale, etc...
            */
            'server_name' => env('PERSONAL_SERVER_NAME', ''),
            /*
            | This is the server username or email
            */
            'email_or_username' => env('PERSONAL_SERVER_USERNAME', ''),
            /*
            | This is the server password
            */
            'password' => env('PERSONAL_SERVER_PWD', ''),
            /*
            | This is the port to used to connect to the server
            */
            'port' => env('PERSONAL_SERVER_PORT', 993),
            /*
            | Flag to enable or disable TLS connections
            */
            'tls' => env('PERSONAL_SERVER_TLS', true),
            /*
            | This is the host to connect to (the imap host address)
            */
            'host' => env('PERSONAL_SERVER_HOST', ''),
            /*
            | This is used to enable or disable the smtp support for this server
            */
            'support_smtp' => env('PERSONAL_SERVER_SUPPORT_SMTP', true),
            /*
            | if support_smtp is set to true, this is the smtp server to connect to
            */
            'smtp' => [
                /*
                | This is the smtp server to connect to
                */
                'server' => env('PERSONAL_SMTP_SERVER_HOST', ''),
                /*
                | This is the port to connect to the smtp server
                */
                'port' => env('PERSONAL_SMTP_SERVER_PORT', 465),
                /*
                | Flag to enable or disable TLS connections
                */
                'tls'=> env('PERSONAL_SMTP_SERVER_TLS', true)
            ]
        ],
    ],
];
```

kroky:
@Shadow243 thanks for this but that's not a declarative config. You still need UI to import the csv. It also has problems with storing passwords in plain text. However, if it opens use-cases when people want to bulk-import accounts, then it is fine.

Shadow243:
> @Shadow243 thanks for this but that's not a declarative config. You still need UI to import the csv. It also has problems with storing passwords in plain text. However, if it opens use-cases when people want to bulk-import accounts, then it is fine.

Most definitely @kroky, in this case I will keep the config way too. this will be a bit tricky though for multiple users in one instance.

Any suggestions about that ?

kroky:
As I suggested in #974 - user can copy either the file system settings or the db settings to another machine to make it easier to transfer multiple accounts that are already setup. I don't think we need any config updates in that case.

TheDigitalOrchard:
My 2¢, if such a plain-text file were to be introduced for this, I'd suggest using something like YAML which gives a richer structure and helps to avoid the "endless horizontal field list" that is prevalent in the world of using 2-dimensional tables (which we see with sprawling SQL tables, for example).

Just to show a snippet of the benefits of a format like YAML:

```yaml

Migadu:
    servers:
        jmap:
            hostname:  jmap.abc.com
        imap:
            hostname:  imap.abc.com
        smtp:
            hostname:  smtp.abc.com
```

It's human-readable while offering a richer structure that is future-proofed for growth.

But I agree with the concerns around storing credentials in a plaintext file. That's a no-no.

marclaporte:
Please also see https://wiki.mozilla.org/Thunderbird:Autoconfiguration:ConfigFileFormat and https://github.com/cypht-org/cypht/issues/470

modules/nux/assets/data/server_accounts_sample.yaml
```diff
@@ -0,0 +1,36 @@
+Migadu.com:
+  jmap:
+    jmap_server:
+  user:
+    username: test@email.com
```
marclaporte, line 5: Please use https://example.org for fake data

composer.json
```diff
@@ -38,6 +38,7 @@
         "paragonie/random_compat": "^2.0.18",
         "php": ">=7.4",
         "symfony/dotenv": "^4.3 || 5.4",
+        "symfony/yaml": "^5.4",
```
marclaporte, line 41: It's good that you took same component as Tiki: https://gitlab.com/tikiwiki/tiki/-/blob/master/vendor_bundled/composer.json

modules/nux/setup.php
```diff
@@ -75,6 +79,9 @@
         'nux_name' => FILTER_DEFAULT,
         'nux_pass' => FILTER_UNSAFE_RAW,
         'nux_account_name' => FILTER_DEFAULT,
-        'nux_all_inkl_login' => FILTER_DEFAULT
+        'nux_all_inkl_login' => FILTER_DEFAULT,
+        'accounts_source' => FILTER_DEFAULT,
+        'accounts_source' => FILTER_DEFAULT,
+        'accounts_csv' => FILTER_DEFAULT,
     )
```
marclaporte, line 86: so it is csv, yaml or both?

Shadow243: This on is using yaml for now. Should i add csv too or just remove any mention of it ??

marclaporte: @TheDigitalOrchard You suggested YAML. What do you think?

Shadow243: Any thought about both yaml and csv ?

marclaporte: > both yaml and csv ?

I like it!

composer.json
Shadow243: Done

modules/nux/setup.php
Shadow243: CSV and yaml are now both supported.

modules/nux/modules.php
```diff
@@ -260,47 +273,264 @@ public function process() {
...
+                        Hm_Msgs::add('ERRUsername and password are required for: ' . $key . ' server');
+                        return;
+                    }
+                    if ($value['jmap'] && !empty($value['jmap']['jmap_server'])) {
+                        if (!$this->module_is_supported('jmap')) {
+                            Hm_Msgs::add("ERRJMAP module is not enabled");
+                            return;
+                        }
+                        connect_to_imap_server(
+                            $value['imap']['imap_server'],
```
josaphatim, line 306: The key must be jmap

modules/nux/modules.php
```diff
@@ -260,47 +273,264 @@ public function process() {
...
+                }
+                foreach ($servers as $key => $value) {
+                    if (empty($value['user']['username']) or empty($value['user']['password'])) {
+                        Hm_Msgs::add('ERRUsername and password are required for: ' . $key . ' server');
+                        return;
+                    }
+                    if ($value['jmap'] && !empty($value['jmap']['jmap_server'])) {
+                        if (!$this->module_is_supported('jmap')) {
+                            Hm_Msgs::add("ERRJMAP module is not enabled");
+                            return;
```
josaphatim, line 303: For bulk import we shouldn't return. Also we must check for supported modules once not in the loop to avoid sending multiple times one message.

modules/nux/modules.php
```diff
@@ -260,47 +273,264 @@ public function process() {
...
+                        }
+                        connect_to_imap_server(
```
josaphatim, line 305: I think we can do it like this:
If IMAP and JMAP are set at the same time, save only JMAP.
Test if JMAP is defined, save it and retrieve its id, otherwise retrieve the IMAP id. If both are not defined, the SMTP is saved.
If IMAP/JMAP is defined and the server failed to register do not also register the corresponding SMTP.
On the other hand, if IMAP/JMAP is registered and the corresponding SMTP fails to be registered, also delete the corresponding IMAP/JMAP.
This way we will ensure that if we register a server all the corresponding information will be recorded and in the event of failure everything will not be saved.

modules/nux/modules.php
```diff
@@ -260,47 +273,264 @@ public function process() {
...
+                            $this
+                        );
+                        Hm_Msgs::add("Server " . $key . "  saved");
+                    }
+                }
+
+                $this->save_hm_msgs();
+            }elseif(!is_null($file) && $file['type'] == 'text/csv'){
+                try {
+                    $file = new SplFileObject($file['tmp_name']);
```
josaphatim, line 365: We can abstract this and make it a function which will return an array. Then other tests will be made here

modules/nux/assets/data/server_accounts_sample csv with data.numbers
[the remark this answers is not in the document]
Shadow243: Done

modules/nux/assets/data/server_accounts_sample.yaml
```diff
@@ -0,0 +1,36 @@
...
+    jmap_server:
+  user:
+    username: email@example.org
+    password: secret
+  imap:
+    port: 993
+    tls: true
+    imap_server: imap.example.org
+  smtp:
+    support_smtp: true
```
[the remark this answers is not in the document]
Shadow243: it is used here: `if ($server_data[8] && !empty($server_data[9])) {`

modules/nux/assets/data/server_accounts_sample.yaml
```diff
@@ -0,0 +1,36 @@
...
+    smtp_server: smtp.example.org
+    smtp_port: 465
+    smtp_tls: true
+  sieve:
+    sieve_host: tls://imap.example.org
+    sieve_port: 4190
+Postale.io:
```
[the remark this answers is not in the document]
Shadow243: Done

modules/nux/modules.php
```diff
@@ -246,8 +257,10 @@ public function process() {
 /**
  * @subpackage nux/handler
  */
-class Hm_Handler_get_nux_service_details extends Hm_Handler_Module {
-    public function process() {
+class Hm_Handler_get_nux_service_details extends Hm_Handler_Module
```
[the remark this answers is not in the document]
Shadow243: Done

modules/nux/modules.php
```diff
@@ -260,47 +273,264 @@ public function process() {
...
+        }
+        $title = $this->trans('Import from YAML or CSV file');
+        $notice = 'Please ensure your YAML or CSV  file follows the correct format';
+        $yaml_file_sample_path = WEB_ROOT . 'modules/nux/assets/data/server_accounts_sample.yaml';
+        $csv_file_sample_path = WEB_ROOT . 'modules/nux/assets/data/server_accounts_sample.csv';
+
+        return '<div class="quick_add_multiple_section">' .
+            '<div class="row"><div class="col col-lg-6"><div class="form-floating mb-3">' .
+            '<form class="quick_add_multiple_server_form" action="?page=servers" method="POST" enctype="multipart/form-data">' .
+            '<button class="browser_acc_file mt-2 btn btn-light" title="' . $notice . '"><i class="bi bi-filetype-yml me-2"></i>' . $title . '</button>' .
```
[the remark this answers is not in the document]
Shadow243: I'm reloading the page soo i am able to reread all servers.

modules/nux/modules.php
```diff
@@ -130,36 +141,34 @@ public function process() {
...
-                Hm_Msgs::add('ERR'.ucwords(str_replace('_', ' ', $this->request->get['error'])));
-            }
-            else {
+            } elseif (array_key_exists('error', $this->request->get)) {
+                Hm_Msgs::add('ERR' . ucwords(str_replace('_', ' ', $this->request->get['error'])));
+            } else {
                 Hm_Msgs::add('ERRAn Error Occurred');
             }
             $this->save_hm_msgs();
-            Hm_Dispatch::page_redirect('?page=servers');
```
josaphatim, line 145: This line is too important to redirect to servers page. This handler runs on home page

modules/nux/modules.php
```diff
@@ -199,12 +208,9 @@ public function process() {
                     $this->save_hm_msgs();
                     $this->session->close_early();
                     $this->out('nux_account_added', true);
-                    if ($this->module_is_supported('imap_folders')) {
```
josaphatim, line 202: Can you please review this MR? You delete accidentaly many lines.

Shadow243:
@josaphatim anything to change here ?

josaphatim:
@kroky can you check please ?

kroky (approved):
Looks good to me!

Issue #879, opened 2024-01-31
Release Cypht 2.0


marclaporte:
After 1.4.x, we will go to 2.0 because there are many big changes.

The main ones:

- https://github.com/cypht-org/cypht/pull/828
- https://github.com/cypht-org/cypht/pull/625
- https://github.com/cypht-org/cypht/pull/855

So let's clean this out: https://github.com/cypht-org/cypht/pulls, stabilize and release!

marclaporte:
Two blockers for Cypht 2.0:
- https://github.com/cypht-org/cypht/pull/855
- https://github.com/cypht-org/cypht/pull/921

marclaporte:
Also, we plan to have JMAP working for the basics: https://github.com/cypht-org/cypht/issues/180

And some more advanced features will be for later (ex.: Cypht 2.1)
https://github.com/cypht-org/cypht/issues/931

marclaporte:
The milestones are done except for JMAP: https://github.com/cypht-org/cypht/milestone/2

We have 4 non-draft MRs that it would make sense to add:
https://github.com/cypht-org/cypht/pulls?q=is%3Apr+is%3Aopen+draft%3Afalse

ConnectGrid:
This is exciting! Thanks for your hard work on this!

marclaporte:
We have 6 non-draft MRs that it would make sense to add:
https://github.com/cypht-org/cypht/pulls?q=is%3Apr+is%3Aopen+draft%3Afalse

marclaporte:
@josaphatim @kroky Let's clean this list and release 2.0
https://github.com/cypht-org/cypht/issues/879

marclaporte:
The remaining PRs are not blockers so we can release Cypht 2.0 any time!

Baraka24:
> The remaining PRs are not blockers so we can release Cypht 2.0 any time!

We are now experiencing Cypht 2.0 for  good notes.

marclaporte:
I did a round of testing on Cypht master, and I can tell you there is a huge number of "paper cut bugs"
https://en.wikipedia.org/wiki/Paper_cut_bug

Why it is buggy:
- We added tons of new features since 1.4.x
- Major changes like moving to Bootstrap 5 and new account interface
- Most recently active developers are new to the code base
- We don't yet have a good dev/testing/qa feeback loop (we'll work on this later)

[...]

PR #980, opened 2024-04-28
POC: Docker files in local directory

jonocodes:
This is a proof of concept how how to move and use docker into the main repo, as discussed here:
https://github.com/cypht-org/cypht-docker/issues/31#issuecomment-2081175722

It does not fully work since I have not lined up the env vars, but you can get it started and see the home page by running:

`docker compose up --build`

then visit http://localhost/

I have mostly copied over files from here: https://github.com/cypht-org/cypht-docker/tree/master/image
and I left in the commented out lines so you can see some of the changes I introduced.

docker-compose.yaml here gives several benefits:
1. It provides a runtime spec for the inputs required to run the program.
2. You can easily pull down the git repo and use it right away too see how cypht behaves in docker.
3. We can set it up such that this is also usable for a developer - such that they can work on the project without having php (and other dependencies) installed locally on the machine.
4. We can use docker-compose.yaml to build and distribute production images - likely triggered by a github action or something.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,13 +1,23 @@
 APP_NAME=Cypht
 
+# DB_CONNECTION_TYPE=host
+# DB_DRIVER=mysql
+# DB_PORT=
+# DB_HOST=localhost
+# DB_NAME=test
+# DB_USER=test
+# DB_PASS=123456
+# DB_SOCKET=/var/lib/mysqld/mysqld.sock
+
+# AUTH_USERNAME=admin
+# AUTH_PASSWORD=admin_password
 DB_CONNECTION_TYPE=host
+DB_HOST=db
+DB_NAME=cypht
+DB_USER=cypht
+DB_PASS=cypht_password
+SESSION_TYPE=db
 DB_DRIVER=mysql
-DB_PORT=
-DB_HOST=localhost
-DB_NAME=test
-DB_USER=test
-DB_PASS=123456
-DB_SOCKET=/var/lib/mysqld/mysqld.sock
 
 SESSION_TYPE=PHP
 AUTH_TYPE=DB
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -18,6 +18,7 @@ services:
       dockerfile: ./docker/Dockerfile
     volumes:
       - ./cypht/users:/var/lib/hm3/users
+    # TODO: add health check here
     ports:
       - "80:80"
     environment:
@@ -29,3 +30,5 @@ services:
       - CYPHT_DB_USER=cypht
       - CYPHT_DB_PASS=cypht_password
       - CYPHT_SESSION_TYPE=db
+
+  # TODO: add memcache and redis to this sample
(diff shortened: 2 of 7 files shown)


jonocodes:
I also added a few TODO notes on other things that could be updated.

marclaporte:
Thank you @jonocodes 

Good Docker support will be super good for the health of the project!

jonocodes:
Ok. Not sure the best way to proceed - in terms of review, acceptance, etc.
But for now I will continue on this POC and turn it into a working version along with my other suggestions.

marclaporte:
As the saying goes: "rough consensus and running code"

kroky:
I just noticed a huge docker entrypoint file trying to keep a hm3.ini file up to date - master and Cypht 2.0 is no longer using ini files for config - keeping the config in .env (or other env variables in dockerfiles) is just fine.

marclaporte:
> Where is cypht 2.0? Is that the same as master at this point?

Yes, it will be branched and released any day now: https://github.com/cypht-org/cypht/issues/879

marclaporte:
Inspiration from other projects:
- https://hub.docker.com/r/roundcube/roundcubemail/
- https://github.com/the-djmaze/snappymail/blob/master/docker-compose.yml

kroky:
> Where is cypht 2.0? Is that the same as master at this point?

2.0 has just been released. You can push your changes to master for upcoming releases (2.1+), backport/push to 2.0.x branch for 2.0+ releases and to 1.4.x for older cypht version.

wangxiaoerYah (commented):
The dockerfile is already inside the project and should no longer be downloaded using the wget archive. Also, you should keep a tight control over the files in the docker container, e.g. only copy the files that are relevant to the runtime. Also, the .hm3 configuration file entry has been deprecated and the dockerfile needs to be updated.

wangxiaoerYah:
And for consistency with the mainline, the env used inside docker should be copied directly from the project's env file, not pre-defined for it, and overridden later by environment variables or other means.

jonocodes:
> The dockerfile is already inside the project and should no longer be downloaded using the wget archive.

Agreed. Thats why I removed wget.

> Also, you should keep a tight control over the files in the docker container, e.g. only copy the files that are relevant to the runtime.

I have mixed feelings on that, as I am trying to simplify and streamline the docker process. I may address this with a .dockerignore file, but will think more on it.

> Also, the .hm3 configuration file entry has been deprecated and the dockerfile needs to be updated.

Yeah, I dont quite understand the .hm3 file yet, but will try to do away with it when I get to that part.

jonocodes:
> And for consistency with the mainline, the env used inside docker should be copied directly from the project's env file, not pre-defined for it, and overridden later by environment variables or other means.

Is that now what I did? I did not look too deeply at the env implementation, but that was my intention.

jonocodes:
I'm upgrading this POC to a WIP. See here:
https://github.com/cypht-org/cypht/pull/1001

PR #997, opened 2024-05-02
Handled providers that don't permit scripts inclusion

josaphatim:
Some email providers like Postale.io don't permit Sieve scripts to include other scripts ("no more than 0 includes allowed") this MR fixes that by combining all scripts into the main script.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "ext-openssl": "*",
         "ext-session": "*",
         "ezyang/htmlpurifier": "^4.17",
-        "henrique-borba/php-sieve-manager": "^1.0.2",
+        "henrique-borba/php-sieve-manager": "^1.0",
         "league/commonmark": "^2.4",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -33,7 +33,7 @@
         "ext-openssl": "*",
         "ext-session": "*",
         "ezyang/htmlpurifier": "^4.17",
-        "henrique-borba/php-sieve-manager": "^1.0.2",
+        "henrique-borba/php-sieve-manager": "^1.0",
```
marclaporte, line 36: Intentional?

josaphatim: Yes it was intentional. Is it good to lock to 1.0.3 ?

marclaporte: It doesn't matter in practice because composer.lock will grab 1.0.3 (soon 1.0.4) but I was just curious :-)

Issue #985, opened 2024-04-29
"Allow Images" link missing/hidden


PaulTGG:
### Version & Environment

Rev: [57756bc](https://github.com/cypht-org/cypht/commit/57756bc36e1b965323732eec098cf732860b1db6)

OS:  Debian 12.5

### Steps to reproduce
1. Install latest master.
2. Look at an email that has images.
3. There's no link to allow images.

Issue https://github.com/cypht-org/cypht/pull/873 was designed to allow images from a particular sender, but if the link is there it isn't obvious where.

marclaporte:
@jacob-js please advise :-)

PaulTGG:
@jacob-js More info - the links are only missing from the inline view.  Both links (allow, and always allow) appear in the not-inline view.

jacob-js:
> @jacob-js More info - the links are only missing from the inline view. Both links (allow, and always allow) appear in the not-inline view.

This could probably be the case. Let me give it a look.
Thanks @PaulTGG

jacob-js:
@PaulTGG please help me! How do you toggle the inline view?

PaulTGG:
@jacob-js I use these settings:
![setting](https://github.com/cypht-org/cypht/assets/12187131/7c579a9e-ba64-48ec-ad76-2a7f62ab44a1)

...and I open the email from the inbox by clicking on the subject line:
![inlineview](https://github.com/cypht-org/cypht/assets/12187131/089ce84c-e892-4a55-9a0c-d6c8bb9a7052)

jacob-js:
@PaulTGG suprisingly I don't have such settings.
<img width="1108" alt="Screenshot Settings" src="https://github.com/cypht-org/cypht/assets/68013195/7f03dccb-b275-467f-a91f-2cf21e59663a">

I don't understand.
Is your version other than the latest master?

PaulTGG:
@jacob-js I'm running the latest master: 492f634.  Also, I did not do an in-place upgrade.  I use a database, and my database didn't change, but the front-end cypht folder was completely rebuilt from scratch.

PaulTGG:
@jacob-js Do you have the "inline_messages" module active in your .env file?  That module is missing from the .env.example file, so you'd have to add it manually.

jacob-js:
> @jacob-js Do you have the "inline_messages" module active in your .env file? That module is missing from the .env.example file, so you'd have to add it manually.

Thank you! This one fixed.
[...]

PR #1002, opened 2024-05-04
Add external resources handling for inline message style

jacob-js:
https://github.com/cypht-org/cypht/issues/985

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -2469,3 +2469,112 @@ function getEmailProviderKey(email) {
 
     return "";
 }
+
+/**
+ * Allow external resources for the provided element.
+ *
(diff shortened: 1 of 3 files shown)


PaulTGG:
In order for this PR to work on inline messages, you first have to open a message individually, and not in inline view.  Then the added controls will show up in the inline view for that particular message.  They don't show up by default.

jacob-js:
> In order for this PR to work on inline messages, you first have to open a message individually, and not in inline view. Then the added controls will show up in the inline view for that particular message. They don't show up by default.

I agree with you @PaulTGG. https://github.com/cypht-org/cypht/pull/1014 addresses the problem.

PR #1004, opened 2024-05-06
[FIX]smtp/modules.php: Fix "Message contains more than the maximum number of recipients" on reply

IrAlfred:
This pull request fixes the message warning "Message contains more than the maximum number of recipients, proceed with caution" even when we select REPLY instead of REPLY-ALL for a mail that contains more than the maximum number of recipients in To, Cc, or Bcc


modules/smtp/modules.php
```diff
@@ -2157,15 +2166,20 @@ function default_smtp_server($user_config, $session, $request, $config, $user, $
  * @subpackage smtp/functions
  */
 if (!hm_exists('recip_count_check')) {
-function recip_count_check($headers, $omod) {
+function recip_count_check($headers, $omod, $reply_type='') {
```
josaphatim, line 2169: You can initialize $reply_type with reply_all as it was the default behavior.

modules/smtp/modules.php
```diff
@@ -28,8 +28,17 @@ public function process() {
                 $this->request->get['uid']
             );
             $reply_details = $this->session->get($cache_name, false);
+
+            $reply_type = '';
             if ($reply_details) {
-                recip_count_check($reply_details['msg_headers'], $this);
+                if (array_key_exists('reply', $this->request->get)) {
```
josaphatim, line 34: This same thing is repeated in Hm_Handler_load_smtp_servers_from_config::process. Can you make it a function ?

modules/smtp/modules.php
```diff
@@ -2157,15 +2166,20 @@ function default_smtp_server($user_config, $session, $request, $config, $user, $
...
-function recip_count_check($headers, $omod) {
+function recip_count_check($headers, $omod, $reply_type='') {
     $headers = lc_headers($headers);
     $recip_count = 0;
-    if (array_key_exists('to', $headers) && $headers['to']) {
-        $recip_count += count(process_address_fld($headers['to']));
-    }
-    if (array_key_exists('cc', $headers) && $headers['cc']) {
-        $recip_count += count(process_address_fld($headers['cc']));
+    if ($reply_type == 'reply') {
```
josaphatim, line 2172: This is is unnecessary you just and other condition to **if (array_key_exists('cc', $headers) && $headers['cc'])** that tests also if reply it is reply_all

josaphatim, line 2172: This is is unnecessary you just and other condition to the second if which tests cc to also tests if reply_type is reply_all

IrAlfred:
Thank you @josaphatim for your review. I work on it

modules/smtp/modules.php
IrAlfred: The proposed changes here currently fails if the number of email recipients in the "to" field exceeds the maximum limit (which is expected behavior). However, it works correctly when the maximum number of recipients is in the "cc" field. To ensure it works correctly in both cases, I have added conditions to handle both the "to" and "cc" fields. What do you think about this ?

lib/module.php
```diff
@@ -438,6 +438,27 @@ public function save_hm_msgs() {
         }
     }
 
+    /**
```
josaphatim, line 441: This function should be in modules/smtp/functions.php and follow the same structure as others

modules/smtp/modules.php
```diff
@@ -2157,15 +2152,19 @@ function default_smtp_server($user_config, $session, $request, $config, $user, $
  * @subpackage smtp/functions
  */
 if (!hm_exists('recip_count_check')) {
-function recip_count_check($headers, $omod) {
+function recip_count_check($headers, $omod, $reply_type='reply_all') {
     $headers = lc_headers($headers);
-    $recip_count = 0;
-    if (array_key_exists('to', $headers) && $headers['to']) {
+    $recip_count = 1;
```
josaphatim, line 2157: You should not change this

modules/smtp/modules.php
```diff
@@ -2157,15 +2152,19 @@ function default_smtp_server($user_config, $session, $request, $config, $user, $
...
+
+    if (array_key_exists('to', $headers) && $headers['to'] && $reply_type == 'reply_all') {
```
josaphatim, line 2159: This includes reply and reply all. It should stay as it was

PR #1005, opened 2024-05-07
 [UX] servers page : make servers page to be responsive on mobile

Bienvenumoringa:
Working on responsive design of servers page `?page=servers`, see screenshot : 
photo_2024-05-07_13-14-41

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -464,24 +464,25 @@ class Hm_Output_display_configured_imap_servers extends Hm_Output_Module {
             $res .= '<form class="imap_connect" method="POST">';
             $res .= '<input type="hidden" name="hm_page_key" value="'.$this->html_safe(Hm_Request_Key::generate()).'" />';
             $res .= '<input type="hidden" name="imap_server_id" class="imap_server_id" value="'.$this->html_safe($server_id).'" />';
-            $res .= '<div class="row"><div class="col-sm-2">';
+            $res .= '<div class="row m-0 p-0"><div class="col-xl-2 col-lg-2 col-md-6">';
             $res .= sprintf('
                 <div class="text-muted"><strong>%s</strong></div>
                 <div class="server_subtitle">%s/%d %s</div>',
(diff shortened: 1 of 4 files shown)


modules/imap/output_modules.php
```diff
@@ -464,34 +464,35 @@ protected function output() {
...
                 $this->html_safe($vals['name']), $this->html_safe($vals['server']), $this->html_safe($vals['port']),
                 $vals['tls'] ? 'TLS' : '' );
-            $res .= '</div><div class="col-sm-2">';
+
+            $res .= '</div> <div class="col-xl-7 col-lg-7 col-md-9 p-3 " > <div class="row"> <div class="col-xl-4 col-lg-4 col-md-4  px-2 py-2">';
 
             // IMAP Username
             $res .= '<div class="form-floating">';
-            $res .= '<input '.$disabled.' id="imap_user_'.$server_id.'" class="form-control credentials" type="text" name="imap_user" value="'.$this->html_safe($user_pc).'" placeholder="'.$this->trans('Username').'">';
+            $res .= '<input '.$disabled.' id="imap_user_'.$server_id.'" class="form-control " type="text" name="imap_user" value="'.$this->html_safe($user_pc).'" placeholder="'.$this->trans('Username').'">';
```
josaphatim, line 478: @Bienvenumoringa you should not remove class credentials as it is used in javascript. Just remove its css.

Bienvenumoringa: I've already put back the credentials class so that the js works as expected. And I've removed it from the css side.

josaphatim: @Bienvenumoringa this fix only for IMAP servers but not SMTP servers

Bienvenumoringa: Applied changes

josaphatim: @Bienvenumoringa there is too much space between 2 servers. Can you fix ?

Bienvenumoringa: Applied changes

Issue #1000, opened 2024-05-03
Being mistakenly warned about an empty email body


PaulTGG:
## 🐛 Bugreport

#866 was done to allow sending emails where there is no text in the body.  Since then, the compose page has been updated, and now, it's possible to get the message warning of an empty email body when there actually is text in the body.  Not sure if this also has something to do with the visual glitch where the word "message" appears over the toolbar (also seen in the screenshot below).

![send error](https://github.com/cypht-org/cypht/assets/12187131/6aba381a-5270-48dc-bc52-b880da80412e)

### Version & Environment

Rev: 492f634

### Steps to reproduce
1. Compose an email
2. Type in the recipient and subject
3. Add text to the body
4. Click Send.  The message pops up.

PaulTGG:
I still have the same issue with the latest master as of this moment (117f04b).
![bodynotempty](https://github.com/cypht-org/cypht/assets/12187131/9bf23189-8dfd-4fb5-beab-68030f07d4e9)

jacob-js:
@PaulTGG Sure, the PR is still open. On master, you should still be getting the same behavior.
Maybe try it directly from the pull request?

PaulTGG:
@jacob-js Sorry about that - I wasn't paying close attention.  I tried from the PR, and it works now, thanks!

PR #1006, opened 2024-05-07
Synchronize html editor and md editor with the compose_body textarea and adjust their appearance to match bs styling used

jacob-js:
https://github.com/cypht-org/cypht/issues/1000


marclaporte:
@PaulTGG confirms this works: https://github.com/cypht-org/cypht/issues/1000

PR #1008, opened 2024-05-07
change the order of options for account creation

David-Muhasa:
### Issues

- [X] None

### Checklist

- [X] None

### How2Test

- [X] None

### Todo

- [X] None

### DIFF
--- a/modules/feeds/setup.php
+++ b/modules/feeds/setup.php
@@ -16,7 +16,7 @@ add_handler('servers', 'load_feeds_from_config',  true, 'feeds', 'load_user_data
 add_handler('servers', 'process_add_feed', true, 'feeds', 'load_feeds_from_config', 'after');
 add_handler('servers', 'add_feeds_to_page_data', true, 'feeds', 'process_add_feed', 'after');
 add_handler('servers', 'save_feeds',  true, 'feeds', 'add_feeds_to_page_data', 'after');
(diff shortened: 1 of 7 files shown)


josaphatim:
@David-Muhasa If think this changes are wrong
They make modules tied. If you try to disable module wordpress you will notice that other sections in servers page will disappear as well

If think the best way to do this is:
**In modules/feeds/setup.php**
replace server_content_start by server_content_end and change the marker as well from after to before

**In modules/nux/setup.php**
Replace server_config_stepper_accordion_end_part by server_content_start

In this manner the order for sections will be
```
Add an E-mail Account
IMAP - JMAP - SMTP Servers
Feeds
NASA APIs
WordPress.com Connect
Github Connect
```

Please also handle the case when smtp, imap and jmap modules are not disabled
Hm_Output_server_config_stepper_end_part must be visible if at least one of those modules is enabled

David-Muhasa:
@josaphatim  I try to put your orientation into practice and then I will come back to you

David-Muhasa:
hello @josaphatim

marclaporte:
@David-Muhasa since you are new to the project, you should not merge your own PRs, but instead let @josaphatim or another experienced dev do so.

David-Muhasa:
It was by mistake. I directly called Mr Josaphat

Issue #756, opened 2023-09-05
Proper unicode support (was: Look into handling Unicode in folder names (was: "Offline" in the top right trying to access one account, others are fine))


rikberkelder:
## 💬 Question
I've gotten Cypht set up with three accounts. one gmail, one custom domain email on zoho, and one custom domain email on yandex. The first two work just fine. The yandex account, added through the "add an email account" wizard. It authenticates just fine, both smtp and imap work when i hit the test button. But when I click the account in the sidebar, a red square with the text "offline" appears in the top right. Any ideas what could be going wrong?

marclaporte:
I don't know but can you create a test email for a dev to try?

rikberkelder:
So the plot thickens a bit. I ended up migrating my yandex account over to zoho as well, which i'd been meaning to do. I then added the second zoho account the way i added the first one, which works fine. It authenticates and gets added to the list. To my surprise, trying to open it still pops up the "Offline" indicator in the top right.

What even is this indicator supposed to mean? Is there any place i can find logs as to what's happening?

marclaporte:
The debug mode is documented in https://github.com/cypht-org/cypht/blob/master/INSTALL

marclaporte:
> What even is this indicator supposed to mean?

Spin off task: https://github.com/cypht-org/cypht/issues/761

rikberkelder:
I didn't find the time to get the debug mode set up, but i did find the issue. The yandex account auto-generated some folder with cryllic characters. This also got migrated along to zoho. Cypht did not like this folder, and showed the "offline" message while failing to list the folders using the sidebar. I was still getting the emails from the account in the "everything" folder. I ended up deleting the folder, and now the account works fine. Might be worth looking into handling Unicode in folder names

marclaporte:
I've seen some nasty issues in my days when migrating mailboxes with folder names beyond ASCII. Sometimes, it was not even possible to rename the folder (!)

@kambereBr Please take a look to:
- Make sure Cypht has folder name support comparable to popular mail clients like Roundcube and Thunderbird.
- Review the "offline" message to see if it could have provided useful feedback in this case.

Thanks!

kambereBr:
@marclaporte I take a look at this.

Thanks!

marclaporte:
@kambereBr We need you

kambereBr:
PR: https://github.com/cypht-org/cypht/pull/1012

marclaporte:
[...]

PR #1012, opened 2024-05-08
Handle Unicode in folder names

kambereBr:
## Pullrequest
This PR ensures compatibility with multibyte characters, including Cryllic characters in folder names

### Issues
https://github.com/cypht-org/cypht/issues/756

### How2Test

msedge_w99l4OwJny


kroky:
It is OK to merge this as we already require mbstring extension in composer and sparingly use it mainly to convert encodings. However, if we want proper unicode support, we should do much more with the code. I see 170 different usages of substr only and there are a bunch of other string related functions that need to be converted to their multibyte counterparts - e.g. strtolower, strtoupper, strlen, strstr and probably others. @marclaporte maybe we merge this one now to fix the reported issue and release 2.0.1 and then open a PR for the rest of the places that need fixing for version 2.1.0?

Issue #687, opened 2023-03-08
MDN | DSN Missing in compose (?)


resu-cireneg:
## 🚀 Feature
2 Checkboxes , 
 ( as seen on tons of other mailers )

![Example](https://user-images.githubusercontent.com/69966873/223824859-cdf88544-7350-43cd-ab34-a9a8f2a91e37.png)

https://de.wikipedia.org/wiki/Message_Disposition_Notification
https://en.wikipedia.org/wiki/Bounce_message

### Design, Layout, Architecture

Make the PHP-Sending part set the 2 flags , implement 2 checkboxes

marclaporte:
I agree. Let's do it.

jasonmunro:
Personally I disagree, read receipt support was IMO never a good idea. If this gets added I implore you to require the recipient to opt-in each time a read receipt is delivered to a message sender. Requesting one on Send is OK I guess, but again not a fan of the whole idea.

marclaporte:
@resu-cireneg

marclaporte:
@resu-cireneg We need your feedback on current issue and https://github.com/cypht-org/cypht/pull/703

Else, we will close this as it's controversial.

marclaporte:
@resu-cireneg Last call

marclaporte:
This features in now in Cypht 2.0

For Cypht 2.1, it should be made optional (default off)

marclaporte:
@resu-cireneg 

This is a great topic for our monthly Cypht community meetings: https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings

PR #1015, opened 2024-05-09
[FIX] Disable by default delivery receipt in compose form

amaninyumu1:
[issues](https://github.com/cypht-org/cypht/issues/687#issuecomment-2087917218)

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -1133,7 +1133,7 @@ class Hm_Output_compose_form_content extends Hm_Output_Module {
                     '<textarea id="compose_body" name="compose_body" class="compose_body form-control" placeholder="'.$this->trans('Message').'">'.$this->html_safe($body).'</textarea>'.
                     (!$html ? '<label for="compose_body">'.$this->trans('Message').'</label>': '').
                 '</div>'.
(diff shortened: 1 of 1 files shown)


modules/smtp/modules.php
```diff
@@ -1133,7 +1133,7 @@ protected function output() {
                     '<textarea id="compose_body" name="compose_body" class="compose_body form-control" placeholder="'.$this->trans('Message').'">'.$this->html_safe($body).'</textarea>'.
                     (!$html ? '<label for="compose_body">'.$this->trans('Message').'</label>': '').
                 '</div>'.
-                '<div class="form-check mb-3"><input value="1" name="compose_delivery_receipt" id="compose_delivery_receipt" type="checkbox" class="form-check-input" /><label for="compose_delivery_receipt" class="form-check-label">'.$this->trans('Request a delivery receipt').'</label></div>';
+                '<div class="form-check mb-3"><input value="0" name="compose_delivery_receipt" id="compose_delivery_receipt" type="checkbox" class="form-check-input" /><label for="compose_delivery_receipt" class="form-check-label">'.$this->trans('Request a delivery receipt').'</label></div>';
```
josaphatim, line 1136: @amaninyumu1 This is the only one change needed here. Can you remove others before we merge ?

amaninyumu1: I think because I modified the default values ​​of the checkbox  from 1 to 0

amaninyumu1: I made some changes please review

Issue #817, opened 2023-11-02
IMAP server is not added to the auto-generated user profile


ivanov17:
## 🐛 Bugreport

After logging into the Cyhpt interface using IMAP authentication, the IMAP server is not added to the automatically generated user profile. Accordingly, the username is not recognized and is not added to the address and reply fields:

![Screen Shot 2023-11-02 at 21 24 43](https://github.com/cypht-org/cypht/assets/553855/74d09c3b-df5a-44d7-9ecb-ef804c5d40e1)

At the same time, the IMAP server is added to the servers and saved correctly:

![Screen Shot 2023-11-02 at 21 23 57](https://github.com/cypht-org/cypht/assets/553855/6a785413-b76f-442f-a843-e6a08c0be043)

On the `Compose` page I see the message `Please create a profile for saving sent messages`:

![Screen Shot 2023-11-02 at 21 25 55](https://github.com/cypht-org/cypht/assets/553855/65cb6024-95a2-48c8-acc7-1b0114622d8b)

Authentication is carried out using a full email address. 

The same thing happens with both an empty `default_email_domain` parameter and a specified domain.

### Version & Environment

Rev: `v1.4.0`

OS: CentOS Stream 9 container

### Steps to reproduce
1. Login to the Cypht interface using IMAP authentication and full email address
2. Check the `Profiles` page
3. Check the `Servers` page
4. Check the `Compose` page

marclaporte:
Can you see if you can find a link to "folders"?

ivanov17:
I hope I understand the question correctly. 

I can select the IMAP server on the "Folders" page: 

![Screen Shot 2023-11-03 at 12 56 48](https://github.com/cypht-org/cypht/assets/553855/d9847462-0624-497c-8f6a-a91391aec05d)

After that I can see a list of options and select all the required IMAP folders:

![Screen Shot 2023-11-03 at 12 59 52](https://github.com/cypht-org/cypht/assets/553855/947c389a-da7f-4891-ac1c-081fd8f37ca0)

I think my IMAP folders should already have all necessary special-use flags, since desktop clients (`Claws Mail` and `Thunderbird`) recognize their types correctly. Unfortunately Cypht doesn't do this, but I don't know if he has to do this or not.

marclaporte:
Ok, I see https://www.rfc-editor.org/rfc/rfc6154.html

@Yannick243 Please investigate. 

Related commit:
[...]

PR #855, opened 2024-01-03
Refactor/stmp imap server configuration

henochit:
### Issues

- [X] check this [reference](https://avan.tech/item81120-Cypht-Make-it-easier-to-set-up-IMAP-SMTP-and-connect-them-together-A-profile-with-optional-signature) 

### Checklist

- [X] None

### How2Test

- [X] Configure an SMTP and IMAP server at the same time (Enable Profile creation)
- [X] Configure an JMAP server
- [X] Configure an SMTP server only
- [X] Configure an IMAP server only

### Todo

- [X] None

PR #1022, opened 2024-05-14
Auto detect IMAP special folders when configuring account

Yannick243:
## PR
After [https://github.com/cypht-org/cypht/pull/625](https://github.com/cypht-org/cypht/pull/625), cypht was unable to auto detect IMAP special folders (Trash/Sent/Drafts) even though they exist on the server.
Based on "RFC 6154: IMAP LIST Extension for Special-Use Mailboxes"

### Related issue:
[https://github.com/cypht-org/cypht/issues/817](https://github.com/cypht-org/cypht/issues/817)

### DIFF
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -1103,8 +1103,9 @@ class Hm_Handler_quick_servers_setup extends Hm_Handler_Module {
 
                 Hm_Msgs::add("Server saved");
                 $this->out('just_saved_credentials', true);
-                $this->out('nux_server_id', $this->imap_server_id);
-                $this->out('nux_service_name', $provider);
+                $this->out('imap_server_id', $this->imap_server_id);
+                $this->out('imap_service_name', $provider);
+                $this->out('imap_folders_enabled', $this->module_is_supported('imap_folders'));
             }
         }
     }
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -1103,6 +1103,8 @@ class Hm_Handler_quick_servers_setup extends Hm_Handler_Module {
 
                 Hm_Msgs::add("Server saved");
                 $this->out('just_saved_credentials', true);
+                $this->out('nux_server_id', $this->imap_server_id);
+                $this->out('nux_service_name', $provider);
             }
         }
     }
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -1101,11 +1101,12 @@ class Hm_Handler_quick_servers_setup extends Hm_Handler_Module {
                     add_profile($profileName, $profileSignature, $profileReplyTo, $profileIsDefault, $email, $imapAddress, $this->smtp_server_id, $this->imap_server_id, $this);
                 }
 
-                Hm_Msgs::add("Server saved");
+                if ($this->module_is_supported('imap_folders')) {
+                    $this->out('imap_server_id', $this->imap_server_id);
+                    $this->out('imap_service_name', $provider);
+                }
                 $this->out('just_saved_credentials', true);
-                $this->out('imap_server_id', $this->imap_server_id);
-                $this->out('imap_service_name', $provider);
-                $this->out('imap_folders_enabled', $this->module_is_supported('imap_folders'));
+                Hm_Msgs::add("Server saved");
             }
         }
     }
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -2188,11 +2188,11 @@ function submitSmtpImapServer() {
         Hm_Notices.show(res.router_user_msgs);
 
         if (res.just_saved_credentials) {
-            if (res.nux_server_id) {
+            if (res.imap_server_id && res.imap_folders_enabled) {
                 Hm_Ajax.request(
                     [{'name': 'hm_ajax_hook', 'value': 'ajax_imap_accept_special_folders'},
-                    {'name': 'imap_server_id', value: res.nux_server_id},
-                    {'name': 'imap_service_name', value: res.nux_service_name}],
+                    {'name': 'imap_server_id', value: res.imap_server_id},
+                    {'name': 'imap_service_name', value: res.imap_service_name}],
                     function () {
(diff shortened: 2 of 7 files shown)


marclaporte:
> After #625, cypht was unable

I don't remember ever seeing this work before.  #625 arrived in Cypht 2.0.0

Can you confirm if this works OK in Cypht 1.4.x?

Thanks!

Yannick243:
The feature was introduced here: [https://github.com/cypht-org/cypht/pull/586](https://github.com/cypht-org/cypht/pull/586)
After testing on version 1.4.x, it appears that it was working partially (on some folders).

With this adjustment to 2.0.0, it is working with all folders when using the configuration form below.
CleanShot 2024-05-14 at 20 52 11@2x
As we can see the folders are automatically detected:
CleanShot 2024-05-14 at 20 53 08@2x

For now, the remaining part that is not working is when using the advanced configuration form below:
[https://imgur.com/a/HW4yFZZ](https://imgur.com/a/HW4yFZZ)

marclaporte:
Ok, but is it working because we are matching folder names or via "RFC 6154: IMAP LIST Extension for Special-Use Mailboxes"?

marclaporte:
@josaphatim any wisdom?

josaphatim:
> Ok, but is it working because we are matching folder names or via "RFC 6154: IMAP LIST Extension for Special-Use Mailboxes"?

> @josaphatim any wisdom?

RFC 6154 says that: This extension adds new optional mailbox attributes that a server may include in IMAP LIST command responses to identify special-use mailboxes to the client. This being said, special-use mailboxes will not always be present for all providers and some will only provide a few special-use mailboxes.

marclaporte:
Ok. So @Yannick243: please make sure to leverage RFC 6154 and not just folder names.

marclaporte:
Sounds good to me. @josaphatim?

josaphatim:
@Yannick243 why this still in draft ?

Yannick243:
This is not working when using the new advanced configuration form [https://github.com/cypht-org/cypht/pull/855](https://github.com/cypht-org/cypht/pull/855). 
That's what I am fixing.

Yannick243:
The PR is ready for a review.

modules/core/site.js
```diff
@@ -2188,34 +2188,49 @@ function submitSmtpImapServer() {
...
-            $("#srv_setup_stepper_enable_sieve").prop('checked', false);
-            $("#srv_setup_stepper_only_jmap").prop('checked', false);
-            $('#step_config-imap_bloc').show();
-            $('#step_config-smtp_bloc').show();
-            $('#srv_setup_stepper_profile_bloc').show();
-
-            Hm_Utils.set_unsaved_changes(1);
-            Hm_Folders.reload_folders(true);
-            location.reload();
+            if (res.nux_server_id) {
```
josaphatim, line 2191: @Yannick243 just have some requests

Can you rename nux_server_id and nux_service_name as we are not in nux module here ?

Also what if imap_folders module is not enabled ? Can fix also for the nux module ?

Last thing we had a confirmation dialog before setting folders https://github.com/cypht-org/cypht/commit/ac3b9c283b8b8c5094f7362adacd8e0368c1c9d6#diff-3e95247a6dbbf437a9b5e7341db84da72c039b7b5ad532928b4cd8678dd5ff17R45. Shall we keep it for both or remove it for nux module too ? @marclaporte can you advise on this ? For my hand I'd prefer to remove it for nux module.

marclaporte: I also prefer we remove this confirmation dialog. People can easily change the folders after if they wish (which is super rare).

Yannick243: Hello @josaphatim,
The changes have been applied.

modules/core/handler_modules.php
```diff
@@ -1103,6 +1103,9 @@ public function process() {
 
                 Hm_Msgs::add("Server saved");
                 $this->out('just_saved_credentials', true);
+                $this->out('imap_server_id', $this->imap_server_id);
```
josaphatim, line 1106: @Yannick243 we can avoid adding imap_folders_enabled in allowed output as we don't really need it in js with something like:

```
if ($this->module_is_supported('imap_folders')) {
    $this->out('imap_server_id', $this->imap_server_id);
    $this->out('imap_service_name', $provider);
}
```

modules/core/site.js
```diff
@@ -2188,34 +2188,49 @@ function submitSmtpImapServer() {
...
-            $("#srv_setup_stepper_enable_sieve").prop('checked', false);
-            $("#srv_setup_stepper_only_jmap").prop('checked', false);
-            $('#step_config-imap_bloc').show();
-            $('#step_config-smtp_bloc').show();
-            $('#srv_setup_stepper_profile_bloc').show();
-
-            Hm_Utils.set_unsaved_changes(1);
-            Hm_Folders.reload_folders(true);
-            location.reload();
+            if (res.imap_server_id && res.imap_folders_enabled) {
```
josaphatim, line 2191: Also change this to `if (res.imap_server_id)`

modules/imap_folders/modules.php
```diff
@@ -129,6 +129,12 @@ public function process() {
  */
 class Hm_Handler_process_accept_special_folders extends Hm_Handler_Module {
     public function process() {
+        
+        if (!$this->module_is_supported('imap_folders')) {
+            Hm_Msgs('ERRIMAP folders module is not enabled');
```
josaphatim, line 134: I think we can just return as users won't have a clue on this or add it or debug instead.

modules/nux/site.js
```diff
@@ -47,16 +47,18 @@ var nux_add_account = function() {
 
 var display_final_nux_step = function(res) {
     if (res.nux_account_added) {
-        if (confirm('Do you accept special folders?')) {
+        if (res.nux_server_id && res.nux_imap_folders_enabled) {
```
josaphatim, line 50: Same here

Yannick243: Done

PR #1042, opened 2024-05-22
Creating script to locally run unit tests in sqlite

jonocodes:
## Pullrequest

This decouples tests from github CI. The run.sh script can be run anywhere and uses environment variables to run instead of modifying config files in place.

### DIFF
--- a/tests/phpunit/data/schema.sql
+++ b/tests/phpunit/data/schema.sql
@@ -1,0 +1,6 @@
+
+CREATE TABLE IF NOT EXISTS hm_user (username varchar(255), hash varchar(255), primary key (username));
+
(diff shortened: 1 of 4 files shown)


marclaporte:
Super!

FYI, we'll soon share some developer documentation and this will make a great addition!

marclaporte:
@jonocodes I look forward to your next steps :-)

jonocodes:
So this only automates unit tests, not selenium. And it uses sqlite only - which I think is sufficient for unit tests. If people agree, I can fairly quickly finish this.

kroky (approved):
Yes, these are good simplifications for running the unit tests. I have created a mysql db to run these but defaults to sqlite3 will solve this dependency.

PR #1051, opened 2024-05-26
Unicode Support: Replace Standard PHP String Functions with Multibyte Counterparts

kambereBr:
This PR refactors standard PHP string functions to their multibyte counterparts to improve Unicode support.

Related PR: https://github.com/cypht-org/cypht/pull/1012

### DIFF
--- a/lib/crypt.php
+++ b/lib/crypt.php
@@ -160,7 +160,7 @@ class Hm_Crypt_Base {
     */
     private static function hash_equals($a, $b) {
         $res = 0;
-        $len = mb_strlen($a);
+        $len = mb_strlen($a, '8bit');
         for ($i = 0; $i < $len; $i++) {
             $res |= ord($a[$i]) ^ ord($b[$i]);
         }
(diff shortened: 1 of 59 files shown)


marclaporte:
In terms of http://semver.org/ will the next release be a x, y or z?

kambereBr:
> In terms of http://semver.org/ will the next release be a x, y or z?

The next release will be a **y** release according to Semantic Versioning (SemVer) principles at http://semver.org/. This means a minor version increment due to the introduction of enhanced Unicode support with multibyte string functions. 

Could you please confirm if this aligns with your understanding?

marclaporte:
That sounds OK.

lib/crypt.php
```diff
@@ -76,18 +76,18 @@ public static function plaintext($string, $key) {
         $string = base64_decode($string);
 
         /* bail if the crypt text is invalid */
-        if (!$string || strlen($string) <= 200) {
+        if (!$string || mb_strlen($string, '8bit') <= 200) {
```
kroky, line 79: Why 8bit? If the encoded string is something only cypht does, we can keep the old non-mb functions as they are 8bit by default, aren't they?

kroky: I assume this would be true for all 8bit usages in the following code - if these are hashes that are not multibyte strings, simply use the non-mb functions.

kambereBr: Sure, they are. I've replaced the 8bit usages with non-mb functions.

Thanks!

marclaporte:
So the next version will be Cypht 2.2.0

marclaporte:
This is partially reverted here: https://github.com/cypht-org/cypht/pull/1217

marclaporte:
Spin off: https://github.com/cypht-org/cypht/issues/1224

Issue #1024, opened 2024-05-15
"Everything" view not updating correctly


PaulTGG:
## 🐛 Bugreport
The unified inbox seems to have a caching issue.  If you move/delete an email using a different email app, or using the individual mailbox view in Cypht, the "Everything" view may not update with the change.  The change will show in the individual mailbox view, but not in the unified inbox.  In order to force the inbox to update, you have to start a new session (restarting your browser, for example).  I'm using redis for caching.

EDIT: Clicking the refresh icon on the Everything page does not fix the problem, and refreshing the browser also does not work.

EDIT 2: This is still happening in 2.0.1.  I've updated the Rev# below.

EDIT 3: Updated the Rev below.

### Version & Environment

Rev: a13e77e

PaulTGG:
Let me know if I can provide more info.  This is a bug one of the best features of Cypht!

josaphatim:
Let me take a look at it

josaphatim:
@PaulTGG There is already a fix for this: https://github.com/cypht-org/cypht/pull/1055

PR #1055, opened 2024-05-29
Fixed everything view not updating correctly

josaphatim:
Fixes https://github.com/cypht-org/cypht/issues/1024

Integer math operation on strings returns NaN and made the code miss elements to delete.


kroky:
This probably came from the move away from integer IDs.

Issue #178, opened 2017-03-21
webhook notification


mazhigali:
Can you add webhook notification while new email comes? 
I would like to use telegram notifications.
Thank you for excellent application and great job

jasonmunro:
Thank you for your kind words!
Interesting idea for a module set. Just to clarify, when new messages are detected we would do some sort of authenticated POST to a webhook for a system like Telegram? I presume this is the API we would want to leverage for that: https://core.telegram.org/bots ?

mazhigali:
Yes you are right. I'm using this script https://github.com/rakshazi/telegram-notify. But I think within the module sypth you can come up with a more functional solution. For example, in the message to send the body of the letter and header, And also be able to remove the email from the telegrams like this @GmailBot

marclaporte:
For the record:
https://en.wikipedia.org/wiki/Webhook

FYI, we just added Webhooks to Tiki Wiki CMS Groupware:
https://gitlab.com/tikiwiki/tiki/-/merge_requests/1649/

marclaporte:
Extending this feature request: I would like notifications for emails that fit a specific criteria. So we could add this to https://github.com/cypht-org/cypht/tree/master/modules/saved_searches

marclaporte:
@mazhigali We will soon have something ready for testing: https://github.com/cypht-org/cypht/pull/1056

airbornelamb:
@marclaporte this is very exciting. Will this work only for telegram or generic webhooks as well?

Generic webhooks would allow Zapier/Make.com usage so that people can shoot the notifications off to other places as they desire. Opens up some powerful use cases.

marclaporte:
@airbornelamb Thank you for the kind words.

I agree. Can you describe a realistic use case so @Shadow243 can handle it?

airbornelamb:
For sure, the biggest use case in front of me would be to send the notification off to Make.com using their generic webhook catch in real-time or near real-time from the message being received. Let me describe in more JTBD...

[...]

PR #1056, opened 2024-05-31
[ENH] Add telegram notification webhook when new messages are detected

Shadow243:
Related issue: [https://github.com/cypht-org/cypht/issues/178](https://github.com/cypht-org/cypht/issues/178)


modules/imap/handler_modules.php
```diff
@@ -1304,6 +1314,17 @@ public function process() {
             $this->out('folder_status', $status);
             $this->out('imap_unread_data', $msg_list);
             $this->out('imap_server_ids', $form['imap_server_ids']);
+
+            $webhook_token = $this->user_config->get('webhook_token_setting');
+            if($webhook_token && !empty($webhook_token)) {
+                $extracted_msgs = array();
+                foreach ($msg_list as $msg) {
```
Shadow243, line 1321: I'm not sure this is the right way to do this, a loop: if I receive 50 messages should I therefore receive 50 notifications in Telegram ?

Baraka24: @josaphatim, @marclaporte what are your thoughts?

marclaporte: I am sure @nekohayo will have some more thoughtful ideas but here is a first one:

Maybe with a 5 minute buffer? Thus: accumulate all the emails, and when no new email has arrived for 5 minutes (configurable), send one message: "You have received 50 emails". If messages don't stop, send one message every 5 minutes.

nekohayo: Does the API lets you reuse & edit (or auto-delete/auto-timeout the previous) notification message sent to the chat while the user has not yet read it? Otherwise, other than having some "exponential" back-off (ex: 5 mins, 10, 20, 40, 60, 90...) while messages are not read by the user on chat (if the API allows you to get read receipts, but I kinda doubt it), I don't see a much more creative way than what Marc has proposed here so far.

modules/imap/handler_modules.php
```diff
@@ -1304,6 +1314,20 @@ public function process() {
             $this->out('folder_status', $status);
             $this->out('imap_unread_data', $msg_list);
             $this->out('imap_server_ids', $form['imap_server_ids']);
+
+            $webhook_token = $this->user_config->get('webhook_token_setting');
+            $msg_count = count($msg_list);
+            $email_to = $msg_list[0]['to'];
+            if ($msg_count > 0) {
+                $interval = 5 * 60;
```
Shadow243, line 1322: 5 must be configurable according to @marclaporte comment.

Baraka24: Okay I shall apply changes. Thanks @Shadow243.

Shadow243:
@Baraka24 your requested change is ready here: https://github.com/cypht-org/cypht/pull/1130

Baraka24:
Closing in favor of https://github.com/cypht-org/cypht/pull/1196

Issue #13, opened 2015-08-04
Enhancement: support for user tagging (tags / labels)


dumblob:
Some very basic tagging/labeling (so that it doesn't bloat the user interface) with filtering based on these tags/labels would be nice to have.

The implementation would use just IMAP flags/keywords (see an older short article with useful hints: http://deflexion.com/2006/05/server-side-message-labels). And of course, I wouldn't pay attention to Mozilla-based clients etc. and just left the default tag set blank without any presets. I.e. solely up to the user what the tags will be (if the user wanted old-thunderbird-compatible tags, then he can just write them as `$My_awesome_tag4` to exchange the fourth Mozilla tag with this one).

jasonmunro:
Having used Cypht since before it was really usable - I'm needing this feature soon. This is a good sign that it will actually get done ;). "flagged" works in the interim, but tags/labels and the ability to "promote" one to the main folder list section would be really cool. Using IMAP flags/keywords is definitely the way to go (maybe with some Gmail optimizations if we can leverage them). We will have to come up with work-arounds for other data sources (POP3, feeds, etc), but it's the same with read/unread status, so no biggie :). BTW, I really appreciate all your thoughtful feedback!

jasonmunro:
Been looking into this today. I want to use IMAP custom keywords so we can store the labels like flags on the IMAP server itself. Not all services support this however, so we will have to work around those that don't (outlook365 does not, surprise!). Looks like Gmail and many of the more popular Open Source servers do support it (dovecot, cyrus, uw, courrier) but it's hit or miss. We could fall back to storing label information locally for services that don't support it (and for content that is not from IMAP) but I think to start I will build it around the IMAP custom flag ability and see how it goes.

dumblob:
Great news! I'll take a look at the progress asap.

jasonmunro:
nothing to see yet, except for basic support in the IMAP lib for setting custom keyword flags. Any thoughts on how you would incorporate this into the UI?

dumblob:
Well, that's a tough question ;) I don't have any strong opinions on this, but I see two main use-cases:

1. adding, removing, and editing tags to/from emails and also the tags themself

[...]

PR #1058, opened 2024-06-03
Adding label/tags to cypht

Shadow243:
Related issue: [https://github.com/cypht-org/cypht/issues/13](https://github.com/cypht-org/cypht/issues/13)

Screenshot 2024-06-04 at 11 58 26

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -499,7 +499,8 @@ function Message_List() {
         'junk': 'formatted_junk_data',
         'trash': 'formatted_trash_data',
         'sent': 'formatted_sent_data',
-        'drafts': 'formatted_drafts_data'
+        'drafts': 'formatted_drafts_data',
+        'tag': 'formatted_tag_data'
     };
 
     this.run_callbacks = function (completed) {
@@ -1155,6 +1156,7 @@ function Message_List() {
     this.set_junk_state = function() { self.set_message_list_state('formatted_junk_data'); };
     this.set_trash_state = function() { self.set_message_list_state('formatted_trash_data'); };
     this.set_draft_state = function() { self.set_message_list_state('formatted_drafts_data'); };
+    this.set_tag_state = function() { self.set_message_list_state('formatted_tag_data'); };
 };
 
 /* folder list */
@@ -1545,7 +1547,7 @@ var Hm_Utils = {
         var results = {}
         var i;
         var hash = window.location.hash;
-        var sections = ['.wp_notifications_setting', '.github_all_setting', '.tfa_setting', '.sent_setting', '.general_setting', '.unread_setting', '.flagged_setting', '.all_setting', '.email_setting', '.junk_setting', '.trash_setting', '.drafts_setting'];
+        var sections = ['.wp_notifications_setting', '.github_all_setting', '.tfa_setting', '.sent_setting', '.general_setting', '.unread_setting', '.flagged_setting', '.all_setting', '.email_setting', '.junk_setting', '.trash_setting', '.drafts_setting','.tag_setting'];
         for (i=0;i<sections.length;i++) {
             dsp = Hm_Utils.get_from_local_storage(sections[i]);
             if (hash) {
@@ -1574,7 +1576,7 @@ var Hm_Utils = {
         return res;
     },
 
-    seach_from_local_storage: function(pattern) {
+    search_from_local_storage: function(pattern) {
         const results = [];
         const key_pattern = new RegExp(pattern);
         for (let i = 0; i < sessionStorage.length; i++) {
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -1574,6 +1574,19 @@ var Hm_Utils = {
         return res;
     },
 
+    seach_from_local_storage: function(pattern) {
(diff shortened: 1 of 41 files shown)


marclaporte:
This was requested 9 years ago, so I expect it will make a lot of people happy! https://github.com/cypht-org/cypht/issues/13

Can we envisage that when Cypht is integrated with other apps (like Tiki), it shares tags/labels? We don't need now, but I expect we will want later.

kroky:
Looks good so far. I'll review once this is ready out of draft mode.

Shadow243:
> I'll review once this is ready out of draft mode.

Ready for review @kroky

modules/core/site.js
```diff
@@ -1573,6 +1575,19 @@ var Hm_Utils = {
         return res;
     },
 
+    seach_from_local_storage: function(pattern) {
```
kroky, line 1578: Search is misspelled here. It seems rather inefficient to look through all the entries in sessionStorage. Also, save_to_local_storage checks for Storage != void(0) to prevent other JS errors, we should do the same here, I think.

What is the purpose of this function? You seem to load all the stored imap source keys in the storage and save back with value of 1. Not sure I understand the need to do this. Do you invalidate some kind of caching?

modules/imap/functions.php
```diff
@@ -1321,6 +1321,49 @@ function snooze_message($imap, $msg_id, $folder, $snooze_tag) {
...
+    }
+    $msg = "X-Cypht-Labels:".implode(',', $tags)."\n".$msg;
+
+    $msg = str_replace("\r\n", "\n", $msg);
+    $msg = str_replace("\n", "\r\n", $msg);
+    $msg = rtrim($msg)."\r\n";
+
+    $res = false;
+
+    if (!count($imap->get_mailbox_status($folder))) {
```
kroky, line 1352: If you just downloaded a message from the folder it seems impossible for the folder to not exist and create it here. Better leave the folder checking at the top and if it doesn't exist return early. Otherwise, just use the folder. We can't create a folder to download a message from it.

modules/imap/functions.php
```diff
@@ -1321,6 +1321,49 @@ function snooze_message($imap, $msg_id, $folder, $snooze_tag) {
...
+        $imap->create_mailbox($folder);
+    }
+    if ($imap->select_mailbox($folder) && $imap->append_start($folder, strlen($msg))) {
+        $imap->append_feed($msg."\r\n");
+        if ($imap->append_end()) {
+            if ($imap->select_mailbox($folder) && $imap->message_action('DELETE', array($msg_id))) {
```
kroky, line 1358: I think you should only select the mailbox once at the top of the method. No need to further select the same folder as you are working in one folder only.

modules/imap/functions.php
```diff
@@ -1417,6 +1460,27 @@ function snooze_dropdown($output, $unsnooze = false) {
     return $txt;
 }}
 
+if (!hm_exists('tags_dropdown')) {
+function tags_dropdown($context, $headers) {
+    $folders = $context->get('tag_folders', array());
```
kroky, line 1465: These seem to be tag names or labels, why do we call them tag folders? Better call them just labels or tags.

modules/tags/functions.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+use PhpParser\Node\Expr\AssignOp\Div;
```
kroky, line 3: Where is this used here?

modules/tags/site.js
```diff
@@ -0,0 +1,19 @@
...
+    //     Hm_Ajax.request(
+    //         form.serializeArray(),
+    //         function(res) {
+    //             console.log("res", res)
+    //             // if (res.) {
+    //             // }
+    //         }
+    //     );
+        
+    // });
```
kroky, line 16: This seems like debugging code - can you please remove it?

modules/core/site.js
Shadow243: > Do you invalidate some kind of caching?

Yes exactly, when I access the message page a cache like: 62_imap_xxxxxxxx_xxxx is created. The problem is that the next time when I open this email again it will be retrieved from the cache so if I add a label I won't be able to get it updated. This is why I added a function that allows me to search the cache for anything that has this format: int_imap_xxxxxxxx_xxxx and delete them

modules/imap/functions.php
Shadow243: Done

modules/imap/functions.php
Shadow243: Removed

modules/imap/functions.php
Shadow243: Done, i'm now calling them `labels`

modules/tags/functions.php
Shadow243: Removed

marclaporte:
Reading the code, I see both tag and label being used. Please standardize to one.

It's OK to have code comment which indicates the alternative name for searchability.

Shadow243:
> Please standardize to one.

@marclaporte I'm now using tag/tags.

PR #1059, opened 2024-06-04
Update docker image from php 7 to 8

jonocodes:
## Pullrequest

:)


marclaporte:
Thank you @jonocodes  I am really happy the minimum version will be PHP 8.1 (and those who want older PHP can use LTS)

PR #1061, opened 2024-06-04
Removing CRAM-MD5 mechanism

Danelif:
This PR aims to remove  CRAM-MD5 mechanism in the code base of cypht

[Issue](https://github.com/cypht-org/cypht/issues/355)

PR #1062, opened 2024-06-04
Revert "Removing CRAM-MD5 mechanism"

Danelif:
Reverts cypht-org/cypht#1061


Neustradamus:
Note: Replaced by https://github.com/cypht-org/cypht/pull/1063

PR #1065, opened 2024-06-05
[EHN]Like Gmail, warn if you are about to reply to a new contact (and perh…

christer77:
Feature to validate CC contacts if they exist in the user contact list. Being default, if the user disables this setting, will no longer be notified before the message is sent.

### DIFF
--- a/modules/contacts/modules.php
+++ b/modules/contacts/modules.php
@@ -144,6 +144,8 @@ class Hm_Handler_load_contacts extends Hm_Handler_Module {
         }
         $this->out('contact_page', $page);
         $this->out('contact_store', $contacts, false);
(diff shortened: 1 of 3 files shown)


Shadow243 (changes requested):
@marclaporte is it enough to just give an alert? we could make it possible to send by directly recording the contact instead of just alerting.

marclaporte:
> is it enough to just give an alert

So alert the first time, and after it's recorded as a approved recipient?

Shadow243:
> > is it enough to just give an alert
> 
> So alert the first time, and after it's recorded as a approved recipient?

Yes

christer77:
> @marclaporte is it enough to just give an alert? we could make it possible to send by directly recording the contact instead of just alerting.

Hello @Shadow243, @marclaporte 

Please enable this and see if it doesn't match your suggestion.

add-automaticaly-adresse-in-contact-list

PR #1077, opened 2024-06-10
[FIX]add special folders migadu

christer77:
Fix add special folders after create account(Migadu)

Before:
before-special-foldres

After

after-special-foldres


modules/nux/site.js
```diff
@@ -47,18 +47,20 @@ var nux_add_account = function() {
...
-        if (confirm('Do you accept special folders?')) {
-            Hm_Ajax.request(
-                [{'name': 'hm_ajax_hook', 'value': 'ajax_imap_accept_special_folders'},
-                {'name': 'imap_server_id', value: res.nux_server_id},
-                {'name': 'imap_service_name', value: res.nux_service_name}],
-                function(res) {
-                    window.location.href = "?page=servers";
-                }
-            );
+        if (res.nux_account_added) {
```
kroky, line 50: What is the purpose of checking res.nux_account_added twice?

modules/imap_folders/modules.php
```diff
@@ -140,18 +140,25 @@ public function process() {
...
             } else {
                 $specials[$form['imap_server_id']] = array('sent' => '', 'draft' => '', 'trash' => '', 'archive' => '', 'junk' => '');
-            }     
+            }
             $this->user_config->set('special_imap_folders', $specials);
 
             $user_data = $this->user_config->dump();
             $this->session->set('user_data', $user_data);
-            
+
```
kroky, line 161: Please be careful with whitespace - we don't want lines containing trailing whitespaces.

christer77: noted

modules/nux/site.js
```diff
@@ -52,13 +52,12 @@ var display_final_nux_step = function(res) {
                 [{'name': 'hm_ajax_hook', 'value': 'ajax_imap_accept_special_folders'},
                 {'name': 'imap_server_id', value: res.nux_server_id},
                 {'name': 'imap_service_name', value: res.nux_service_name}],
-                function(res) {
-                    window.location.href = "?page=servers";
+                function() {
+                    Hm_Utils.redirect();
```
kroky, line 56: Don't we want to keep the redirect to servers page? Otherwise, it simply reloads the nux page which doesn't make sense - user should be able to see the added server on servers page - I think that's the original idea. Do I miss something?

Issue #605, opened 2022-08-25
Maybe we should have 2 forwarding actions, like Nextcloud Mail?


marclaporte:
<img width="337" alt="2022-08-24_091525" src="https://user-images.githubusercontent.com/1004261/186735524-a56ce970-03b5-4d4d-b149-7755b801f1a3.png">

One is forward the whole thing as one attachment, without any modifications. The other is forward by converting to text that I can modify before I send. I usually the later, but I know Jason prefers the former.

Related:
https://github.com/jasonmunro/cypht/pull/558
https://github.com/jasonmunro/cypht/issues/557

dumblob:
This has been discussed before somewhere. I think the problem was which email part to choose (emails nowadays are heavily multipart and heavily nested) and conversions between mime types (this second issue is though solvable by giving the user two choices: 1. *plain copy & paste /i.e. keep the mime type and compose in that mime type/* 2. *smart heuristics*).

dumblob:
Related:
https://github.com/jasonmunro/cypht/issues/12
https://github.com/jasonmunro/cypht/issues/382

amaninyumu1:
https://github.com/cypht-org/cypht/pull/1080
![Forward](https://github.com/cypht-org/cypht/assets/51401043/a2f07eb1-445e-4031-8a30-91d2c266294a)

marclaporte:
Work in progress: https://github.com/cypht-org/cypht/pull/1080

Issue #257, opened 2018-01-24
Proper Draft support


jasonmunro:
Currently we only manage drafts in memory. If you logout they are gone. Add support for a real IMAP draft folder to be used to save drafts while in progress.

marclaporte:
Drafts are working and we are incrementally improving them
https://github.com/jasonmunro/cypht/pull/581
https://github.com/jasonmunro/cypht/pull/579
https://github.com/jasonmunro/cypht/pull/464

Issue #166, opened 2017-02-15
Question: How does auto-save currently works?


dumblob:
There was some planned work on auto save of composed messages, but I could not extract how does it work currently. Does it support only Gmail? Or is it implemented for any account? Does it use Drafts or Templates folders? How frequently is the email saved (I would prefer every 2 seconds)? How does it cope with encryption?

Thank you.

jasonmunro:
Drafts are saved in the session every 30 seconds (or when you use the "save" button). When the session expires (logout, browser close) drafts are destroyed. Session data is encrypted with a unique key generated at login and stored in a cookie. Once a draft has been saved, you can access it after revisting the compose page with the document icon in the upper right of the page. Drafts are not tied directly to a particular mail provider. I have recently built support in the imap_folders module to set a draft folder, but it is not used by the draft system yet.

dumblob:
Thank you @jasonmunro for explanation. For my needs, I would like to change the interval of autosaving to a much shorter value (e.g. 2 seconds). Autosave should trigger only if any change to the draft appeared (to save bandwidth). Could this autosave interval be configurable (either system-wide or even user-specific)? Or is it too low level and I should maintain my own fork?

jasonmunro:
I'm not too keen on exposing this as a user option, I'm considering a site level ini file setting. In the meantime however, I don't want unneeded forking :) So, I'm leveraging the module system and the "site" module to make this possible without having to fork:

https://github.com/jasonmunro/cypht/blob/master/modules/site/site.js

All you have to do is enable that module, edit the site.js file above to your liking, and run your config gen script and you should be good to go. The point of the site module set is to override default Cypht behavior without making code changes, so this seems like a good place to put this for now.

I don't think the draft saving code is smart enough to not fire on each interval when nothing has changed in the draft, I'm looking into that now.

jasonmunro:
Draft logic updated in https://github.com/jasonmunro/cypht/commit/012cc310a115140619d61f44127054fa1a107042 to only re-save if the draft has changed since the last save.

dumblob:
Thanks a lot @jasonmunro! I'm closing this as solved.

PR #1080, opened 2024-06-13
[ENH] add fonctionality forward as message attachment

amaninyumu1:
[Issue](https://github.com/cypht-org/cypht/issues/605)

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1415,6 +1415,21 @@ function snooze_dropdown($output, $unsnooze = false) {
 /**
  * @subpackage imap/functions
  */
+if (!hm_exists('forward_dropdown')) {
+    function forward_dropdown($output,$reply_args) {
+        $txt = '<div class="dropdown d-inline-block">
+                    <button type="button" class="btn btn-outline-success btn-sm dropdown-toggle" id="dropdownMenuForward" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="true">'.$output->trans('Forward').'</button>
+                    <ul class="dropdown-menu" aria-labelledby="dropdownMenuForward">';
+        $txt .= '<li><a href="?page=compose&amp;forward_as_attachment=1'.$reply_args.'" class="forward_link hlink dropdown-item d-flex justify-content-between gap-5" ><span>'.$output->trans('Forward as message attachment').'</a></li>';
+        $txt .= '<li><a href="?page=compose&amp;forward=1'.$reply_args.'" class="forward_link hlink dropdown-item d-flex justify-content-between gap-5"><span>'.$output->trans('Edit as new message').'</a></li>';
+        $txt .= '</ul></div>';
+        return $txt;
+    }
+}
+
+/**
+ * @subpackage imap/functions
+ */
 if (!hm_exists('parse_sieve_config_host')) {
 function parse_sieve_config_host($host) {
     $url = parse_url($host);
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -348,7 +348,7 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                 else {
                     $txt .= ' | <a class="reply_all_link hlink disabled_link">'.$this->trans('Reply-all').'</a>';
                 }
-                $txt .= ' | <a class="forward_link hlink" href="?page=compose&amp;forward=1'.$reply_args.'">'.$this->trans('Forward').'</a>';
+                $txt .= ' | ' . forward_dropdown($this, $reply_args);
             }
             if ($msg_part === '0') {
                 $txt .= ' | <a class="normal_link hlink msg_part_link normal_link" data-message-part="" href="#">'.$this->trans('normal').'</a>';
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -117,13 +117,59 @@ class Hm_Handler_load_smtp_is_imap_draft extends Hm_Handler_Module {
 /**
  * @subpackage smtp/handler
  */
-class Hm_Handler_load_smtp_is_imap_forward extends Hm_Handler_Module
+class Hm_Handler_load_smtp_is_imap_forward_as_attachment extends Hm_Handler_Module
 {
     public function process() {
         if (!$this->module_is_supported('imap')) {
             return;
         }
 
+        if (array_key_exists('forward_as_attachment', $this->request->get)) {
+            $path = explode('_', $this->request->get['list_path']);
+            $imap = Hm_IMAP_List::connect($path[1]);
+            if ($imap && $imap->select_mailbox(hex2bin($path[2]))) {
+                $msg_struct = $imap->get_message_structure($this->request->get['uid']);
+                $msg_header = $imap->get_message_headers($this->request->get['uid']);
+                if (!array_key_exists('From', $msg_header) || count($msg_header) == 0) {
+                    return;
+                }
+                $content = $imap->get_message_content($this->request->get['uid'], 0, false, false);
+
+                # Attachment Download
+                $attached_files = [];
+                $this->session->set('uploaded_files', array());
+
+                $file_dir = $this->config->get('attachment_dir') . DIRECTORY_SEPARATOR . md5($this->session->get('username', false)) . DIRECTORY_SEPARATOR;
+                if (!is_dir($file_dir)) {
+                    mkdir($file_dir);
+                }
+                $name = $msg_header['subject'] . '.eml';
+                $file_path = $file_dir . $name;
+                $attached_files[$this->request->get['uid']][] = array(
+                    'name' => $name,
+                    'type' => 'message/rfc822',
+                    'size' => strlen($content),
+                    'tmp_name' => $file_path,
+                    'filename' => $file_path,
+                    'basename' => $msg_header['subject']
+                );
+                file_put_contents($file_path, $content);
+                $this->session->set('uploaded_files', $attached_files);
+                $this->out('as_attr', true);
+            }
+        }
+    }
+}
+
+/**
+ * @subpackage smtp/handler
+ */
+class Hm_Handler_load_smtp_is_imap_forward extends Hm_Handler_Module
+{
+    public function process() {
+        if (!$this->module_is_supported('imap')) {
+            return;
+        }
         if (array_key_exists('forward', $this->request->get)) {
             $path = explode('_', $this->request->get['list_path']);
             $imap = Hm_IMAP_List::connect($path[1]);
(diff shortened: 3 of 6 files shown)


modules/imap/functions.php
```diff
@@ -1412,6 +1412,56 @@ function snooze_dropdown($output, $unsnooze = false) {
     return $txt;
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('get_forward_content')) {
+    function get_forward_content($format, $only_label = false) {
+        if ($format == 'Forward as message attachement') {
+            $label = 'Forward as message attachement';
```
marclaporte, line 1421: attachement -> attachment

modules/imap/functions.php
```diff
@@ -1412,6 +1412,56 @@ function snooze_dropdown($output, $unsnooze = false) {
...
+        }
+    }}
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('forward_content')) {
+    function forward_content() {
+        $values = array(
+            'Forward as message attachement',
```
marclaporte, line 1436: attachement -> attachment

modules/imap/functions.php
amaninyumu1: I corrected

modules/imap/functions.php
```diff
@@ -1412,6 +1412,56 @@ function snooze_dropdown($output, $unsnooze = false) {
     return $txt;
 }}
 
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('get_forward_content')) {
```
josaphatim, line 1418: Can you try to not use functions for this please?

modules/imap/functions.php
```diff
@@ -1412,6 +1412,56 @@ function snooze_dropdown($output, $unsnooze = false) {
...
+        }
+        if ($only_label) {
+            return [$label];
+        }
+    }}
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('forward_content')) {
```
josaphatim, line 1433: Same think here. No need to create a function.

modules/imap/functions.php
```diff
@@ -1412,6 +1412,56 @@ function snooze_dropdown($output, $unsnooze = false) {
...
+ * @subpackage imap/functions
+ */
+if (!hm_exists('forward_dropdown')) {
+    function forward_dropdown($output,$reply_args) {
+        $values = forward_content();
+    
+        $txt = '<div class="dropdown d-inline-block">
+                    <button type="button" class="btn btn-outline-success btn-sm dropdown-toggle" id="dropdownMenuForward" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="true">'.$output->trans('Forward').'</button>
+                    <ul class="dropdown-menu" aria-labelledby="dropdownMenuForward">';
+        foreach ($values as $format) {
```
josaphatim, line 1452: no need to look just hardcode them as they are 2 values

modules/imap/output_modules.php
```diff
@@ -171,6 +171,11 @@ protected function output() {
                 $this->html_safe($this->get('msg_list_path')),
                 $this->html_safe($this->get('msg_text_uid'))
             );
+            /* $reply_args_forward_attachment = sprintf('&amp;list_path=%s&amp;uid=%d',
```
josaphatim, line 174: Please delete this comment if not needed

modules/smtp/modules.php
```diff
@@ -117,13 +117,66 @@ public function process() {
...
+/**
+ * @subpackage smtp/handler
+ */
+class Hm_Handler_load_smtp_is_imap_forward extends Hm_Handler_Module
+{
+    public function process() {
+        if (!$this->module_is_supported('imap')) {
+            return;
+        }
+        if ($this->get('as_attr', false)) {
```
josaphatim, line 176: No need to pass this as an output and get it back. You can just use the array_key_exists testing forward_as_attachment in post.

modules/imap/functions.php
amaninyumu1: I just removed this function

amaninyumu1: Please review

modules/imap/functions.php
amaninyumu1: I just removed this function please review

modules/imap/output_modules.php
amaninyumu1: I modified this function and made fixes

modules/smtp/modules.php
amaninyumu1: I deleted this line

modules/imap/functions.php
amaninyumu1: Please Review

josaphatim:
Hello @kroky. This MR looks good to me. Can you check please ?

modules/smtp/setup.php
```diff
@@ -18,6 +18,7 @@
 add_output('compose', 'compose_form_end', true, 'smtp', 'compose_form_content', 'after');
 add_output('compose', 'compose_form_attach', true, 'smtp', 'compose_form_end', 'after');
 add_handler('compose', 'load_smtp_is_imap_forward', true, 'smtp', 'load_user_data', 'after');
+add_handler('compose', 'load_smtp_is_imap_forward_as_attachment', true, 'smtp', 'load_user_data', 'after');
```
kroky, line 21: I'd suggest you do a 'before' insert before the is_imap_forward method or just switch these 2 places - execute forward_as_attachment after load_user_data and then forward after forward_as_attachment. We need to make sure forward_as_attachment is executed before the main forward method.

modules/smtp/modules.php
```diff
@@ -117,13 +117,62 @@ public function process() {
...
         if (!$this->module_is_supported('imap')) {
             return;
         }
 
+        if (array_key_exists('forward_as_attachment', $this->request->get)) {
+            $path = explode('_', $this->request->get['list_path']);
+            $imap = Hm_IMAP_List::connect($path[1]);
+            if ($imap && $imap->select_mailbox(hex2bin($path[2]))) {
+                $msg_struct = $imap->get_message_structure($this->request->get['uid']);
+                list($part, $msg_text) = $imap->get_first_message_part($this->request->get['uid'], 'text', 'plain', $msg_struct);
```
kroky, line 132: Why do you get the first message part and text if you don't use them?

modules/smtp/modules.php
```diff
@@ -117,13 +117,62 @@ public function process() {
...
+                $msg_header = $imap->get_message_headers($this->request->get['uid']);
+                if (!array_key_exists('From', $msg_header) || count($msg_header) == 0) {
+                    return;
+                }
+                $content = $imap->get_message_content($this->request->get['uid'], 0, false, false);
```
kroky, line 137: We need to make sure we get the full email as stored by the remote imap server including all headers and message parts.

kroky: I think it will also make sense to forward the plain text / html parts as an eml and forward the attached files to the original email as separate attachments. Not sure - @marclaporte what do you think?

modules/smtp/modules.php
```diff
@@ -117,13 +117,62 @@ public function process() {
...
+                $attached_files[$this->request->get['uid']][] = array(
+                    'name' => $name,
+                    'type' => 'message/rfc822',
+                    'size' => strlen($content),
+                    'tmp_name' => $file_path,
+                    'filename' => $file_path,
+                    'basename' => $msg_header['subject']
+                );
+                
+                $content = Hm_Crypt::ciphertext($content, Hm_Request_Key::generate());
```
kroky, line 158: Shouldn't we use the current generated request key instead of generating a new one?

modules/smtp/modules.php
```diff
@@ -117,13 +117,62 @@ public function process() {
...
+                list($part, $msg_text) = $imap->get_first_message_part($this->request->get['uid'], 'text', 'plain', $msg_struct);
+                $msg_header = $imap->get_message_headers($this->request->get['uid']);
+                if (!array_key_exists('From', $msg_header) || count($msg_header) == 0) {
+                    return;
+                }
+                $content = $imap->get_message_content($this->request->get['uid'], 0, false, false);
+
+                # Attachment Download
+                $attached_files = [];
+                $this->session->set('uploaded_files', array());
```
kroky, line 141: What if there is a draft message in compose page with files attached? Won't we remove those files in that case?

modules/smtp/modules.php
marclaporte: I just checked and in Kolab (Roundcube), forwarding bundles everything in one .eml. I think this makes sense. If folks want to pick and choose what to forward, they can use "Edit as new message"

modules/smtp/modules.php
amaninyumu1: I don't understand the question please

Baraka24: @kroky, @amaninyumu1 I hope If a draft message is present, we can save its attachments temporarily.

amaninyumu1: transferring a message as an attachment behaves like a new message if you want to save it as a draft you have to click on the save button

marclaporte: > if you want to save it as a draft you have to click on the save button

Please confirm if there is an auto-save which does on its own after a while. If not, please create an issue. We auto-save when drafts were saved in the session (https://github.com/cypht-org/cypht/issues/166) but I don't if this was kept when we move to saving drafts in IMAP: https://github.com/cypht-org/cypht/issues/257

amaninyumu1: the logistics is that when you transfer a message as an attachment, it is like a new message, so the automatic saving can no longer be done after a few minutes. you have to click on save if you want to save this message as a draft.

Baraka24: Are you sure ? For a new message there is an auto-save which does on its own after a while. I have tested this.... See image below:
autosavedraftnewmessage

amaninyumu1: Maybe it escaped me. I will try again to confirm. Thank

kroky: @amaninyumu1 my original question was this - you open a compose form, attach some files (they get stored as uploaded_files in the session), you save this form as a draft and go somewhere else, you then click to forward a message as attachment - this destroys the files you uploaded in a previous draft which seems like a regression in functionality. Maybe at least warn the user that they have unsent email that will be lost?

amaninyumu1: @kroky the attachment files are not saved in the session but rather in the attachment folder dir. So it can't have any impact on other messages. I just tested this confirmation, there was no problem

amaninyumu1: @Baraka24 Thank you very much for your curiosity, autosave works for the functionality that I added. I just tested

kroky: @amaninyumu1 I don't mean the actual file contents. uploaded_files session key is still used extensively in smtp/modules.php to store references and metadata of attached files from the compose form - drafts, imap messages coming from drafts, etc. If you carelessly clean that session array, all those references are gone even if the file system has the actual files. Honestly, I don't think the attachment file session handling is good enough in Cypht - it is prone to errors with this hard-cleanup even in existing code. One obvious example would be someone opening up 2 compose form tabs and attaching files to both forms. Since this is not part of the original issue, I am going to merge this but we need to improve attachment handling internally.

modules/smtp/modules.php
kroky: OK, let's keep the whole eml including attachments type of forward for now.

PR #1085, opened 2024-06-14
Fix undeclared properties in hm-feed.php

josaphatim:
Fixes `PHP Fatal error:  Uncaught Error: Call to undefined method Hm_Contact::dump() in /path/to/cypht/lib/repository.php:59` and creation of dynamic proporties

### DIFF
--- a/modules/feeds/hm-feed.php
+++ b/modules/feeds/hm-feed.php
@@ -75,6 +75,8 @@ class Hm_Feed {
     var $init_cache;
     var $cache_limit;
     var $sort;
+    var $status_code;
+    var $feed_type;
 
     /**
      * Setup defaults


lib/repository.php
```diff
@@ -55,7 +55,12 @@ public static function add($entity, $save = true) {
 
     public static function edit($id, $entity) {
         if (array_key_exists($id, self::$entities)) {
-            self::$entities[$id] = array_merge(self::$entities[$id], $entity);
+            if (is_array(self::$entities[$id])) {
+                $data =  array_merge(self::$entities[$id], $entity);;
```
kroky, line 59: Please fix the double semicolon at the end.

lib/repository.php
```diff
@@ -55,7 +55,12 @@ public static function add($entity, $save = true) {
...
+            } else {
+                $data = $entity;
```
kroky, line 61: When that's an object, don't we still need a merge? Otherwise, it is an overwrite.

josaphatim: when it is an object it already comes with the necessary modifications unlike an array where only certain properties can be modified

lib/repository.php
josaphatim: fixed

lib/repository.php
kroky: Ok but this makes the whole method quite confusing - imagine someone trying to use it - how dothey know if this method actually replaces the entity or updates the passed entity params only? I still suggest you unify - either do a full replace and require calling code to pass full entities or merge. Having a different business logic based on type of incoming parameters is a bad programming practice.

josaphatim: The same thing has been done here and merged

https://github.com/cypht-org/cypht/pull/1132/files#diff-e8ea654d76e94bfce8824e84e96fec6cf4d3497b6e3d446ab8e1f81d8eb983ceR58

marclaporte:
@josaphatim This branch has conflicts that must be resolved

josaphatim:
> @josaphatim This branch has conflicts that must be resolved

Fixed, corrected and ready to be merged

Issue #576, opened 2022-07-15
Delay-send / "Send later" / Scheduled sending


nekohayo:
Issue #572 talks about snoozing (i.e. hiding a received email from the inbox for a certain amount of time), but what I'd like to suggest here is a different feature. The usecase for me is this:

*  Business or events-related outreach emails typically have a context/timing-sensitive aspect, so I'd like to be able to schedule them (especially considering the timezone of the person I'm contacting)

* When dealing with public-facing businesses, it is often counterproductive to send emails on certain days or hours (ex: Friday afternoons, when you know everything is likely to be forgotten over the week-end), so you want to schedule your email for a better time during business hours (ex: Tuesday morning).

* Human-centric staff managers and directors like me know that it is better to schedule mails to staffers to their business hours, avoiding emails that would arrive outside work hours and cause undue stress or cognitive load. Basically, mindfulness for staffers' "[right to disconnect](https://en.wikipedia.org/wiki/Right_to_disconnect)".

The corresponding issue in my local mail client (where some technical approaches are mentioned) is https://gitlab.gnome.org/GNOME/evolution/-/issues/411#note_874473

marclaporte:
See also: https://github.com/jasonmunro/cypht/issues/641

dumblob:
Yeah, I outlined an envisioned potential implementation in #641 . Feel free to take a look.

dumblob:
Regarding the proposal of the "local approach" in https://gitlab.gnome.org/GNOME/evolution/-/issues/411#note_874473 I do not think it is viable for Cypht. It simply has to be "fire and forget" - i.e. closing the web site (incl. suspending the web browser tab) must not hinder this functionality in any way.

nekohayo:
A bit like the "snooze" counterpart, I've requested clarification on JMAP side at https://github.com/jmapio/jmap/issues/373 because it is unclear to me (perhaps because I'm not familiar with such things as protocols and specs and whatnot)

marclaporte:
Is this support by IMAP/SMTP?

If not (how) can this be done without a cron job?

marclaporte:
Some ideas from Roundcube: https://github.com/roundcube/roundcubemail/issues/8110

marclaporte:
Much to my surprise, this issue is number 3 for most reactions:
https://github.com/cypht-org/cypht/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions

marclaporte:
[...]

Issue #572, opened 2022-06-19
"Snooze" (reply-later) feature for any message: Hide this email for x hours/days/weeks/months


marclaporte:
Like many (most?) people, my email inboxes and other email folders often contain todos and reminders. When I am completely finished with an email, I will delete it (If I am sure I will never need it again) or I archive it (out of my view but still findable): https://github.com/jasonmunro/cypht-website/issues/24

Cypht nicely permits to [combine views](https://github.com/jasonmunro/cypht-website/issues/25), so this accentuates the challenge of quite a few emails with diverging levels of urgency.

While [inbox zero](https://search.brave.com/search?q=inbox+zero) is elusive, I regularly review emails from most recent to oldest. This "[last in, first out aka LIFO method](https://www.themuse.com/advice/4-strategies-for-keeping-your-inbox-empty-which-one-is-right-for-you)" is debatable logic as older emails can be more important, and they have been waiting longer. However, newer emails sometimes make older emails no longer important, so it makes sense to start with latest information. And [threading will help](https://github.com/jasonmunro/cypht/issues/10). And I refuse to [declare email bankruptcy](https://en.wikipedia.org/wiki/Email_bankruptcy).

[Better filters being a way to deal with information overload](https://www.youtube.com/watch?v=LabqeJEOQyI), let's not analyze the same email again and again (without anything immediately actionable). Examples:
- I am in CC and someone else is supposed to answer. I want to make sure they don't forget because this is important.
- Someone replied to me with a quick confirmation of receipt, and it will take a few days/weeks/months before I can expect a follow-up on this topic
- An important but not urgent email, and I want to make sure it's not lost
- A confirmation of an appointment/online event far away in the future. Ex.: with information on how to access the webinar
- etc.

In cases like this, I would like to snooze the email with choices like 1 day, 1 week, 1 month, pick a date. The snoozed emails should all be findable in a "snoozed" folder (or similar name), with a column for the snoozed-until date. Here, you can un-snooze them (it becomes a normal email), or change the snooze date. At the end of the snooze period, the email should appear in a dedicated section above all other emails. Then, you deal with them, or snooze them again.

It should be possible to "snooze" emails in any folder (sent, draft, etc.). If you send an important email, you can then add yourself a reminder to check up on it in one month (for example).

Why not use the flag tag? The flag tag is indeed useful, but then, you can end up evaluating an email again and again. We really need to hide the email until a specific date for that email.

Alternate names for this feature:
- Snooze
- Defer
[...]

PR #1091, opened 2024-06-15
feat(backend): add scheduled send functionality

amaninyumu1:
[issue](https://github.com/cypht-org/cypht/issues/572)
[issue](https://github.com/cypht-org/cypht/issues/576)

### DIFF
--- a/language/en.php
+++ b/language/en.php
@@ -636,6 +636,7 @@ return array(
     'You must provide a name for your script' => false,
     'Empty script' => false,
     'Please create a profile for saving sent messages option' => false,
+    'You have %d scheduled messages that won\'t be executed if you quit' => false,
     'Attachment storage unavailable, please contact your site administrator' => false,
     'Your subject is empty!' => false,
     'Your body is empty!' => false,
--- a/language/fa.php
+++ b/language/fa.php
@@ -670,6 +670,7 @@ return array(
     'You must provide a name for your script' => false,
     'Empty script' => false,
     'Please create a profile for saving sent messages option' => false,
+    'You have %d scheduled messages that won\'t be executed if you quit' => false,
     'Attachment storage unavailable, please contact your site administrator' => false,
     'Your subject is empty!' => false,
     'Your body is empty!' => false,
--- a/lib/servers.php
+++ b/lib/servers.php
@@ -273,4 +273,14 @@ trait Hm_Server_List {
         }
         return false;
     }
+
+    public static function getForMailbox($id) {
+        $server = self::get($id, true);
+        if ($server) {
+            $server['password'] = $server['pass'];
+            $server['username'] = $server['user'];
+            return $server;
+        }
+        return false;
+    }
 }
--- a/lib/servers.php
+++ b/lib/servers.php
@@ -274,13 +274,26 @@ trait Hm_Server_List {
         return false;
     }
 
+    private static function appendPpasswordAndUsername(array $server) {
+        $server['password'] = $server['pass'];
+        $server['username'] = $server['user'];
+        return $server;
+    }
+
     public static function getForMailbox($id) {
         $server = self::get($id, true);
         if ($server) {
-            $server['password'] = $server['pass'];
-            $server['username'] = $server['user'];
-            return $server;
+            return self::appendPpasswordAndUsername($server);
         }
         return false;
+    }
+
+    public static function dumpForMailbox($id = false) {
+        $list = self::dump($id, true);
+        foreach ($list as $index => $server) {
+            $server = self::appendPpasswordAndUsername($server);
+            $list[$index] = $server;
+        }
+        return $list;
     }
 }
--- a/lib/servers.php
+++ b/lib/servers.php
@@ -274,7 +274,7 @@ trait Hm_Server_List {
         return false;
     }
 
-    private static function appendPpasswordAndUsername(array $server) {
+    private static function appendPasswordAndUsername(array $server) {
         $server['password'] = $server['pass'];
         $server['username'] = $server['user'];
         return $server;
@@ -283,7 +283,7 @@ trait Hm_Server_List {
     public static function getForMailbox($id) {
         $server = self::get($id, true);
         if ($server) {
-            return self::appendPpasswordAndUsername($server);
+            return self::appendPasswordAndUsername($server);
         }
         return false;
     }
@@ -291,7 +291,7 @@ trait Hm_Server_List {
     public static function dumpForMailbox($id = false) {
         $list = self::dump($id, true);
         foreach ($list as $index => $server) {
-            $server = self::appendPpasswordAndUsername($server);
+            $server = self::appendPasswordAndUsername($server);
             $list[$index] = $server;
         }
         return $list;
--- a/lib/servers.php
+++ b/lib/servers.php
@@ -290,6 +290,9 @@ trait Hm_Server_List {
 
     public static function dumpForMailbox($id = false) {
         $list = self::dump($id, true);
+        if ($id !== false) {
+            return self::appendPasswordAndUsername($list);
+        }
         foreach ($list as $index => $server) {
             $server = self::appendPasswordAndUsername($server);
             $list[$index] = $server;
--- a/modules/core/functions.php
+++ b/modules/core/functions.php
@@ -637,3 +637,99 @@ function privacy_setting_callback($val, $key, $mod) {
     }
     return $val;
 }
+
+if (!hm_exists('get_scheduled_date')) {
+function get_scheduled_date($format, $only_label = false) {
+    if ($format == 'later_in_day') {
+        $date_string = 'today 18:00';
+        $label = 'Later in the day';
+    } elseif ($format == 'tomorrow') {
+        $date_string = '+1 day 08:00';
+        $label = 'Tomorrow';
+    } elseif ($format == 'next_weekend') {
+        $date_string = 'next Saturday 08:00';
+        $label = 'Next weekend';
+    } elseif ($format == 'next_week') {
+        $date_string = 'next week 08:00';
+        $label = 'Next week';
+    } elseif ($format == 'next_month') {
+        $date_string = 'next month 08:00';
+        $label = 'Next month';
+    } else {
+        $date_string = $format;
+        $label = 'Certain date';
+    }
+    $time = strtotime($date_string);
+    if ($only_label) {
+        return [$label, date('D, H:i', $time)];
+    }
+    return date('D, d M Y H:i', $time);
+}}
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('nexter_formats')) {
+function nexter_formats() {
+    $values = array(
+        'tomorrow',
+        'next_weekend',
+        'next_week',
+        'next_month'
+    );
+    if (date('H') <= 16) {
+        array_push($values, 'later_in_day');
+    }
+    return $values;
+}}
+
+if (!hm_exists('schedule_dropdown')) {
+function schedule_dropdown($output, $send_now = false) {
+    $values = nexter_formats();
+
+    $txt = '';
+    if ($send_now) {
+        $txt .= '<div class="dropdown d-inline-block">
+                <button type="button" class="btn btn-light btn-sm dropdown-toggle" id="dropdownMenuNexterDate" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="true">'.$output->trans('Schedule sending').'</button>';
+    }
+    $txt .= '<ul class="dropdown-menu nexter_dropdown schedule_dropdown" aria-labelledby="dropdownMenuNexterDate">';
+    foreach ($values as $format) {
+        $labels = get_scheduled_date($format, true);
+        $txt .= '<li><a href="#" class="nexter_date_helper dropdown-item d-flex justify-content-between gap-5" data-value="'.$format.'"><span>'.$output->trans($labels[0]).'</span> <span class="text-end">'.$labels[1].'</span></a></li>';
+    }
+    $txt .= '<li><hr class="dropdown-divider"></li>';
+    $txt .= '<li><label for="nexter_input_date" class="nexter_date_picker dropdown-item cursor-pointer">'.$output->trans('Pick a date').'</label>';
+    $txt .= '<input id="nexter_input_date" type="datetime-local" min="'.date('Y-m-d\Th:m').'" class="nexter_input_date" style="visibility: hidden; position: absolute; height: 0;">';
+    $txt .= '<input class="nexter_input" style="display:none;"></li>';
+    if ($send_now) {
+        $txt .= '<li><hr class="dropdown-divider"></li>';
+        $txt .= '<li><a href="#" data-value="now" class="nexter_date_helper dropdown-item"">'.$output->trans('Send now').'</a></li>';
+    }
+    $txt .= '</ul>';
+    if ($send_now) {
+        $txt .= '</div>';
+    }
+
+    return $txt;
+}}
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_nexter_header')) {
+    function parse_nexter_header($header, $name)
+    {
+        $header = str_replace("$name: ", '', $header);
+        $result = [];
+        foreach (explode(';', $header) as $kv)
+        {
+            $kv = trim($kv);
+            $spacePos = strpos($kv, ' ');
+            if ($spacePos > 0) {
+                $result[rtrim(substr($kv, 0, $spacePos), ':')] = trim(substr($kv, $spacePos+1));
+            } else {
+                $result[$kv] = true;
+            }
+        }
+        return $result;
+    }}
--- a/modules/core/functions.php
+++ b/modules/core/functions.php
@@ -639,32 +639,44 @@ function privacy_setting_callback($val, $key, $mod) {
 }
 
 if (!hm_exists('get_scheduled_date')) {
-function get_scheduled_date($format, $only_label = false) {
-    if ($format == 'later_in_day') {
-        $date_string = 'today 18:00';
-        $label = 'Later in the day';
-    } elseif ($format == 'tomorrow') {
-        $date_string = '+1 day 08:00';
-        $label = 'Tomorrow';
-    } elseif ($format == 'next_weekend') {
-        $date_string = 'next Saturday 08:00';
-        $label = 'Next weekend';
-    } elseif ($format == 'next_week') {
-        $date_string = 'next week 08:00';
-        $label = 'Next week';
-    } elseif ($format == 'next_month') {
-        $date_string = 'next month 08:00';
-        $label = 'Next month';
-    } else {
-        $date_string = $format;
-        $label = 'Certain date';
+    function get_scheduled_date($format, $only_label = false) {
+        switch ($format) {
+            case 'later_in_day':
+                $date_string = 'today 18:00';
+                $label = 'Later in the day';
+                break;
+            case 'tomorrow':
+                $date_string = '+1 day 08:00';
+                $label = 'Tomorrow';
+                break;
+            case 'next_weekend':
+                $date_string = 'next Saturday 08:00';
+                $label = 'Next weekend';
+                break;
+            case 'next_week':
+                $date_string = 'next week 08:00';
+                $label = 'Next week';
+                break;
+            case 'next_month':
+                $date_string = 'next month 08:00';
+                $label = 'Next month';
+                break;
+            default:
+                $date_string = $format;
+                $label = 'Certain date';
+                break;
+        }
+
+        $time = strtotime($date_string);
+
+        if ($only_label) {
+            return [$label, date('D, H:i', $time)];
+        }
+
+        return date('D, d M Y H:i', $time);
     }
-    $time = strtotime($date_string);
-    if ($only_label) {
-        return [$label, date('D, H:i', $time)];
-    }
-    return date('D, d M Y H:i', $time);
-}}
+}
+
 
 /**
  * @subpackage imap/functions
@@ -721,15 +733,16 @@ if (!hm_exists('parse_nexter_header')) {
     {
         $header = str_replace("$name: ", '', $header);
         $result = [];
-        foreach (explode(';', $header) as $kv)
+        foreach (explode(';', $header) as $keyValue)
         {
-            $kv = trim($kv);
-            $spacePos = strpos($kv, ' ');
+            $keyValue = trim($keyValue);
+            $spacePos = strpos($keyValue, ' ');
             if ($spacePos > 0) {
-                $result[rtrim(substr($kv, 0, $spacePos), ':')] = trim(substr($kv, $spacePos+1));
+                $result[rtrim(substr($keyValue, 0, $spacePos), ':')] = trim(substr($keyValue, $spacePos+1));
             } else {
-                $result[$kv] = true;
+                $result[$keyValue] = true;
             }
         }
         return $result;
-    }}
+    }
+}
--- a/modules/core/functions.php
+++ b/modules/core/functions.php
@@ -673,7 +673,7 @@ if (!hm_exists('get_scheduled_date')) {
             return [$label, date('D, H:i', $time)];
         }
 
-        return date('D, d M Y H:i', $time);
+        return date('D, d M Y H:i T', $time);
     }
 }
 
--- a/modules/core/functions.php
+++ b/modules/core/functions.php
@@ -702,7 +702,7 @@ function schedule_dropdown($output, $send_now = false) {
     $txt = '';
     if ($send_now) {
         $txt .= '<div class="dropdown d-inline-block">
-                <button type="button" class="btn btn-light btn-sm dropdown-toggle" id="dropdownMenuNexterDate" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="true">'.$output->trans('Schedule sending').'</button>';
+                <button type="button" class="btn btn-light btn-sm dropdown-toggle" id="dropdownMenuNexterDate" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="true">'.$output->trans('Reschedule').'</button>';
     }
     $txt .= '<ul class="dropdown-menu nexter_dropdown schedule_dropdown" aria-labelledby="dropdownMenuNexterDate">';
     foreach ($values as $format) {
@@ -728,8 +728,8 @@ function schedule_dropdown($output, $send_now = false) {
 /**
  * @subpackage imap/functions
  */
-if (!hm_exists('parse_nexter_header')) {
-    function parse_nexter_header($header, $name)
+if (!hm_exists('parse_delayed_header')) {
+    function parse_delayed_header($header, $name)
     {
         $header = str_replace("$name: ", '', $header);
         $result = [];
--- a/modules/core/message_list_functions.php
+++ b/modules/core/message_list_functions.php
@@ -336,11 +336,11 @@ function subject_callback($vals, $style, $output_mod) {
  */
 if (!hm_exists('date_callback')) {
 function date_callback($vals, $style, $output_mod) {
-    $nexter_class = isset($vals[2]) && $vals[2]? ' nexter_date': '';
+    $delayed_class = isset($vals[2]) && $vals[2]? ' delayed_date': '';
     if ($style == 'news') {
-        return sprintf('<div class="msg_date%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></div>', $nexter_class, $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
+        return sprintf('<div class="msg_date%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></div>', $delayed_class, $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
     }
-    return sprintf('<td class="msg_date%s" title="%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></td>', $nexter_class, $output_mod->html_safe(date('r', $vals[1])), $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
+    return sprintf('<td class="msg_date%s" title="%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></td>', $delayed_class, $output_mod->html_safe(date('r', $vals[1])), $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
 }}
 
 function dates_holders_callback($vals) {
--- a/modules/core/message_list_functions.php
+++ b/modules/core/message_list_functions.php
@@ -336,11 +336,11 @@ function subject_callback($vals, $style, $output_mod) {
  */
 if (!hm_exists('date_callback')) {
 function date_callback($vals, $style, $output_mod) {
-    $snooze_class = isset($vals[2]) && $vals[2]? ' snoozed_date': '';
+    $nexter_class = isset($vals[2]) && $vals[2]? ' nexter_date': '';
     if ($style == 'news') {
-        return sprintf('<div class="msg_date%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></div>', $snooze_class, $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
+        return sprintf('<div class="msg_date%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></div>', $nexter_class, $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
     }
-    return sprintf('<td class="msg_date%s" title="%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></td>', $snooze_class, $output_mod->html_safe(date('r', $vals[1])), $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
+    return sprintf('<td class="msg_date%s" title="%s">%s<input type="hidden" class="msg_timestamp" value="%s" /></td>', $nexter_class, $output_mod->html_safe(date('r', $vals[1])), $output_mod->html_safe($vals[0]), $output_mod->html_safe($vals[1]));
 }}
 
 function dates_holders_callback($vals) {
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -2425,8 +2425,9 @@ function setupActionSchedule(callback) {
         }
     });
     $(document).on('change', '.nexter_input_date', function (e) {
-        if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
-            $('.nexter_input').val($(this).val()).trigger('change');
+        const selectedDate = new Date($(this).val());
+        if ($(this).val() && new Date().getTime() < selectedDate.getTime()) {
+            $('.nexter_input').val(selectedDate.toISOString()).trigger('change');
         }
     });
     $(document).on('change', '.nexter_input', callback);
@@ -2451,8 +2452,9 @@ function setupActionSnooze(callback) {
         }
     });
     $(document).on('change', '.nexter_input_date_snooze', function (e) {
-        if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
-            $('.nexter_input_snooze').val($(this).val()).trigger('change');
+        const selectedDate = new Date($(this).val());
+        if ($(this).val() && new Date().getTime() < selectedDate.getTime()) {
+            $('.nexter_input_snooze').val(selectedDate.toISOString()).trigger('change');
         }
     });
     $(document).on('change', '.nexter_input_snooze', callback);
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -47,7 +47,7 @@ var Hm_Ajax = {
     active_reqs: 0,
     icon_loading_id: false,
 
-    get_ajax_hook_name: function (args) {
+    get_ajax_hook_name: function(args) {
         var index;
         for (index in args) {
             if (args[index]['name'] == 'hm_ajax_hook') {
@@ -57,7 +57,7 @@ var Hm_Ajax = {
         return;
     },
 
-    request: function (args, callback, extra, no_icon, batch_callback, on_failure, signal) {
+    request: function(args, callback, extra, no_icon, batch_callback, on_failure, signal) {
         var bcb = false;
         if (typeof batch_callback != 'undefined' && $.inArray(batch_callback, this.batch_callbacks) === -1) {
             bcb = batch_callback.toString();
@@ -79,27 +79,27 @@ var Hm_Ajax = {
         return ajax.make_request(args, callback, extra, name, on_failure, batch_callback, signal);
     },
 
-    show_loading_icon: function () {
+    show_loading_icon: function() {
         if (Hm_Ajax.icon_loading_id !== false) {
             return;
         }
-        var hm_loading_pos = $('.loading_icon').width() / 40;
+        var hm_loading_pos = $('.loading_icon').width()/40;
         $('.loading_icon').show();
         function move_background_image() {
             hm_loading_pos = hm_loading_pos + 50;
-            $('.loading_icon').css('background-position', hm_loading_pos + 'px 0');
+            $('.loading_icon').css('background-position', hm_loading_pos+'px 0');
             Hm_Ajax.icon_loading_id = setTimeout(move_background_image, 100);
         }
         move_background_image();
     },
 
-    stop_loading_icon: function (loading_id) {
+    stop_loading_icon : function(loading_id) {
         clearTimeout(loading_id);
         $('.loading_icon').hide();
         Hm_Ajax.icon_loading_id = false;
     },
 
-    process_callback_hooks: function (name, res) {
+    process_callback_hooks: function(name, res) {
         var hook;
         var func;
         for (var i in Hm_Ajax.callback_hooks) {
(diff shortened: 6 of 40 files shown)


marclaporte:
Will this work with JMAP as well?

kroky:
Yes, anything we add for IMAP, we should check JMAP support as well. Overall, it is built on top of IMAP, so it should work but please check if the new Scheduled folder is OK for JMAP.

Overall, the code is good but I have one concern - it only sends out while you are actively using Cypht and it seems to warn of unsent/scheduled messages each time you leave a page (onbeforeunload). That would be annoying for the end user. I believe this comment https://github.com/cypht-org/cypht/issues/576#issuecomment-2119670366 proposes using sendAt support from jmap servers or scheduled send - if it is supported, we use it, if not, we fallback to a more annoying option. At any rate, it should be configurable.

This could be fully-supported in the Cypht-Tiki integration where there could be a Tiki command run periodically in the scheduler  that checks and sends scheduled messages.

josaphatim:
@kroky I  added a commit to fix issue concerning onbeforeunload event and also add the option to change schedule time or send the message immediately. Can you check please.

marclaporte:
@amaninyumu1 If you need help with JMAP, please reach out to @Shadow243 as he set up a JMAP server for testing.

marclaporte:
@amaninyumu1 "This branch has conflicts that must be resolved"

marclaporte:
@amaninyumu1 This branch has conflicts that must be resolved

amaninyumu1:
Hello @marclaporte , @kroky I resolved the conflicts. please review

modules/smtp/modules.php
```diff
@@ -2291,3 +2452,102 @@ function recip_count_check($headers, $omod) {
         Hm_Msgs::add('ERRMessage contains more than the maximum number of recipients, proceed with caution');
     }
 }}
+
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('send_scheduled_message')) {
+function send_scheduled_message($handler, $imap, $msg, $server_id) {
```
kroky, line 2460: Can you combine this with the send_scheduled_messages handler method, so we don't repeat so much code?

modules/smtp/modules.php
```diff
@@ -939,6 +946,132 @@ protected function output() {
...
+     * Send delayed messages
+     * This should use cron
+     */
+    public function process() {
+        if (!($this->module_is_supported('imap') || $this->module_is_supported('profiles'))) {
+            return;
+        }
+        $servers = Hm_IMAP_List::dump();
+        $scheduled_msg_count = 0;
+            foreach (array_keys($servers) as $server_id) {
```
kroky, line 965: The whole indentation of this block is wrong and hard to read.

modules/smtp/modules.php
```diff
@@ -939,6 +946,132 @@ protected function output() {
...
+                                            $imap->message_action('EXPUNGE', [$msg['uid']]);
+                                        }
+                                        save_sent_msg($this, $server_id, $imap, $imap_details, $msg_content, $msg['uid'], false);
+                                        $scheduled_msg_count--;
+                                    }
+                                }
+                            }
+                        } catch (Exception $e) {
+                            Hm_Debug::add(sprintf('ERRCannot send message: %s', $msg_headers['subject']));
+                        if (send_scheduled_message($this, $imap, $msg, $server_id)) {
```
kroky, line 1023: If you retry sending in some other way, combine the above sending with the one in this method, so we don't repeat so much code.

modules/smtp/modules.php
```diff
@@ -2291,3 +2452,102 @@ function recip_count_check($headers, $omod) {
...
+            }
+        }
+    } catch (Exception $e) {
+        Hm_Debug::add(sprintf('ERRCannot send message: %s', $msg_headers['subject']));
+    }
+    return;
+}}
+
+if (!hm_exists('reschedule_message_sending')) {
+function reschedule_message_sending($handler, $imap, $msg_id, $folder, $new_date, $server_id) {
```
kroky, line 2519: Use modules.php only for handle or output modules. Move the functions to its own functions.php file to keep the existing structure in other modules (e.g. imap and core).

kroky:
There are still conflicts and I believe big part of the reason is https://github.com/cypht-org/cypht/pull/1266 - @jacob-js , can you check if some handlers in this PR need to be updated according to the router changes you did in 1266?

modules/imap/site.js
```diff
@@ -757,29 +769,64 @@ var imap_message_view_finished = function(msg_uid, detail, skip_links) {
...
+        else if (hm_list_parent() === 'drafts') {
+            Hm_Message_List.prev_next_links('formatted_drafts_data', class_name);
+        }
+        else if (hm_list_parent() === 'tag') {
+            Hm_Message_List.prev_next_links('formatted_tag_data', class_name);
+        }
+        else {
+            var key = 'imap_' + Hm_Utils.get_url_page_number() + '_' + getListPathParam();
+            Hm_Message_List.prev_next_links(key, class_name);
+        }
```
jacob-js, line 806: This rebase isn't accurate. Master has passed over these changes.

modules/imap/site.js
```diff
@@ -822,32 +869,32 @@ var get_local_message_content = function(msg_uid, path) {
     return Hm_Utils.get_from_local_storage(getMessageStorageKey(msg_uid));
 };
 
-var imap_setup_message_view_page = function(uid, details, list_path, callback) {
+var imap_setup_message_view_page = function (uid, details, list_path, callback) {
     if (!uid) {
         uid = getMessageUidParam();
     }
-    const callbackFn = (...args) => {
```
jacob-js, line 829: Could you avoid such changes in the future? it's really difficult to focus on what has actually changed.

jacob-js:
The small piece of code added won't need an update since it is not executed on page loads.

modules/imap/site.js
kroky: @amaninyumu1 I think you should be careful merging the changes from master - there are conflicts because of the upstream changes and these need to be merged correctly (keep the master ones and copy your code to the relevant places).

modules/smtp/modules.php
amaninyumu1: For the concern about code repetition. the two methods have slightly different execution contexts. one handles message rescheduling and performs specific tasks related to IMAP connection and cache fetching, while the other is more geared towards sending messages at a specific time based on message headers. planning. I have tried several ways to merge but I think that merging these two methods could complicate the logic and could make the code more difficult to maintain.

modules/smtp/modules.php
amaninyumu1: I aligned for better readability, which allows you to quickly understand the hierarchy and logic of the code

modules/core/site.js
```diff
@@ -1961,10 +1961,23 @@ var imap_smtp_edit_action = function(event) {
...
+    return format.replace(/%([sd])/g, (match, type) => {
+        let arg = args[i++];
+        switch (type) {
+            case 's': return String(arg);
+            case 'd': return Number(arg);
+            default: return match;
+        }
+    });
+}
+
```
jacob-js, line 1980: I don't think this is necessary. If including variables in a string is all you need, you can use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

modules/core/site.js
```diff
@@ -2012,6 +2025,12 @@ $(function() {
         try { navigator.registerProtocolHandler("mailto", "?page=compose&compose_to=%s", "Cypht"); } catch(e) {}
     }
 
+    if (hm_page_name() == 'home') {
+        $('.pw_update').on("click", function() { update_password($(this).data('id')); });
+    }
+    if (hm_page_name() == 'servers') {
+        $('.edit_server_connection').on('click', imap_smtp_edit_action);
```
jacob-js, line 2032: You should do the same here. Check if the statement is not yet in the serversPageHandlers and move it to there.

modules/imap/site.js
```diff
@@ -1236,6 +1236,73 @@ $(function() {
...
+    }
+    else if (hm_page_name() === 'message' && hm_list_path().substr(0, 4) === 'imap') {
+        imap_setup_message_view_page();
+    }
+    else if (hm_page_name() === 'servers') {
+        imap_setup_server_page();
+    }
+    else if (hm_page_name() === 'info') {
+        setTimeout(imap_status_update, 100);
+    }
```
jacob-js, line 1253: All of this should not be here anymore.

modules/imap/site.js
```diff
@@ -1236,6 +1236,73 @@ $(function() {
...
+
+    if (hm_page_name() === 'message_list' || hm_page_name() === 'message') {
+        imap_setup_snooze();
+        imap_setup_tags();
+        setup_nexter_date(function(e) {
```
jacob-js, line 1258: Move this function to the appropriate page handlers

modules/smtp/site.js
```diff
@@ -427,3 +436,264 @@ function smtpSettingsPageHandler() {
         );
     });
 }
+$(function () {
```
jacob-js, line 439: master has passed over these changes. If you have additional lines of code, ensure they are in the right route handlers.

modules/smtp/modules.php
kroky: I fail to see how repeating 49 lines of code will be easier to maintain in the future. This is the code I see that is 100% repeated:
```
                        if (new DateTime($msg_headers['X-Schedule']) <= new DateTime()) {
                            $profile = Hm_Profiles::get($msg_headers['X-Profile-ID']);
                            if (!$profile) {
                                $profiles = Hm_Profiles::search('server', $imap_details['server']);

                                if (!$profiles) {
                                    Hm_Debug::add(sprintf('ERRCannot find profiles corresponding with IMAP server: %s', $imap_details['server']));
                                    return;
                                }
                                $profile = $profiles[0];
                            }

                            $smtp = Hm_SMTP_List::connect($profile['smtp_id'], false);

                            if (smtp_authed($smtp)) {
                                if (isset($msg_headers['X-Delivery'])) {
                                    $from_params = 'RET=HDRS';
                                    $recipients_params = 'NOTIFY=SUCCESS,FAILURE';
                                } else {
                                    $from_params = '';
                                    $recipients_params = '';
                                }

                                $recipients = [];
                                foreach (['To', 'Cc', 'Bcc'] as $fld) {
                                    if (array_key_exists($fld, $msg_headers)) {
                                        $recipients = array_merge($recipients, Hm_MIME_Msg::find_addresses($msg_headers[$fld]));
                                    }
                                }

                                $msg_content = $imap->get_message_content($msg['uid'], 0);
                                $from = process_address_fld($msg_headers['From']);

                                $err_msg = $smtp->send_message($from[0]['email'], $recipients, $msg_content, $from_params, $recipients_params);

                                if (!$err_msg) {
                                    if ($imap->message_action('DELETE', [$msg['uid']])) {
                                        $imap->message_action('EXPUNGE', [$msg['uid']]);
                                    }
                                    save_sent_msg($handler, $server_id, $imap, $imap_details, $msg_content, $msg['uid'], false);
                                    return true;
                                }
                            }
                        }
                    } catch (Exception $e) {
                        Hm_Debug::add(sprintf('ERRCannot send message: %s', $msg_headers['subject']));
                    }
                    return;
                }}
```
Please find a way to abstract it and reuse.

amaninyumu1: Hello @kroky  I am making changes. please review

amaninyumu1:
Hello @jacob-js I don't understand very well please

modules/smtp/modules.php
kroky: Yes, it is better. You can move all the code in send_message_if_scheduled to send_scheduled_message and remove send_message_if_scheduled. There is no need to call 2 separate functions with exactly the same arguments and make one call the other without doing anything else - just use one function.

amaninyumu1: @kroky I just made some changes. Please review

kroky: Yes, it is fine.

jacob-js:
> Hello @jacob-js I don't understand very well please

I added more details in your inbox, I hope it becomes clear!

modules/core/site.js
```diff
@@ -2012,6 +2025,12 @@ $(function() {
         try { navigator.registerProtocolHandler("mailto", "?page=compose&compose_to=%s", "Cypht"); } catch(e) {}
     }
 
+    if (hm_page_name() == 'home') {
+        $('.pw_update').on("click", function() { update_password($(this).data('id')); });
```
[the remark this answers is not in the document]
amaninyumu1: @jacob-js I made some modifications by removing the hm_page_name() in the core module js file. please review

amaninyumu1:
Hello @jacob-js  I make changes please review

modules/imap/js_modules/route_handlers.js
```diff
@@ -9,20 +10,78 @@ function applyImapMessageListPageHandlers(routeParams) {
...
+    $('.toggle_link').on("click", function (e) {
         e.preventDefault();
         Hm_Message_List.toggle_rows();
     });
 
+    if (listPath.substr(0, 4) === 'imap') {
+        setup_imap_folder_page(listPath);
+    } else if (listPath === 'combined_inbox') {
+        setup_imap_message_list_content_page();
+    }
```
jacob-js, line 27: Why are you doing this?

modules/imap/js_modules/route_handlers.js
```diff
@@ -34,6 +93,11 @@ function applyImapMessageContentPageHandlers(routeParams) {
     imap_setup_snooze();
     imap_setup_tags();
 
+    if (routeParams && routeParams.uid && routeParams.list_path) {
+        imap_setup_message_view_page(routeParams.uid, null, routeParams.list_path, handleExternalResources);
+    } else if (hm_list_path().substr(0, 4) === 'imap') {
+        imap_setup_message_view_page();
+    }
```
jacob-js, line 100: Why?

modules/imap/js_modules/route_handlers.js
```diff
@@ -51,4 +115,12 @@ function applyImapMessageContentPageHandlers(routeParams) {
...
     if (window.pgpMessageContentPageHandler) pgpMessageContentPageHandler();
     if (window.wpMessageContentPageHandler) wpMessageContentPageHandler(routeParams);
+}
+
+function applyImapServersPageHandlers() {
+    imap_setup_server_page();
+}
+
+function applyImapInfoPageHandlers() {
+    setTimeout(imap_status_update, 100);
```
jacob-js, line 125: You added new handlers, but for which pages are they getting called?
FYI handlers are bound to appropriate pages as in https://github.com/cypht-org/cypht/blob/master/modules/core/navigation/routes.js

modules/core/site.js
```diff
@@ -2498,3 +2511,29 @@ const observeMessageTextMutationAndHandleExternalResources = (inline) => {
         });
     }
 };
+
+var setup_nexter_date = function(callback) {
```
jacob-js, line 2515: For the naming convention, it’s better to stick to camel case for JavaScript code and avoid using `var` when creating functions and variables. Instead, declare them as named functions, as functions are hoisted and available in any JavaScript file where they are used.

modules/imap/js_modules/route_handlers.js
amaninyumu1: The if condition (listPath.substr(0, 4) === 'imap') allows IMAP folders to be treated separately, ensuring that messages are displayed correctly for each folder
and The else if (listPath === 'combined_inbox') condition ensures that when the user navigates to an inbox that combines messages from multiple folders, the code calls a suitable function to display that view.

jacob-js: Aren't messages correctly displayed currently for each folder?

amaninyumu1: yes this is the case for the Scheduled folder

amaninyumu1:
Hello @jacob-js . I just made some changes. try, please review

jacob-js:
Great, please resolve the threads for which you believe you have met the change criteria.

modules/imap/js_modules/route_handlers.js
amaninyumu1: The logic behind this condition is to call imap_setup_message_view_page with different sets of arguments depending on the presence of routeParams.uid and routeParams.list_path. If these parameters are available, it means that the user has probably selected a specific message to view, and the function is called with message and path details, as well as a handler for external resources (handleExternalResources).

modules/imap/js_modules/route_handlers.js
amaninyumu1: `else if (listPath === 'combined_inbox') {setup_imap_message_list_content_page();}`
I just deleted this part, a thinking error on my part.

modules/imap/js_modules/route_handlers.js
jacob-js: I did not ask for a code explanation, what I want you to explain is the reason behind doing it, because:
1. in most cases you won't get the correct list_path parameter from `hm_list_path()`. It is supposed to be provided by the route parameters.
2. imap_setup_message_view_page() is meaningless without any parameter, what message do you want it to display?

modules/core/site.js
amaninyumu1: Hello @jacob-js . I removed the sprint function and used template literals. Please review

modules/imap/js_modules/route_handlers.js
```diff
@@ -9,31 +10,91 @@ function applyImapMessageListPageHandlers(routeParams) {
...
-    $('.toggle_link').on("click", function(e) {
+
+    $('.toggle_link').on("click", function (e) {
         e.preventDefault();
         Hm_Message_List.toggle_rows();
     });
 
+    if (listPath.substr(0, 4) === 'imap') {
+        setup_imap_folder_page(listPath);
+    }
```
jacob-js, line 27: Why are you still doing this?

modules/imap/js_modules/route_handlers.js
```diff
@@ -9,31 +10,91 @@ function applyImapMessageListPageHandlers(routeParams) {
...
+    return async function () {
         const [refreshIntervalId, abortController] = await setupPageResult;
         abortController.abort();
         clearInterval(refreshIntervalId);
     }
 }
 
+function handleNexterDateAction(e) {
+        $('.nexter_dropdown').hide();
+        var ids = [];
```
jacob-js, line 42: Please avoid using var

modules/imap/js_modules/route_handlers.js
```diff
@@ -9,31 +10,91 @@ function applyImapMessageListPageHandlers(routeParams) {
...
+}
+
 function applyImapMessageContentPageHandlers(routeParams) {
     imap_setup_message_view_page(routeParams.uid, null, routeParams.list_path, handleExternalResources);
     imap_setup_snooze();
     imap_setup_tags();
 
+    if (routeParams && routeParams.uid && routeParams.list_path) {
+        imap_setup_message_view_page(routeParams.uid, null, routeParams.list_path, handleExternalResources);
+    }
```
jacob-js, line 97: Why are you still doing this?

modules/imap/js_modules/route_handlers.js
amaninyumu1: I removed this condition

modules/imap/js_modules/route_handlers.js
amaninyumu1: I changed var to let. please review

modules/imap/js_modules/route_handlers.js
amaninyumu1: Please Review. I removed this condition

modules/imap/js_modules/route_handlers.js
```diff
@@ -1,5 +1,6 @@
 function applyImapMessageListPageHandlers(routeParams) {
     const setupPageResult = setup_imap_folder_page(routeParams.list_path);
+    const listPath = routeParams.list_path;
```
jacob-js, line 3: Please remove this unnecessary declaration.

amaninyumu1:
Hello @mercihabam. I wanted to ask if there is any other guidance on this PR. THANKS

modules/core/functions.php
```diff
@@ -624,20 +624,98 @@ function check_file_upload($request, $key) {
     return true;
 }}
 
-function privacy_setting_callback($val, $key, $mod) {
-    $setting = Hm_Output_privacy_settings::$settings[$key];
-    $key .= '_setting';
-    $user_setting = $mod->user_config->get($key);
-    $update = $mod->request->post['update'];
+if (!hm_exists('get_nexter_date')) {
+function get_nexter_date($format, $only_label = false) {
```
mercihabam, line 628: It's always good practice to name functions based on what the developer intends to achieve with their declaration. The current name "get_nexter_date" doesn't convey this effectively.

modules/smtp/site.js
```diff
@@ -381,3 +390,264 @@ function smtpSettingsPageHandler() {
...
+    if (!hm_is_logged()) {
+        return;
+    }
+    if (hm_page_name() === 'settings') {
```
mercihabam, line 397: Please resolve this thread: https://github.com/cypht-org/cypht/pull/1091/files#r1818718989

modules/core/functions.php
amaninyumu1: I just renamed this function get_scheduled_date

modules/smtp/modules.php
marclaporte: @amaninyumu1 What do you think?

modules/imap/functions.php
```diff
@@ -318,8 +322,8 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
                     array('checkbox_callback', $id),
                     array('safe_output_callback', 'source', $source, $icon),
                     array('safe_output_callback', 'from'.$nofrom, $from, null, str_replace(array($from, '<', '>'), '', $msg['from'])),
-                    array('subject_callback', $subject, $url, $flags, null, $preview_msg),
```
kroky, line 321: Why do you remove the preview message parameter pass here? Doesn't seem related with the functionality you are adding.

modules/smtp/modules.php
kroky: Yes, this was resolved.

kroky:
I added just one small question. We should be good to merge after that.

modules/imap/js_modules/utils/handleNexterDateAction.js
```diff
@@ -0,0 +1,55 @@
+function handleNexterDateAction(e) {
+    let reload_and_redirect = function () {
+        Hm_Folders.reload_folders(true);
+        let path = hm_list_parent() ? hm_list_parent() : hm_list_path();
+        window.location.replace('?page=message_list&list_path=' + path);
```
mercihabam, line 5: We don't do this to change routings. You should use [navigate()](https://github.com/cypht-org/cypht/blob/master/modules/core/navigation/navigation.js#L47)

modules/profiles/hm-profiles.php
```diff
@@ -109,4 +109,14 @@ public static function loadLegacy($hmod) {
             }
         }
     }
+
+    public static function search($fld, $value) {
```
mercihabam, line 113: Make as much effort as possible to avoid abbreviations in code because such code is harder to read and even debug.

mercihabam:
@amaninyumu1 in the future, when making code changes, please ensure the commit history is clean by avoiding combining all changes into the initial commit.

amaninyumu1:
@mercihabam Understood. Thanks.

modules/profiles/hm-profiles.php
amaninyumu1: `$fld is $field` I added a little documentation to avoid confusion. Please Review

language/en.php
```diff
@@ -645,7 +646,6 @@
     'Are you sure you want to send this message?' => false,
     'IMAP and JMAP Servers' => false,
     'Junk' => false,
-    'Trash' => false,
```
marclaporte, line 648: https://en.m.wikipedia.org/wiki/Atomic_commit

language/fa.php
```diff
@@ -679,7 +680,6 @@
     'Are you sure you want to send this message?' => false,
     'IMAP and JMAP Servers' => false,
     'Junk' => false,
-    'Trash' => false,
```
marclaporte, line 682: https://en.m.wikipedia.org/wiki/Atomic_commit

modules/core/site.js
```diff
@@ -1,18 +1,18 @@
 'use strict';
 
 /* swipe event handler */
-var swipe_event = function(el, callback, direction) {
+var swipe_event = function (el, callback, direction) {
```
marclaporte, line 4: Mixing code reformating with other changes makes it difficult to read.
https://en.wikipedia.org/wiki/Atomic_commit

amaninyumu1:
Hello @marclaporte, @mercihabam  , @kroky I wanted to ask if there are any other guides on this PR. Thank you Please review

marclaporte:
> I wanted to ask if there are any other guides on this PR

One of the tests is failing.

amaninyumu1:
Hello @marclaporte all tests pass successfully

kroky:
I am fine with the changes. If @mercihabam approves as well, we can merge.

modules/profiles/hm-profiles.php
mercihabam: Adding documentation is good but not what I was asking for. Please use full names.

modules/imap/js_modules/utils/handleNexterDateAction.js
```diff
@@ -0,0 +1,55 @@
...
+                ],
+                function (res) {
+                    if (res.scheduled_msg_count > 0) {
+                        reload_and_redirect();
+                    }
+                }
+            );
+    });
+
+    setupNexterDateSnooze(function () {
```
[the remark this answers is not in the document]
mercihabam: There are some more occurrences that you have not yet renamed.

modules/imap/js_modules/utils/handleNexterDateAction.js
```diff
@@ -0,0 +1,67 @@
...
+
+        try {
+            Hm_Folders.reload_folders(true);
+            let path = hm_list_parent() ? hm_list_parent() : hm_list_path();
+            await navigate(`?page=message_list&list_path=${path}`);
+        } catch (error) {
+            Hm_Notices.show([`ERR${error.message}`]);
+        } finally {
+            isReloading = false;
+        }
```
mercihabam, line 16: What error are you expecting here?

modules/imap/js_modules/utils/handleNexterDateAction.js
```diff
@@ -0,0 +1,67 @@
+function handleNexterDateAction(e) {
+    let isReloading = false;
+
+    let reload_and_redirect = async function () {
+        if (isReloading) { return; }
+        isReloading = true;
+
+        try {
+            Hm_Folders.reload_folders(true);
+            let path = hm_list_parent() ? hm_list_parent() : hm_list_path();
```
mercihabam, line 10: I'm sure you won't see the values: ` hm_list_parent() , hm_list_path()` referenced anywhere because route parameters are handled by https://github.com/cypht-org/cypht/blob/master/modules/core/navigation/utils.js

modules/imap/js_modules/route_handlers.js
```diff
@@ -4,10 +4,15 @@ function applyImapMessageListPageHandlers(routeParams) {
     imap_setup_snooze();
     imap_setup_tags();
 
+    Hm_Message_List.set_row_events();
+
+    handleNexterDateAction();
+
+    if (window.githubMessageListPageHandler) githubMessageListPageHandler(routeParams);
```
mercihabam, line 11: This line should not be in your code

mercihabam:
> @amaninyumu1 in the future, when making code changes, please ensure the commit history is clean by avoiding combining all changes into the initial commit.

Please pay attention to the above statement.

Have a commit for each thread you resolve, that way, it is easy to keep up with the scope of the changes you do.

modules/imap/hm-imap.php
```diff
@@ -889,11 +889,7 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
             if ($this->is_supported( 'X-GM-EXT-1' )) {
                 $command .= 'X-GM-MSGID X-GM-THRID X-GM-LABELS ';
             }
-            $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID X-SNOOZED)]";
-            if ($include_content_body) {
```
josaphatim, line 893: Hello @amaninyumu1. Can you please restore this deleted change.

modules/imap/js_modules/utils/handleNexterDateAction.js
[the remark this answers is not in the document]
amaninyumu1: I just renamed just for the functions, so I ask if it is necessary to rename even the classes and css ids. Thanks

modules/imap/js_modules/utils/handleNexterDateAction.js
amaninyumu1: I just added some changes and use the `showRoutingToast();` and `hideRoutingToast();` method to show and hide the redirect message. please review

mercihabam:
@amaninyumu1 you need to correct the commit history before I give you any other round of review.

amaninyumu1:
Hello @mercihabam I merged the two commits that were before to make one. Maybe it's me who doesn't understand the maneuver to follow. please guide me

mercihabam:
@amaninyumu1 I did not ask you to combine the commits into one. I'm telling you to do the opposite. Please re-read carefully the related comments!

amaninyumu1:
I understand I misunderstood the comment. To be honest about the commit history, I don't know how to go about it. please guide me. THANKS

amaninyumu1:
Hello @mercihabam , @kroky . please merge this PR if there is anything to fix we can do it later, we risk returning to square one with this presence repeating conflicts

mercihabam:
@amaninyumu1, I'm sorry, but we cannot merge this as long as it still contains poor code that risks introducing regressions.
Also, the commit history! I still need to emphasize that aspect. You recently pushed a commit, but what changes did it bring to your existing code?

amaninyumu1:
if you can't merge this while it still contains poor code that risks introducing regressions. I'm ready to fix that for the PR to merge. But when you say that it's the commit history that blocks evolution... I don't understand that

marclaporte:
> But when you say that it's the commit history that blocks evolution... I don't understand that

Please see: https://github.com/cypht-org/cypht/pull/1091#issuecomment-2546037976

The current situation is causing more work for the reviewers because they need to re-review everything instesd of just your latest modifications. Once everything is done, there will be the option to squash commits.

amaninyumu1:
Okay I understand @marclaporte , soon I will make a commit for each change. because going back to the old history risks losing the current changes.

modules/imap/js_modules/utils/handleNexterDateAction.js
mercihabam: Look how the `navigate()` function is implemented. You don't need to add any more calls to `showRoutingToast()` and `hideRoutingToast()`.

modules/imap/js_modules/utils/handleNexterDateAction.js
mercihabam: No error is directly thrown in `navigate()` or `Hm_Folders.reload_folders()`. So, you should not expect any failure at all.

modules/imap/js_modules/utils/handleNexterDateAction.js
```diff
@@ -0,0 +1,68 @@
+function processNextActionDate(e) {
+    let isReloading = false;
+
+    let reload_and_redirect = async function () {
+        if (isReloading) {
+            return;
+        }
+        isReloading = true;
+
+        try {
```
mercihabam, line 10: This block remains unnecessary.
Also, why do you need the state `isReloading`?

mercihabam:
Kudos, @amaninyumu1, for resolving all the threads! Now, let's review the end-user functionality of what the code produces. Please include a video demonstration in the chat.

amaninyumu1:
Hello @mercihabam I just added a new commit to fix server error Uncaught Error: Call to undefined method Hm_Mailbox::get_mailbox_status() which occurred after rebasing with the master branch. Please review

amaninyumu1:
For the video demonstration I am blocked by this error server ` Undefined array key "list_page" in modules\imap\handler_modules.php which blocks the display of messages in all folders

modules/smtp/modules.php
```diff
@@ -888,6 +904,97 @@ protected function output() {
...
+        $scheduled_msg_count = 0;
+        $new_schedule_date = $form['schedule_date'];
+        if ($form['schedule_date'] != 'now') {
+            $new_schedule_date = get_scheduled_date($form['schedule_date']);
+        }       
+        $ids = explode(',', $form['scheduled_msg_ids']);
+        foreach ($ids as $msg_part) {
+            list($imap_server_id, $msg_id, $folder) = explode('_', $msg_part);
+            $cache = Hm_IMAP_List::get_cache($this->cache, $imap_server_id);
+            $imap = Hm_IMAP_List::connect($imap_server_id, $cache);
```
kroky, line 977: Now that EWS has been merged to master, we bypass direct imap object usage from Hm_IMAP_List in all handler modules and instead use Hm_Mailbox which is a bridge between handlers and imap/ews implementation classes. Please update this code according to what is currently present in imap handler_modules after merging (or rebasing) latest mater.

amaninyumu1:
Hello @marclaporte , @mercihabam , @kroky , @josaphatim . After making several adjustments to this PR, I noticed that the expected results caused unexpected contingencies, making it difficult to achieve the initially defined objectives.

In order to ensure better clarity and simplify the correction process, I propose to close this PR and start from a fresh start. This will reorganize the code more efficiently and avoid the problems encountered so far. Thanks

josaphatim:
@kroky, @mercihabam can check changes ? Many changes have been made to Cypht that I was not aware of.

modules/smtp/modules.php
```diff
@@ -1540,6 +1557,105 @@ protected function output() {
...
+    if (!($this->module_is_supported('imap') || $this->module_is_supported('profiles'))) {
+        return;
+    }
+
+    $servers = Hm_IMAP_List::dump();
+    $scheduled_msg_count = 0;
+
+    foreach (array_keys($servers) as $server_id) {
+        $cache = Hm_IMAP_List::get_cache($this->cache, $server_id);
+        $imap = Hm_IMAP_List::connect($server_id, $cache);
```
kroky, line 1579: This is no longer the way to get a connection to imap server. Imap, Jmap and EWS were unified to work behind a bridge called Mailbox. See other handler modules for examples how to retrieve a mailbox and work with it. Basically, anything that works with IMAP should work with EWS as well if you go through the bridge.

modules/smtp/functions.php
```diff
@@ -61,3 +61,92 @@ function get_reply_type($request) {
...
+        Hm_Debug::add(sprintf('ERRCannot send message: %s', $msg_headers['subject']));
+    }
+    return false; 
+}}
+
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('reschedule_message_sending')) {
+function reschedule_message_sending($handler, $imap, $msg_id, $folder, $new_date) {
```
kroky, line 124: This should receive a Mailbox object, not imap.

modules/smtp/functions.php
```diff
@@ -61,3 +61,92 @@ function get_reply_type($request) {
         return false;
     }
 }
+
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('send_scheduled_message')) {
+function send_scheduled_message($handler, $imap, $folder, $msg_id, $send_now = false) {
```
kroky, line 69: This should receive a Mailbox object, not imap.

kroky:
There seems to be a problem with the unit tests...

modules/smtp/functions.php
josaphatim: Thank you @kroky. I just need to rename it but the object is of type Hm_Mailbox if you see the way it is used you will notice that.

josaphatim:
> There seems to be a problem with the selenium tests...

Let's work on this

amaninyumu1:
[Demonstration-videos](https://github.com/amaninyumu1/cypht-test)
@mercihabam Please clone the repository and view. thanks

amaninyumu1:
Hello @mercihabam , @kroky Can you please check this PR, I have already sent the demonstration videos. thanks

mercihabam:
> Hello @mercihabam , @kroky Can you please check this PR, I have already sent the demonstration videos. thanks

Will check ASAP.

modules/smtp/functions.php
```diff
@@ -61,3 +61,92 @@ function get_reply_type($request) {
...
+                $profiles = Hm_Profiles::search('server', $mailbox_details['server']);
+
+                if (!$profiles) {
+                    Hm_Debug::add(sprintf('ERRCannot find profiles corresponding with MAILBOX server: %s', $mailbox_details['server']));
+                    return false;
+                }
+                $profile = $profiles[0];
+            }
+
+            $smtp = Hm_SMTP_List::connect($profile['smtp_id'], false);
```
kroky, line 90: Now that EWS has been merged, we no longer use direct smtp connections. Please go through Mailbox class like other smtp modules do it.

modules/imap/functions.php
```diff
@@ -1662,3 +1666,39 @@ function flattenMessagesLists($messagesLists, $listSize) {
 
     return ['messages' => $endList, 'offsets' => $sizesTaken];
 }
+
+if (!hm_exists('save_sent_msg')) {
+function save_sent_msg($handler, $imap_id, $imap, $imap_details, $msg, $msg_id, $show_errors = true) {
```
kroky, line 1671: If $imap is a mailbox, call it so, otherwise it is confusing.

modules/imap/functions.php
```diff
@@ -1662,3 +1666,39 @@ function flattenMessagesLists($messagesLists, $listSize) {
...
+            return;
+        }
+        $sent_folder = $auto_sent['sent'];
+    }
+    if (!$sent_folder) {
+        Hm_Debug::add(sprintf("Unable to save sent message, no sent folder for IMAP %s", $imap_details['server']));
+    }
+    $uid = null;
+    if ($sent_folder) {
+        Hm_Debug::add(sprintf("Attempting to save sent message for IMAP server %s in folder %s", $imap_details['server'], $sent_folder));
```
kroky, line 1689: This might not be an IMAP server but JMAP or EWS. Servers list has a type - use that in the edebug/error messages.

modules/imap/functions.php
```diff
@@ -1662,3 +1666,39 @@ function flattenMessagesLists($messagesLists, $listSize) {
...
+        if (! $imap->store_message($sent_folder, $msg)) {
```
kroky, line 1690: store_message should already return the uid of the new msg - no need to get the list of messages once again to retrieve the uid.

mercihabam:
> [Demonstration-videos](https://github.com/amaninyumu1/cypht-test) @mercihabam Please clone the repository and view. thanks

@amaninyumu1, here is my feedback:

- Scheduling takes too long to proceed.
- The confirmation message pops up while the “Please wait…” message is still displayed.
- The scheduled mailbox header has an action labeled “Scheduled sending,” which should be renamed to “Reschedule” for better clarity.
- Rescheduling should not redirect or reload folders.

Additionally, I would like to see a demo where you schedule a message to be sent at a specific time, then wait until that time arrives and show the message being sent.

lib/servers.php
```diff
@@ -273,4 +273,27 @@ private static function match($server, $user, $name) {
         }
         return false;
     }
+
+    private static function appendPpasswordAndUsername(array $server) {
```
mercihabam, line 277: Is the double `p` a typo?

mercihabam: More importantly, where does the need to have the properties "pass, user,  password, username" on the server come from? Why can't the `pass` and `user` be solely used?

modules/core/functions.php
```diff
@@ -637,3 +637,99 @@ function privacy_setting_callback($val, $key, $mod) {
...
+    } elseif ($format == 'next_week') {
+        $date_string = 'next week 08:00';
+        $label = 'Next week';
+    } elseif ($format == 'next_month') {
+        $date_string = 'next month 08:00';
+        $label = 'Next month';
+    } else {
+        $date_string = $format;
+        $label = 'Certain date';
+    }
```
mercihabam, line 661: This is a typical case where you would use a [switch statement](https://www.php.net/manual/en/control-structures.switch.php).

modules/core/functions.php
```diff
@@ -637,3 +637,99 @@ function privacy_setting_callback($val, $key, $mod) {
...
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('parse_nexter_header')) {
+    function parse_nexter_header($header, $name)
+    {
+        $header = str_replace("$name: ", '', $header);
+        $result = [];
+        foreach (explode(';', $header) as $kv)
```
mercihabam, line 724: Please do as much as possible to avoid acronyms and abbreviations when writing code. What do you mean by `$kv`?

modules/core/message_list_functions.php
```diff
@@ -335,11 +335,11 @@ function subject_callback($vals, $style, $output_mod) {
  */
 if (!hm_exists('date_callback')) {
 function date_callback($vals, $style, $output_mod) {
-    $snooze_class = isset($vals[2]) && $vals[2]? ' snoozed_date': '';
+    $nexter_class = isset($vals[2]) && $vals[2]? ' nexter_date': '';
```
mercihabam, line 338: The word "nexter" does not mean anything. Please rename appropriately.

mercihabam: Also, it seems like you're breaking something minor by removing the `snoozed_date` class. Are you sure that class was not supposed to be there?

modules/core/site.js
```diff
@@ -2409,3 +2409,55 @@ const observeMessageTextMutationAndHandleExternalResources = (inline) => {
...
+        } else {
+            $('.nexter_date_picker').css({ 'border': 'unset', 'border-top': '1px solid #ddd' });
+        }
+    });
+    $(document).on('change', '.nexter_input_date', function (e) {
+        if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
+            $('.nexter_input').val($(this).val()).trigger('change');
+        }
+    });
+    $(document).on('change', '.nexter_input', callback);
```
mercihabam, line 2436: The issue with event delegation at this stage is the risk of attaching the event handler multiple times when the wrapping function is called, which could result in the action being executed multiple times.

You should select and attach the event handler to the target element itself.

modules/core/site.js
```diff
@@ -2409,3 +2409,55 @@ const observeMessageTextMutationAndHandleExternalResources = (inline) => {
...
+        } else {
+            $('.nexter_date_picker_snooze').css({ 'border': 'unset', 'border-top': '1px solid #ddd' });
+        }
+    });
+    $(document).on('change', '.nexter_input_date_snooze', function (e) {
+        if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
+            $('.nexter_input_snooze').val($(this).val()).trigger('change');
+        }
+    });
+    $(document).on('change', '.nexter_input_snooze', callback);
```
mercihabam, line 2462: The previous comment about event delegations applies here as well.

modules/imap/functions.php
```diff
@@ -251,10 +251,14 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
             $nofrom = ' nofrom';
         }
         $is_snoozed = !empty($msg['x_snoozed']) && hex2bin($msg['folder']) == 'Snoozed';
+        $is_scheduled = !empty($msg['x_schedule']) && hex2bin($msg['folder']) == 'Scheduled';
         if ($is_snoozed) {
-            $snooze_header = parse_snooze_header('X-Snoozed: '.$msg['x_snoozed']);
+            $snooze_header = parse_nexter_header('X-Snoozed: '.$msg['x_snoozed'], 'X-Snoozed');
```
mercihabam, line 256: This is incoherent. You are getting the `snooze_header` though you seem to have replaced the correct function `parse_snooze_header` with `parse_nexter_header` which by its name does not intend to produce the expected output.

modules/imap/output_modules.php
```diff
@@ -206,9 +206,9 @@ protected function output() {
                             $txt .= $this->html_safe($value).'</th></tr>';
                         }
                         elseif ($fld == 'x-snoozed') {
-                            $snooze_header = parse_snooze_header($value);
+                            $snooze_header = parse_nexter_header($value, 'X-Snoozed');
```
mercihabam, line 209: snooze ≠ nexter

modules/smtp/functions.php
```diff
@@ -61,3 +61,92 @@ function get_reply_type($request) {
         return false;
     }
 }
+
+/**
+ * @subpackage smtp/functions
+ */
+if (!hm_exists('send_scheduled_message')) {
+function send_scheduled_message($handler, $mailbox, $folder, $msg_id, $send_now = false) {
```
mercihabam, line 69: Why should an unsent message go through the IMAP server instead of being stored on the user's server (e.g., database or session)? Considering that each request to the IMAP server is time-consuming, it would be far more efficient to handle such actions locally, eliminating the need for a connection to an external server.

@kroky do you have an opinion?

modules/smtp/js_modules/route_handlers.js
```diff
@@ -109,16 +113,23 @@ function applySmtpComposePageHandlers() {
...
-            
-        
             if (handleMissingAttachment()) {
-                document.getElementsByClassName("smtp_send")[0].click();
+                if ($('.nexter_input').val()) {
+                    save_compose_state(false, true, $('.nexter_input').val(), function(res) {
+                        if (res.draft_id) {
+                            reset_smtp_form(false);
+                            Hm_Folders.reload_folders(true);
+                            Hm_Utils.redirect();
```
mercihabam, line 127: No need to redirect. Resetting the form and showing a confirmation message is enough.

modules/smtp/js_modules/route_handlers.js
```diff
@@ -109,16 +113,23 @@ function applySmtpComposePageHandlers() {
         }
 
         async function handleSendAnyway() {
-
-            if ($('.compose_draft_id').val() == '0') {
-            Hm_Notices.show([hm_trans('Please wait, sending message...')]);
-            await waitForValueChange('.compose_draft_id', '0');
+            if ($('.saving_draft').val() !== '0') {
```
mercihabam, line 116: Was the renaming from `compose_draft_id` to `saving_draft` and changing the condition from an affirmation to a negation necessary here?

modules/smtp/site.js
```diff
@@ -375,3 +389,25 @@ function smtpSettingsPageHandler() {
         );
     });
 }
+
+$(function() {
+    if (hm_is_logged()) {
+        var scheduled_msg_count = 0;
+        var sendScheduledMessages = function() {
```
mercihabam, line 396: I have mentioned this many times, and I’ll repeat it: you should avoid using `var` and instead choose between `const` and `let`.

modules/core/message_list_functions.php
amaninyumu1: the class was supposed to be there but schedule and snooze use the same component

lib/servers.php
```diff
@@ -274,7 +274,7 @@ private static function match($server, $user, $name) {
         return false;
     }
 
-    private static function appendPpasswordAndUsername(array $server) {
+    private static function appendPasswordAndUsername(array $server) {
         $server['password'] = $server['pass'];
         $server['username'] = $server['user'];
```
mercihabam, line 279: Why do you need these assignments? Can't `$server['pass']` and `$server['user']` be used?

modules/imap/output_modules.php
amaninyumu1: in the scheduled folder we have the possibility to perform the snoozed and schedule actions. but I just renamed nexter a delayed

amaninyumu1:
`$mailbox = new Hm_Mailbox($server_id, $user_config, $session, $config);` for this new way of initializing Mailbox triggers unexpected errors like: 
- `$mailbox->connect()` : Uncaught Error: Call to a member function `connect()` on null while the $mailbox variable and the connection are not null after inspection
`$mailbox->send_message($from[0]['email'], $recipients, $msg_content, $delivery_receipt);` : PHP Fatal error: Uncaught Error: Call to undefined method Hm_IMAP::send_message(). Is there a new way to send messages with Hm_Mailbox. thanks
@kroky  or @mercihabam  can have an opinion here.

amaninyumu1:
$mailbox = new Hm_Mailbox($server_id, $user_config, $session, $config); for this new way of initializing Mailbox triggers unexpected errors like:
$mailbox->connect() : Uncaught Error: Call to a member function connect() on null while the $mailbox variable and the connection are not null after inspection
$mailbox->send_message($from[0]['email'], $recipients, $msg_content, $delivery_receipt); : PHP Fatal error: Uncaught Error: Call to undefined method Hm_IMAP::send_message(). Is there a new way to send messages with Hm_Mailbox. thanks
@kroky or @mercihabam can have an opinion here.

modules/smtp/functions.php
kroky: Keeping it on the IMAP/JMAP/EWS server has the benefit of user being able to reuse the email if they use another webmail client or a normal mail client. It is usually preferred to store data on the server and keep cypht-stored data to a minimum as it is designed to be a webmail.

kroky:
@amaninyumu1 you cannot use send_message method on an IMAP account - what would it mean? Mailbox on top of SMTP or EWS can send out emails and send_message is OK there but not on an IMAP account. For IMAP, there is store_message which stores a message in a specific folder. Please check your calling code.

amaninyumu1:
@kroky  the message is already stored in the scheduled folder I would like to implement the `send now` functionality for scheduled messages I use IMAP account

kroky:
You can't send through an IMAP account. Use its corresponding SMTP or EWS account stored in a profile or use the default SMTP account. Show error if there is no configured account to send from.

amaninyumu1:
@kroky I think this method is better for sending a message `$mailbox = Hm_SMTP_List::connect($profile['smtp_id'], false);` as with IMAP account you cannot send a message directly. thanks

modules/smtp/functions.php
kroky: I don't think we can count them as Cypht-only. It is a normal email scheduled to be sent. Scheduling logic remains in Cypht but user can actually send it from another client if that's more convenient for them. Also, imagine they access the same mailbox with multiple cypht webmail instances - if we store in local cypht, we again loose the functionality. It doesn't really make sense to me to keep a scheduled email locally.

modules/core/site.js
```diff
@@ -528,11 +528,21 @@ function Message_List() {
...
+                    bval = $($('td', b)[Math.abs(fld)]).text().replace(/^\s+/g, '');
+                    break;
+                case 4:
+                default:
+                    aval = $('input', $($('td', a)[Math.abs(fld)])).val();
+                    bval = $('input', $($('td', b)[Math.abs(fld)])).val();
+                    break;
+            }
+            if (fld == 4 || fld == -4 || !fld) {
+                if (fld == -4) {
```
mercihabam, line 545: This rebase was not accurate. See the actual state in this [diff](https://github.com/cypht-org/cypht/commit/dedcfb7dfb571a1db79db839e4a787cc9a082680#diff-eef52370546e86abe4041840d8277b2a1fb9d3716ffc670451b4b839884bfa2e).

modules/core/site.js
```diff
@@ -1801,6 +1811,7 @@ var hasLeadingOrTrailingSpaces = function(str) {
 var Hm_Message_List = new Message_List();
 
 function sortHandlerForMessageListAndSearchPage() {
+    $('.combined_sort').on("change", function() { Hm_Message_List.sort($(this).val()); });
```
mercihabam, line 1814: This line also is invalid.

modules/core/site.js
```diff
@@ -2399,3 +2410,74 @@ const observeMessageTextMutationAndHandleExternalResources = (inline) => {
...
+        .off('change', '.nexter_input_date_snooze')
+        .on('change', '.nexter_input_date_snooze', function () {
+            if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
+                $('.nexter_input_snooze').val($(this).val()).trigger('change');
+            }
+        });
+
+    $(document)
+        .off('change', '.nexter_input_snooze')
+        .on('change', '.nexter_input_snooze', callback);
```
mercihabam, line 2482: See my previous [comment](https://github.com/cypht-org/cypht/pull/1091#discussion_r1933637951) about event delegation.

modules/core/site.js
```diff
@@ -2399,3 +2410,74 @@ const observeMessageTextMutationAndHandleExternalResources = (inline) => {
...
+        }
+    });
+
+    $('.nexter_input_date').off('change').on('change', function (e) {
+        if ($(this).val() && new Date().getTime() < new Date($(this).val()).getTime()) {
+            $('.nexter_input').val($(this).val()).trigger('change');
+        }
+    });
+
+    $('.nexter_input').off('change').on('change', callback);
```
mercihabam, line 2441: The `off → on` pattern is problematic. If you want to clean up event handlers, page handlers allow you to return a function that cleans up states or executes any necessary code when a page is unmounted. For example, see line 10 in [this file](https://github.com/cypht-org/cypht/blob/master/modules/imap/js_modules/route_handlers.js).

However, cleanup is not necessary here because each call to your handler (`setupActionSchedule`) will apply to a new and unique instance of the main element in the DOM.

modules/smtp/functions.php
mercihabam: Make sense, let's keep it that way.

modules/smtp/js_modules/route_handlers.js
mercihabam: Something is not right here. As the resulting side-effect, see the second point in this [comment](https://github.com/cypht-org/cypht/pull/1091#issuecomment-2621124309)

lib/servers.php
mercihabam: I don't understand why you would need to make these modifications because of the occurrences in modules\nux\modules.php. Are you trying to address some errors?

amaninyumu1: https://github.com/cypht-org/cypht/pull/1091/commits/b1c45e8253e26d29ffbe2323e1c1309981716149

modules/smtp/js_modules/route_handlers.js
amaninyumu1: this problem is already solved

amaninyumu1:
@mercihabam  For your feedback https://github.com/cypht-org/cypht/pull/1091#issuecomment-2621124309
Everything is good 
For the message sent at a certain time, the demonstration is in the videos that I had sent, but I had cut a little to show the final result seeing the state of the connection which is not stable. If there are other recommendations, I'm ready. thanks

amaninyumu1:
Hello @kroky On my side, everything is fine. I await your recommendations. thanks

mercihabam:
> other

Please do not omit any essential part in your demonstration. That part is crucial for your feature, which is why I asked for it to be included.

modules/core/site.js
mercihabam: You are still omitting important parts of the code while including unnecessary ones. Please refer to the version on master to pick the correct code.

amaninyumu1:
> > @mercihabam For your feedback [#1091 (comment)](https://github.com/cypht-org/cypht/pull/1091#issuecomment-2621124309) Everything is good For the message sent at a certain time, the demonstration is in the videos that I had sent, but I had cut a little to show the final result seeing the state of the connection which is not stable. If there are other recommendations, I'm ready. thanks
> 
> Please do not omit any essential part in your demonstration. That part is crucial for your feature, which is why I asked for it to be included.

Yes I understand you @mercihabam the functionality in the videos.  everything works

mercihabam:
No, it's not.

modules/imap/functions.php
```diff
@@ -1605,34 +1608,9 @@ function connect_to_imap_server($address, $name, $port, $user, $pass, $tls, $ima
     }
 }
 
-/**
- * @param array $sources
- * @param object $cache
- * @param array $search
- */
-function getCombinedMessagesLists($sources, $cache, $search) {
```
josaphatim: This too has been altered. Let me restore it. @amaninyumu1 be careful while doing rebase.

modules/imap/functions.php
```diff
@@ -331,9 +335,8 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
                     array('safe_output_callback', 'source', $source, $icon),
                     array('safe_output_callback', 'from'.$nofrom, $from, null, str_replace(array($from, '<', '>'), '', $msg['from'])),
                     array('subject_callback', $subject, $url, $flags, null, $preview_msg),
-                    array('date_callback', $date, $timestamp, $is_snoozed),
-                    array('icon_callback', $flags),
-                    array('dates_holders_callback', $msg['internal_date'], $msg['date']),
```
josaphatim, line 336: This line too.

josaphatim:
Hello @kroky. This PR is working and ready to be merged if you agree.

marclaporte:
This branch has conflicts that must be resolved
Use the [web editor](https://github.com/cypht-org/cypht/pull/1091/conflicts) or the  to resolve conflicts.
Conflicting files
modules/imap/handler_modules.php
modules/smtp/js_modules/route_handlers.js
modules/smtp/modules.php

amaninyumu1:
Hello @kroky. All conflits are resolved

modules/imap/functions.php
```diff
@@ -1728,3 +1732,39 @@ function sortCombinedMessages($list, $sort) {
...
+        }
+        $sent_folder = $auto_sent['sent'];
+    }
+    if (!$sent_folder) {
+        Hm_Debug::add(sprintf("Unable to save sent message, no sent folder for server %s %s", $mailbox->server_type(), $imap_details['server']));
+    }
+    $uid = null;
+    if ($sent_folder) {
+        Hm_Debug::add(sprintf("Attempting to save sent message for server %s in folder %s", $mailbox->server_type(), $imap_details['server'], $sent_folder));
+        if (! $mailbox->store_message($sent_folder, $msg)) {
```
kroky, line 1756: store message returns the UID, no need to get messages sorting by arrival and get it by the ID just to return the UID.

josaphatim: This was already solved. I think a wrong rebase deleted it.

mercihabam:
@amaninyumu1, I just opened my webmail and the first ajax request I got launched was `ajax_send_scheduled_messages` which responded with a 500 response showing me a "server error" message.

marclaporte:
> 500 response showing me a "server error" message.

Also reported at https://github.com/cypht-org/cypht/issues/1450

amaninyumu1:
> > 500 response showing me a "server error" message.
> 
> Also reported at #1450

https://github.com/cypht-org/cypht/pull/1454

PR #1093, opened 2024-06-17
[UDP] Sieve: Update henrique-borba/php-sieve-manager to 1.0.6

Shadow243:
Upgrade sieve to latest version 1.0.6 and correct sievefilters/modules.php at actions loop usage accordingly.

[https://github.com/cypht-org/php-sieve-manager/releases/tag/v1.0.6](https://github.com/cypht-org/php-sieve-manager/releases/tag/v1.0.6)

### DIFF
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -890,7 +890,7 @@ class Hm_Handler_sieve_save_filter extends Hm_Handler_Module {
                 $custom_condition->addCriteria($cond);
             }
         }
-
+        
         foreach ($actions as $action) {
             if ($action->action == 'discard') {
                 $custom_condition->addAction(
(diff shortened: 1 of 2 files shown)


marclaporte:
Will Cypht 1.4.x require similar updates?

modules/sievefilters/modules.php
```diff
@@ -909,42 +909,43 @@ public function process() {
...
                 $custom_condition->addAction(
-                    new \PhpSieveManager\Filters\Actions\RemoveFlagFilterAction([$action->value])
+                    new \PhpSieveManager\Filters\Actions\RemoveFlagFilterAction(['flags' => [$action->value]])
                 );
             }
             if ($action->action == 'move') {
                 $filter->addRequirement('fileinto');
                 $custom_condition->addAction(
-                    new \PhpSieveManager\Filters\Actions\FileIntoFilterAction([$action->value])
+                    new \PhpSieveManager\Filters\Actions\FileIntoFilterAction(['flags' => [$action->value]])
```
josaphatim, line 936: isn't mailbox ?

modules/sievefilters/modules.php
```diff
@@ -953,7 +954,7 @@ public function process() {
             if ($action->action == 'autoreply') {
                 $filter->addRequirement('vacation');
                 $custom_condition->addAction(
-                    new \PhpSieveManager\Filters\Actions\VacationFilterAction([$action->extra_option_value, $action->value])
+                    new \PhpSieveManager\Filters\Actions\VacationFilterAction(['reason' => $action->extra_option_value, 'flags' => [$action->value]])
```
josaphatim, line 957: reason and subject not flags

modules/sievefilters/modules.php
```diff
@@ -953,7 +954,7 @@ public function process() {
             if ($action->action == 'autoreply') {
                 $filter->addRequirement('vacation');
```
josaphatim, line 955: We can now remove all $filter->addRequirement as they are added dynamically in each action

composer.json
```diff
@@ -46,7 +46,7 @@
         "ext-openssl": "*",
         "ext-session": "*",
         "ezyang/htmlpurifier": "^4.17",
-        "henrique-borba/php-sieve-manager": "^1.0",
+        "henrique-borba/php-sieve-manager": "^1.0.6",
```
josaphatim, line 49: I think the removed code was correct. Can you confirm please @marclaporte ?

marclaporte: composer.lock will provide 1.0.6 so it doesn't matter much.

As per http://semver.org/, going from 1.0.5 to 1.0.6 should not require any changes for Cypht. If Cypht requires changes, maybe it should have been 1.1.0.

It's not critical now since the lib is not yet included in many projects. But in the future, let's try to adhere to SemVer for all code susceptible to be leveraged or included by another project.

Shadow243:
@josaphatim  Corrections added

josaphatim:
> Will Cypht 1.4.x require similar updates?

Not if we don't use php manager v1.0.5 or above

marclaporte:
> Not if we don't use php manager v1.0.5 or above

Cypht 1.4.x is supported until 2025-05 as per:
https://github.com/cypht-org/cypht/wiki/Lifecycle

And there could be other fixes later in 1.0.7 that we will want in Cypht 1.4.x and 2.1.x so we might as well keep everything aligned.

marclaporte:
It's happening: https://gitlab.com/tikiwiki/tiki/-/merge_requests/5291/diffs

marclaporte:
Backport to 1.4.x: https://github.com/cypht-org/cypht/pull/1134

Issue #1092, opened 2024-06-17
Duplicated profile for the same SMTP and IMAP servers


Baraka24:
### Version & Environment

All versions are concerned

### Steps to reproduce
1.  Go to the profiles page.
2. Attempting to create a profile with the same SMTP and IMAP server settings will result in a duplicate profile being created, which can lead to confusion.

Baraka24:
<img width="859" alt="profilesBug" src="https://github.com/cypht-org/cypht/assets/81784141/2fed9043-0818-44c5-9dc9-7c68e5fb0a63">

PR #1096, opened 2024-06-19
[FIX]Profiles: Fix duplicated profile for the same email adress

Baraka24:
[Related issue](https://github.com/cypht-org/cypht/issues/1092)

### DIFF
--- a/modules/profiles/modules.php
+++ b/modules/profiles/modules.php
@@ -133,7 +133,19 @@ class Hm_Handler_process_profile_update extends Hm_Handler_Module {
             $profile['id'] = $form['profile_id'];
             Hm_Profiles::edit($form['profile_id'], $profile);
         } else {
+            $profiles = $this->get('profiles');
+
+            foreach ($profiles as $existing_profile) {
+                if (
+                    ($existing_profile["address"] === $profile["address"] && $existing_profile["smtp_id"] === $profile["smtp_id"]) ||
+                    ($existing_profile["replyto"] === $profile["replyto"] && $existing_profile["smtp_id"] === $profile["smtp_id"])
+                ) {
+                    Hm_Msgs::add('ERRProfile with this email address or reply-to address already exists');
+                    return;
+                }
+            }
             Hm_Profiles::add($profile);
+            Hm_Msgs::add('Profile Created');
         }
 
         if ($default) {


modules/profiles/modules.php
```diff
@@ -133,7 +133,16 @@ public function process() {
             $profile['id'] = $form['profile_id'];
             Hm_Profiles::edit($form['profile_id'], $profile);
         } else {
+            $profiles = $this->get('profiles');
+
+        foreach ($profiles as $existing_profile) {
+            if ($existing_profile["smtp_id"] === $profile["smtp_id"]) {
```
Shadow243, line 139: Here you are testing with smtp_id, this will prevent from having many profiles even with different email on that smtp. If the idea is to prevent duplication, use the email field 'replyto' and 'address' instead.

Shadow243, line 139: Here you are testing with smtp_id, this will prevent from having many profiles even with different email on that smtp. If the idea is to prevent duplication, use the email field 'replyto' and 'address' instead.

Shadow243 (changes requested):
A part from that, That's looks good for me.

modules/profiles/modules.php
```diff
@@ -133,7 +133,16 @@ public function process() {
             $profile['id'] = $form['profile_id'];
             Hm_Profiles::edit($form['profile_id'], $profile);
         } else {
+            $profiles = $this->get('profiles');
+
+            foreach ($profiles as $existing_profile) {
+                if ($existing_profile["address"] === $profile["address"]) {
```
Shadow243, line 139: We must have smtp_id in this condition. The idea is to check the address but considering that we can have same adresses(emails) on a different smtp.

Shadow243 (changes requested):

Baraka24:
@Shadow243 Thanks for a kind review, I did not figure out some points, you make them clear to me and I applied changes.

Shadow243:
> @Shadow243 Thanks for a kind review, I did not figure out some points, you make them clear to me and I applied changes.

Perfect. What about 'replyto' ? I think we can add OR in that condition. What do you think about that? The idea is to prevent duplication with 'replyto' field.

Baraka24:
> What about 'replyto' ? I think we can add OR in that condition. What do you think about that? The idea is to prevent duplication with 'replyto' field.

Adding the 'replyto' field to the condition makes sense as we want to ensure that no profiles are duplicated based on the 'replyto' field as well.

PR #1098, opened 2024-06-20
[EHN]Email filtering: Cypht Sieve filters: Better handling of 'Stop processing more rules'

christer77:
Sometimes, you might set up several rules that could apply to a single message. If you created a rule to move all messages with a subject line of "Wedding plans" to a folder, for example, and then created another rule to delete all messages with attachments, you probably wouldn't want a message related to wedding plans that has an attachment to be deleted. To prevent this from happening, you can set the option Stop processing more rules on the first rule.

stop-filtering

### DIFF
--- a/modules/sievefilters/site.js
+++ b/modules/sievefilters/site.js
@@ -86,12 +86,6 @@ var hm_sieve_possible_actions = function() {
             extra_field: false
         },
         {
-            name: 'stop',
-            description: 'Stop Filtering',
-            type: 'none',
-            extra_field: false
-        },
-        {
             name: 'copy',
             description: 'Copy email to mailbox',
             placeholder: 'Mailbox Name (Folder)',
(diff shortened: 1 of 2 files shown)


marclaporte:
Related info in another system about when and how we want this: https://support.microsoft.com/en-us/office/stop-processing-more-rules-in-outlook-com-10dca09a-24c7-4c0d-abf3-9fa29fdc3230

josaphatim:
@christer77 I think you should also remove stop filtering in list of available actions as it is replaced by the checkbox.

christer77:
> @christer77 I think you should also remove stop filtering in list of available actions as it is replaced by the checkbox.

Done

modules/sievefilters/site.js
```diff
@@ -525,6 +531,7 @@ $(function () {
             edit_filter_modal.setTitle('Add Filter');
             current_account = $(this).attr('account');
             edit_filter_modal.open();
+            $('#stop_filtering').prop('checked', false);
```
josaphatim, line 534: I think you should not reset its value here

josaphatim:
@christer77 one last thing: can you reset priority, test and the new checkbox you added when clicking add filter ? See `$('.add_filter').on('click', function () `

josaphatim:
Now good! Thanks

PR #1100, opened 2024-06-21
Fixed phpunit broken test

josaphatim:
Follow up to https://github.com/cypht-org/cypht/pull/1099 to fix access to $imap_details when no imap auth server and also unit test

### DIFF
--- a/tests/phpunit/config.php
+++ b/tests/phpunit/config.php
@@ -41,7 +41,7 @@ class Hm_Test_User_File_Config extends TestCase {
         $this->assertEquals(array('imap_servers' => array(array())), $this->config->filter_servers());
 
         $this->config->set('imap_servers', array(array('default' => 1, 'server' => 'localhost')));
-        $this->assertEquals(array('imap_servers' => array(array('default' => 1, 'server' => 'localhost'))), $this->config->filter_servers());
+        $this->assertEquals(array(), $this->config->filter_servers());
 
         $this->config->set('imap_servers', array(array('pass' => 'foo', 'server' => 'localhost')));
         $this->config->set('no_password_save_setting', true);
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -1564,12 +1564,12 @@ class Hm_Handler_load_imap_servers_from_config extends Hm_Handler_Module {
             if (! empty($auth_server['sieve_config_host'])) {
                 $imap_details['sieve_config_host'] = $auth_server['sieve_config_host'];
             }
-        }
-        if (!$default_server_id) {
-            Hm_IMAP_List::add($imap_details);
-        } else {
-            // Perhaps something as changed
-            Hm_IMAP_List::edit($default_server_id, $imap_details);
+            if (!$default_server_id) {
+                Hm_IMAP_List::add($imap_details);
+            } else {
+                // Perhaps something as changed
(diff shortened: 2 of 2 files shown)


tests/phpunit/config.php
```diff
@@ -41,7 +41,7 @@ public function test_filter_servers() {
         $this->assertEquals(array('imap_servers' => array(array())), $this->config->filter_servers());
 
         $this->config->set('imap_servers', array(array('default' => 1, 'server' => 'localhost')));
-        $this->assertEquals(array('imap_servers' => array(array('default' => 1, 'server' => 'localhost'))), $this->config->filter_servers());
```
Shadow243, line 44: Oh, this is related to: https://github.com/cypht-org/cypht/pull/1101.

But I think we don't have to change anything here. I will close my PR, but take a look at it first. We can keep our test as it is and keep this https://github.com/cypht-org/cypht/pull/1099 working.

Shadow243 (changes requested):
We are actually trying to solve the same thing.

tests/phpunit/config.php
kroky: I don't think I understand the suggestion. This MR looks correct.

Shadow243: I was wondering why this test at this line: https://github.com/cypht-org/cypht/pull/1100/files#diff-f1440e4052e7c01c47ca8678b902fec4aaa196a0d617e3de83b150cbe877833fR44 would return an empty array even though we passed it data(ima p_servers) here : https://github.com/cypht-org/cypht/pull/1100/files#diff-f1440e4052e7c01c47ca8678b902fec4aaa196a0d617e3de83b150cbe877833fR43

I then seem to see as if this is trying to accommodate a change that I think is wrong here: https://github.com/cypht-org/cypht/pull/1099/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1536
I think yes we had to define `$default_server_id` and assign `$id` to it in the loop then here: https://github.com/cypht-org/cypht/pull/1099/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1 568 test if id exists for know if we are going to modify or add.

in this condition: https://github.com/cypht-org/cypht/pull/1099/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1568 we test if the default key exists then in it we are assigning `$default_server_id` = `$id` it doesn't seem clear I think.

This is why I proposed https://github.com/cypht-org/cypht/pull/1101/ to try to review the correction here https://github.com/cypht-org/cypht/pull /1101/Files#DIFF-B7FDF5637005EA55555542D9660FA0FE0FE278967CD7CE7DBD9BFA7C4E0ED8R149 and without changing anything in the unit test this will work as before and the corrections Bring in this PR https://github.com/cypht/cypht/ 9 will always work

Shadow243: Apart from that I have the error: 
photo_2024-06-25 14 41 31
On every reload imap_servers are incrementing

Shadow243: This will be solved from here: https://github.com/cypht-org/cypht/pull/1100/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1567

PR #1101, opened 2024-06-21
Refact #1099 to make tests work

Shadow243:
This change [https://github.com/cypht-org/cypht/pull/1099/files#diff-b7fdf5637005ea55ecbcc642d9660fa0fe278967cd7ce7ac7dbd9bfa7c4e0ed8R149](https://github.com/cypht-org/cypht/pull/1099/files#diff-b7fdf5637005ea55ecbcc642d9660fa0fe278967cd7ce7ac7dbd9bfa7c4e0ed8R149) affected the tests, upon rechecking I thought that a refactoring was necessary for the tests to pass without affecting the correction in the PR: https://github.com/cypht-org/cypht/pull/1099

### DIFF
--- a/lib/config.php
+++ b/lib/config.php
@@ -146,7 +146,10 @@ abstract class Hm_Config {
         foreach ($this->config as $key => $vals) {
             if (in_array($key, $excluded, true)) {
                 foreach ($vals as $index => $server) {
-                    if (!array_key_exists('server', $server)) {
+                    if (!empty($server['default'])) {
+                        $removed[$key][$index] = $server;
+                        unset($this->config[$key][$index]);
+                    } elseif (!array_key_exists('server', $server)) {
                         $removed[$key][$index] = $server;
                         unset($this->config[$key][$index]);
                     } else {
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -1533,6 +1533,7 @@ class Hm_Handler_load_imap_servers_from_config extends Hm_Handler_Module {
     public function process() {
         Hm_IMAP_List::init($this->user_config, $this->session);
         $default_server_id = false;
+        $has_default = false;
         foreach (Hm_IMAP_List::getAll() as $id => $server) {
             if ($this->session->loaded) {
                 if (array_key_exists('expiration', $server)) {
@@ -1541,8 +1542,9 @@ class Hm_Handler_load_imap_servers_from_config extends Hm_Handler_Module {
                 }
             }
             if (array_key_exists('default', $server) && $server['default']) {
-                $default_server_id = $id;
+                $has_default = true;
             }
+            $default_server_id = $id;
         }
         $auth_server = $this->session->get('imap_auth_server_settings', array());
         if (!empty($auth_server)) {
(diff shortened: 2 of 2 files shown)


Shadow243:
This PR is fixing this error too:
photo_2024-06-25 14 40 57

Baraka24:
@Shadow243, it is really a bug. You have 4 or more IMAP accounts, but you have not configured any.

modules/imap/handler_modules.php
```diff
@@ -1541,8 +1542,9 @@ public function process() {
                 }
             }
             if (array_key_exists('default', $server) && $server['default']) {
-                $default_server_id = $id;
+                $has_default = true;
             }
+            $default_server_id = $id;
```
kroky, line 1547: Why default_server_Id is every ID in the imap list? Shouldn't it be only the one marked as default or maybe the first one?

lib/config.php
```diff
@@ -146,7 +146,10 @@ public function filter_servers() {
         foreach ($this->config as $key => $vals) {
             if (in_array($key, $excluded, true)) {
                 foreach ($vals as $index => $server) {
-                    if (!array_key_exists('server', $server)) {
+                    if (!empty($server['default'])) {
```
josaphatim, line 149: This returns the error we wanted to fix. The goal of removing this code was to allow you to have the same unique identifier for the default server each time you reconnect. This resulted in losing the configured folders. With this we will have the same problem.

josaphatim:
> This PR is fixing this error too: photo_2024-06-25 14 40 57


@Shadow243 I might be a different issue. Cypht standalone is working fine. You need to fix the issue in Tiki

Shadow243:
> > This PR is fixing this error too: photo_2024-06-25 14 40 57
> 
> 
> 
> 
> 
> @Shadow243 I might be a different issue. Cypht standalone is working fine. You need to fix the issue in Tiki

I had the same issue with Cypht standalone.

kroky:
I am a bit lost here, can @josaphatim or @Shadow243 summarize what the original bug was, how it was fixed and why this current PR is needed? Some tests were failing after the original fix?

Shadow243:
Wil be solved from here: https://github.com/cypht-org/cypht/pull/1100/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1567

modules/imap/handler_modules.php
```diff
@@ -1564,12 +1566,12 @@ public function process() {
             if (! empty($auth_server['sieve_config_host'])) {
                 $imap_details['sieve_config_host'] = $auth_server['sieve_config_host'];
             }
-        }
-        if (!$default_server_id) {
-            Hm_IMAP_List::add($imap_details);
-        } else {
-            // Perhaps something as changed
-            Hm_IMAP_List::edit($default_server_id, $imap_details);
+            if (!$default_server_id) {
```
Shadow243, line 1569: This is what he did here: https://github.com/cypht-org/cypht/pull/1100/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1567

PR #1102, opened 2024-06-26
[ENH]Imap quota: cypht show space used/available

Baraka24:
[Related issue](https://github.com/cypht-org/cypht/issues/626l)

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -722,6 +722,20 @@ class Hm_Handler_imap_folder_expand extends Hm_Handler_Module {
                 $this->session->set('imap_prefetched_ids', array_unique($prefetched, SORT_NUMERIC));
             }
             $with_subscription = isset($this->request->post['subscription_state']) && $this->request->post['subscription_state'];
+            $cache = Hm_IMAP_List::get_cache($this->cache, $form['imap_server_id']);
+            $imap = Hm_IMAP_List::connect($form['imap_server_id'], $cache);
+            if (imap_authed($imap)) {
+                $quota_root = $imap->get_quota_root($folder ? $folder : 'INBOX');
+                $quota = $imap->get_quota($quota_root[0]['name']);
+                if ($quota) {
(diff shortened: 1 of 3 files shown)


Baraka24:
> [Related issue](https://github.com/cypht-org/cypht/issues/626l)

@marclaporte, @kroky See what I proposed for this: 
Capture d'écran 2024-06-27 224041
Waiting for suggestions. Thanks.

modules/imap/handler_modules.php
```diff
@@ -722,15 +722,23 @@ public function process() {
                 $this->session->set('imap_prefetched_ids', array_unique($prefetched, SORT_NUMERIC));
             }
             $with_subscription = isset($this->request->post['subscription_state']) && $this->request->post['subscription_state'];
+            $cache = Hm_IMAP_List::get_cache($this->cache, $form['imap_server_id']);
+            $imap = Hm_IMAP_List::connect($form['imap_server_id'], $cache);
+            if (imap_authed($imap)) {
+                $quota_root = $imap->get_quota_root('INBOX');
```
kroky, line 728: It seems a folder can be posted from incoming params above - use it instead of defaulting to INBOX. Also, when you default to INBOX, why do you get quota again on first result returned? Seems wrong to me. I think if you have just one progress bar under all folders on the account, you should use root quota and not get individual folders' ones.

modules/imap/handler_modules.php
```diff
@@ -722,15 +722,23 @@ public function process() {
...
+                $quota = $imap->get_quota($quota_root[0]['name']);
+                if($quota && $quota[0]['max'] > 0) {
+                    $this->out('quota', ceil(($quota[0]['current'] / $quota[0]['max']) * 100));
```
kroky, line 731: Please parse to int/float before dividing and check if max is not zero - both could throw errors in production.

Baraka24:
Thanks for a clear review @kroky. I did as required.

PR #1104, opened 2024-06-27
[ENH] Added functionality to enable or disable Sieve Filters with toggle button

amaninyumu1:


Shadow243 (changes requested):
Please make sure all tests pass successfully and also remove composer.lock it will be supported by this https://github.com/cypht-org/cypht/pull/1093 because we have some changes that must be applied in sievefilters/modules.php before.

marclaporte:
Sorry, but there is another conflict.

amaninyumu1:
hello @marclaporte . conflict has been resolved

modules/sievefilters/modules.php
```diff
@@ -34,7 +34,7 @@ public function process() {
             $this->out('conditions', json_encode(base64_decode($base64_obj)));
             $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[2]);
             $this->out('actions', json_encode(base64_decode($base64_obj)));
-            if (mb_strstr($script, 'allof')) {
+            if (strstr($script, 'allof')) {
```
marclaporte, line 37: Is this intentional?

I am wondering because of https://github.com/cypht-org/cypht/issues/1224

modules/sievefilters/modules.php
```diff
@@ -223,7 +263,7 @@ public function process() {
             $blocked_wildcard = '@'.$domain;
             $new_blocked_list = [];
             foreach ($blocked_list as $idx => $blocked_sender) {
-                if (!mb_strstr($blocked_sender, $blocked_wildcard)) {
+                if (!strstr($blocked_sender, $blocked_wildcard)) {
```
marclaporte, line 266: Is this intentional?

I am wondering because of https://github.com/cypht-org/cypht/issues/1224

modules/sievefilters/modules.php
```diff
@@ -635,7 +675,7 @@ public function process() {
         $script_name = generate_filter_name($this->request->post['sieve_filter_name'], $priority);
         $conditions = json_decode($this->request->post['conditions_json']);
         $actions = json_decode($this->request->post['actions_json']);
-        $test_type = mb_strtolower($this->request->post['filter_test_type']);
+        $test_type = strtolower($this->request->post['filter_test_type']);
```
marclaporte, line 678: Is this intentional?

I am wondering because of https://github.com/cypht-org/cypht/issues/1224

composer.json
```diff
@@ -46,7 +46,7 @@
         "ext-openssl": "*",
         "ext-session": "*",
         "ezyang/htmlpurifier": "^4.17",
-        "henrique-borba/php-sieve-manager": "^1.0",
+        "henrique-borba/php-sieve-manager": "~1.0.4",
```
kroky, line 49: We don't want such a hard constraint to this lib. It will prevent upgrades of the lib for non-major versions as 1.1.0.

modules/sievefilters/modules.php
kroky: We should keep multibyte functions in 99% of the places. There was only one exception when the size of a string was specified in bytes in the IMAP protocol (literals).

modules/sievefilters/site.js
```diff
@@ -863,7 +850,22 @@ $(function () {
...
+         * Toggle Filter
+         */
+        $('.toggle_filter').on('change', function () {
+                Hm_Ajax.request(
+                    [   {'name': 'hm_ajax_hook', 'value': 'ajax_sieve_toggle_script_state'},
+                        {'name': 'imap_account', 'value': $(this).attr('imap_account')},
+                        {'name': 'script_state', 'value': $(this).prop('checked')},
+                        {'name': 'sieve_script_name', 'value': $(this).attr('script_name')}],
+                    function(res) {
+                        if (res.router_user_msgs[0].startsWith('ERR')) {
```
kroky, line 863: Please return proper json to check if it is successful operation or not. Router error messages are meant to display notices and errors and might not contain the error in the first place.

modules/sievefilters/modules.php
```diff
@@ -909,53 +949,51 @@ public function process() {
...
-                );
-                $custom_condition->addAction(
-                    new \PhpSieveManager\Filters\Actions\KeepFilterAction()
+                    new \PhpSieveManager\Filters\Actions\RedirectFilterAction([$action->value])
                 );
             }
             if ($action->action == 'flag') {
                 $custom_condition->addAction(
-                    new \PhpSieveManager\Filters\Actions\FlagFilterAction(['flags' => [$action->value]])
+                    new \PhpSieveManager\Filters\Actions\FlagFilterAction([$action->value])
```
kroky, line 957: Why do you change all these FilterAction calls to constructors? I believe php-sieve-manager still expects the params in an associative array, doesn't it?

modules/sievefilters/modules.php
amaninyumu1: no it was not intentional. I understand, I correct

modules/sievefilters/modules.php
amaninyumu1: no it was not intentional. I understand, I correct

modules/sievefilters/modules.php
amaninyumu1: yes

modules/sievefilters/site.js
amaninyumu1: hello @kroky  I think I have made the fixes please review

kroky: You still check the router messages. Instead, return a json status from the method and inspect it e.g. `res.status == 'OK'` or something similar. These errors are already displayed by the ajaxComplete or similar hooks of our ajax system, so you shouldn't alert them. See how other ajax methods deal with JSON data returned by the server.

amaninyumu1: hello @kroky 
if (res.status !== 'OK' || res.success !== true) {
  checkbox.prop('checked', !checkbox.prop('checked'));
   }
   I added this condition for Status Check and Success in the JSON response

kroky: That's good but I don't see how you return this status OK structure from an output module...

amaninyumu1:
Hello @kroky 
I added ` echo json_encode([...]);` to send a json response to ajax

modules/sievefilters/modules.php
```diff
@@ -160,6 +160,62 @@ public function process() {
...
+                }
+            }
+
+            $scripts = $client->listScripts();
+            $main_script = generate_main_script($scripts);
+            save_main_script($client, $main_script, $scripts);
+            $client->activateScript('main_script');
+            $client->close();
+
+            echo json_encode([
```
kroky, line 204: Please never directly echo content in the middle of the script execution - it might have severe side effects. Also, that's a handler module, it is not meant to output data. Instead, use code like $this->out('status', 'OK'); to pass these variables to the output modules. Internally, Cypht detects that it is an ajax request and returns JSON ajax response where you will see res.status == 'OK'.

amaninyumu1: I understand thank you. I correct

amaninyumu1: hello @kroky  I just modified please review

kroky: Looks good now. Please address the other feedback items as well, so we can merge.

modules/sievefilters/modules.php
Shadow243: @kroky  means that `php-sieve-manager` expects to receive associative arrays. with the changes you have applied, it will break sieve.

modules/sievefilters/site.js
```diff
@@ -130,13 +130,6 @@ var hm_sieve_possible_actions = function() {
             type: 'string',
             extra_field: false
         },
-        {
```
Shadow243, line 133: we are using this filter at `https://github.com/cypht-org/cypht/blob/master/modules/sievefilters/modules.php#L915` 

I mean I don't understand why you are deleting this here.

modules/sievefilters/site.js
```diff
@@ -539,12 +532,6 @@ $(function () {
             $('#stop_filtering').prop('checked', false);
             current_account = $(this).attr('account');
             edit_filter_modal.open();
-
-            // Reset the form fields when opening the modal
-            $(".modal_sieve_filter_name").val('');
```
Shadow243, line 544: By deleting this code, the form starts to keep the previous data again because it is this code which resets when we click on add again.

modules/sievefilters/modules.php
```diff
@@ -1179,9 +1216,6 @@ protected function output() {
         $res .= get_script_modal_content();
         $res .= '<div class="p-3">';
         foreach($mailboxes as $idx => $mailbox) {
-            if (empty($mailbox['sieve_config_host'])) {
```
Shadow243, line 1182: This condition remains important, isn't it?

Shadow243 (changes requested):
It seems to me that you did a rebase which removed some important behavior. Can you review this PR carefully?

modules/sievefilters/site.js
amaninyumu1: From what I understand, the value \$action->value passed into the addAction method is not explicitly an associative array in the current implementation. Could you please explain to me how this change might "break" sieve's behavior? It would help me better understand the issue and ensure that the implementation meets the expectations of PhpSieveManager. thanks

modules/sievefilters/modules.php
amaninyumu1: Yes

modules/sievefilters/modules.php
Shadow243: https://github.com/cypht-org/cypht/pull/1093

kroky:
Looks good now, thanks!

Issue #991, opened 2024-05-01
Deleting message from inline view shows multiple "message deleted" notifications


PaulTGG:
## 🐛 Bugreport
Deleting message with the "delete" link while viewing the message inline shows multiple "message deleted" notifications.
![message deleted](https://github.com/cypht-org/cypht/assets/12187131/a37af50f-47fa-4088-9f2c-b8074962e096)

This doesn't happen when deleting a message from single-message view with the same link, and it doesn't happen from the mailbox view when using the checkboxes and the top delete button.

### Version & Environment

Rev: 492f634

OS:  Debian 12.5

marclaporte:
@aggestor Please show us your skills :-)

PaulTGG:
Not sure if it's related, but if you use the checkbox + delete button at the top of the page to delete a message that you're viewing inline, the message stays on screen until you open another item, or close it.  (The message line in the list of messages disappears, but the message itself does not automatically close.)  If it's not related, I'll open another but report for it later.

Baraka24:
Hello @PaulTGG, Please is your instance up to date? My side I can not even see the dele message as we could..... You can see from: 

https://github.com/cypht-org/cypht/assets/81784141/b9fc0534-d823-4393-afdc-59ccf8e3ccac

Just confirm us that you are using the latest code on master or no. If no, try to put it up to date and test again.... If you get the same result as on this short video please tell us.... we'll likely to fix so the message will be visible. Thanks.

marclaporte:
> Just confirm us that you are using the latest code on master or no.

He indicated above the revision he used.

Baraka24:
> > Just confirm us that you are using the latest code on master or no.
> 
> He indicated above the revision he used.

Eeeh! well. I can see: 
> Not sure if it's related, but if you use the checkbox + delete button at the top of the page to delete a message that you're viewing inline, the message stays on screen until you open another item, or close it. (The message line in the list of messages disappears, but the message itself does not automatically close.) If it's not related, I'll open another but report for it later.

**Please @PaulTGG can you provide us some more details like a short video for example, I am not getting you well.**

PaulTGG:
@Baraka24 Video is here!: https://dai.ly/kwLnThhImAx4wNAAml6

PaulTGG:
@Baraka24 Were you able to view the video?

> @Baraka24 Video is here!: https://dai.ly/kwLnThhImAx4wNAAml6
> 
[...]

PR #1106, opened 2024-06-30
[FIX]Displaying multiple messages with the same content for a given a…

christer77:
Displaying multiple messages with the same content for a given action

Here are 2 cases that this PR solves:
1. First case: After sending a message, and you have 2 or 3 tabs open, you keep that the message indicating that "the message is sent" is displayed several times on the other tabs open.

Ilistration:
01-first-case

3. 2nd case: When deleting a message in inline mode, we have the message "Message deleted" which also comes up as many times.
Ilistration:
02-second-case

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -257,7 +257,7 @@ var Hm_Ajax_Request = function() { return {
                 Hm_Ajax.err_condition = false;
                 Hm_Notices.hide(true);
             }
-            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {
+            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {             
                 Hm_Notices.show(res.router_user_msgs);
             }
             if (res.folder_status) {
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -280,9 +280,6 @@ class Hm_Handler_imap_process_move extends Hm_Handler_Module {
             elseif (count($moved) == 0) {
                 Hm_Msgs::add('ERRUnable to move/copy selected messages');
             }
-            if ($form['imap_move_action'] == 'move' && $form['imap_move_page'] == 'message') {
-                $this->save_hm_msgs();
-            }
             $this->out('move_count', $moved);
         }
     }
@@ -875,7 +872,6 @@ class Hm_Handler_imap_delete_message extends Hm_Handler_Module {
                 Hm_Msgs::add('Message deleted');
                 $this->out('imap_delete_error', false);
             }
-            $this->save_hm_msgs();
         }
     }
 }
@@ -940,7 +936,6 @@ class Hm_Handler_imap_archive_message extends Hm_Handler_Module {
                 Hm_Msgs::add('ERRAn error occurred archiving the message');
             }
         }
-        $this->save_hm_msgs();
     }
 }
 
@@ -1020,7 +1015,6 @@ class Hm_Handler_imap_snooze_message extends Hm_Handler_Module {
             $msg = 'ERRFailed to snooze selected messages';
         }
         Hm_Msgs::add($msg);
-        $this->save_hm_msgs();
     }
 }
 
--- a/modules/nux/modules.php
+++ b/modules/nux/modules.php
@@ -196,7 +196,6 @@ class Hm_Handler_process_nux_add_service extends Hm_Handler_Module {
                     $this->session->record_unsaved('SMTP server added');
                     $this->session->secure_cookie($this->request, 'hm_reload_folders', '1');
                     Hm_Msgs::add('E-mail account successfully added');
-                    $this->save_hm_msgs();
                     $this->session->close_early();
                     $this->out('nux_account_added', true);
                     if ($this->module_is_supported('imap_folders')) {


PaulTGG:
#991

modules/core/site.js
```diff
@@ -257,8 +257,12 @@ var Hm_Ajax_Request = function() { return {
                 Hm_Ajax.err_condition = false;
                 Hm_Notices.hide(true);
             }
-            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {
-                Hm_Notices.show(res.router_user_msgs);
+            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {             
+                var hm_msg = Hm_Utils.json_encode(res.router_user_msgs);
+                if (! Hm_Notices.displayed_messages.includes(hm_msg)) {
```
kroky, line 262: I am not sure this is the proper fix. I suspect that multiple jquery events are stacking and executing together in order to receive several identical messages. This fix will hide the duplicates but not deal with other issues like submitting several requests to the server for one and the same thing - while delete might not cause issues, other repeatable operations like send might cause. Thus, we should check why do we receive several identical messages and fix the corresponding code - if we add the same event handlers without the page being reloaded we should first remove the old one before adding or don't add the handler at all.

christer77: By the way, I found that messages are saved in a table in session. What JavaScript does is retrieve messages in session and display them.

 A solution came to mind, and if for each message we added an identifier, this could allow us to delete it in the message session table. What do you think ?

kroky: This doesn't explain why do you see several "Message deleted" messages. Again, we don't want to make the messages disappear by checking their ID or similarity. We want to relay the messages coming from the server to the client but we need to make sure that no duplicate ajax calls are executed, even when multiple tabs are open. Please specify here an exact use-case that reproduced the problem and I can check if you can't find the reason for the duplicate messages.

christer77: Here's how I figured out the bug:

Sending the request to delete the message: **_ajax_imap_delete_message_**(1)
    The time the request is being processed;
    The application sends other requests including:
    - **_ajax_imap_unsnooze_**(2)
    - **_ajax_imap_folder_display_**(3)
    - **_ajax_imap_unread_**(4)
    
Queries **_(2),(3), (4)_** run every second.
So far request _**(1)**_ has still not been successful, therefore, _**(2, 3 and 4)**_ as well.

When request _**(1)**_ is successful, the message "**_Message deleted_**" is inserted into the message session.

This is where the other requests respond, and all of them come back with the message array that contains the message "**_Message deleted_**".

The short video attached illustrates these explanations.
Message deleted

christer77: > Here's how I figured out the bug:
> 
> Sending the request to delete the message: **_ajax_imap_delete_message_**(1) The time the request is being processed; The application sends other requests including: - **_ajax_imap_unsnooze_**(2) - **_ajax_imap_folder_display_**(3) - **_ajax_imap_unread_**(4)
> 
> Queries **_(2),(3), (4)_** run every second. So far request _**(1)**_ has still not been successful, therefore, _**(2, 3 and 4)**_ as well.
> 
> When request _**(1)**_ is successful, the message "**_Message deleted_**" is inserted into the message session.
> 
> This is where the other requests respond, and all of them come back with the message array that contains the message "**_Message deleted_**".
> 
> The short video attached illustrates these explanations. Message deleted [ Message deleted ](https://private-user-images.githubusercontent.com/62720246/359032876-dd7ed79f-296d-4850-9d68-d34d18f0c565.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ2NzQxNDcsIm5iZiI6MTcyNDY3Mzg0NywicGF0aCI6Ii82MjcyMDI0Ni8zNTkwMzI4NzYtZGQ3ZWQ3OWYtMjk2ZC00ODUwLTlkNjgtZDM0ZDE4ZjBjNTY1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODI2VDEyMDQwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM2NjZhZDQzZWYxMGVlMDFjMGZmOTI5ZDRlN2E3ZWI4ZGRiZDllNzEzZjY3MmZjYWU0ZWE1N2VmMzQ5MDhmNzAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.n4SMg82TIm4Qgk-_R1Uksi3gVTEM4E1KiE6w3dqnw-o) [ ](https://private-user-images.githubusercontent.com/62720246/359032876-dd7ed79f-296d-4850-9d68-d34d18f0c565.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjQ2NzQxNDcsIm5iZiI6MTcyNDY3Mzg0NywicGF0aCI6Ii82MjcyMDI0Ni8zNTkwMzI4NzYtZGQ3ZWQ3OWYtMjk2ZC00ODUwLTlkNjgtZDM0ZDE4ZjBjNTY1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MjYlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODI2VDEyMDQwN1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTM2NjZhZDQzZWYxMGVlMDFjMGZmOTI5ZDRlN2E3ZWI4ZGRiZDllNzEzZjY3MmZjYWU0ZWE1N2VmMzQ5MDhmNzAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.n4SMg82TIm4Qgk-_R1Uksi3gVTEM4E1KiE6w3dqnw-o)

Hello @kroky , did you manage to reproduce this?

kroky: Then, I think the problem is $this->save_hm_msgs(); at the end of imap_delete_message handler module - we shouldn't store these in the session for subsequent requests to show since we are no reloading the page after deletion and ajax Hm_Ajax_request.done method successfully displays the router messages without reload.

PaulTGG:
👀

modules/core/site.js
kroky: I think you misunderstood me. We shouldn't remove the save_hm_msgs functionality altogether - it is necessary when we have a redirect - otherwise, the message will be lost. I meant we should remove the call in places where we don't have a redirect - e.g. after message deletion.

kroky: @christer77 can you please ping me when you do changes here as this PR is pretty old and we want to move forward with it?

christer77: > I think you misunderstood me. We shouldn't remove the save_hm_msgs functionality altogether - it is necessary when we have a redirect - otherwise, the message will be lost. I meant we should remove the call in places where we don't have a redirect - e.g. after message deletion.

well noted

christer77: But if we delete it as you say, how are we going to know that the action has succeeded, for example in the case of deleting the message as you said.

kroky: Hm_Ajax_request.done method checks for router messages coming from ajax requests and shows them. So, unless there is a redirect, we should be already fine displaying those messages.

modules/core/site.js
```diff
@@ -257,7 +257,8 @@ var Hm_Ajax_Request = function() { return {
                 Hm_Ajax.err_condition = false;
                 Hm_Notices.hide(true);
             }
-            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {
+            if (res.router_user_msgs && !$.isEmptyObject(res.router_user_msgs)) {             
+                var hm_msg = Hm_Utils.json_encode(res.router_user_msgs);
```
kroky, line 261: Why do you assign this to a variable and not use it? Can you remove if that's debugging code?
Otherwise, the PR looks good, we can merge after this is fixed.

christer77: Noted

Issue #1105, opened 2024-06-30
Activating Sieve should be easier


marclaporte:
As of now, activating Sieve is more complicated than in should be: https://github.com/cypht-org/cypht-website/pull/67/files

By default, Sieve should be available in a standard Cypht installation. Just activate with 1 checkbox.

Baraka24:
> Hey guys, has this feature been implemented already?

Hey, for a fresh installation of Cypht, you wil have to:
1. go to settings
2. Site
3. General
4. Check Enable sieve filter
Enjoy filters in Cypht.

PR #1109, opened 2024-07-03
[FIX]Sieve filters: Activating Sieve should be easier

Baraka24:
[Related issue](https://github.com/cypht-org/cypht/issues/1105#event-13342173797)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -191,7 +191,7 @@ WORDPRESS_CLIENT_URI=
 RECAPTCHA_SECRET=
 RECAPTCHA_SITE_KEY=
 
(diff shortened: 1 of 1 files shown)


modules/sievefilters/modules.php
```diff
@@ -1274,7 +1274,7 @@ protected function output() {
 class Hm_Handler_process_enable_sieve_filter_setting extends Hm_Handler_Module {
     public function process() {
         function sieve_enabled_callback($val) { return $val; }
-        process_site_setting('enable_sieve_filter', $this, 'sieve_enabled_callback', true, true);
+        process_site_setting('enable_sieve_filter', $this, 'sieve_enabled_callback', false, true);
```
Shadow243, line 1277: The default value to use if callback is not submitted should be true not false here.

Shadow243 (changes requested):
Apart from that it looks good for me.

PR #1124, opened 2024-07-18
[FIX]compose&reply_all: Remove Selected Email or Sender Email from Recipients List While Using 'Reply All'

Baraka24:
This PR enhances the functionality of the email composer by ensuring that the selected email or sender email is removed from the recipients list when using the "Reply All" feature.


Shadow243:
Looks good for me.

PR #1125, opened 2024-07-19
Correct pagination while displaying contacts

Shadow243:
Paging affects groups.

Ex: if in trusted Contacts there are 100 contacts and in Collected Contacts we have 2 contacts, the pagination will ensure that the 2 contacts in Collected are not displayed.

In this PR we group first then we paginate what is already grouped

### DIFF
--- a/modules/contacts/hm-contacts.php
+++ b/modules/contacts/hm-contacts.php
@@ -139,13 +139,65 @@ class Hm_Contact_Store {
         $this->data = array();
     }
 
-    public function page($page, $size) {
+    public function page($page, $size, $data = null) {
         if ($page < 1) {
             return array();
         }
-        return array_slice($this->data, (($page - 1)*$size), $size, true);
+        if ($data === null) {
+            $data = $this->data;
+        }
+        return array_slice($data, (($page - 1)*$size), $size, true);
     }
 
+    public function group_by($column = 'group') {
+        if (!is_string($column) && !is_int($column) && !is_float($column)) {
+            trigger_error('group_by(): The key should be a string, an integer, or a float', E_USER_ERROR);
+            return null;
+        }
+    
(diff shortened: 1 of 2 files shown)


Shadow243:
There is an other issue reported by @Baraka24  and which concerns this PR.

modules/contacts/hm-contacts.php
```diff
@@ -142,13 +142,66 @@ public function reset() {
...
+        $grouped = [];
+        foreach ($this->data as $value) {
+            
+            $column = null;
+            if (is_callable($func)) {
+                $column = call_user_func($func, $value);
+            } elseif (is_object($value)) {
+                $reflection = new ReflectionClass($value);
+                $property = $reflection->getProperty('data');
+                $property->setAccessible(true); // Make the private property accessible
```
kroky, line 172: There is a reason why private properties are private and using reflection to bypass this is something nasty. If this is a contact object isn't $value->value($_key) enough to get you what you want?

modules/contacts/hm-contacts.php
```diff
@@ -142,13 +142,66 @@ public function reset() {
...
             return array();
         }
-        return array_slice($this->data, (($page - 1)*$size), $size, true);
+        if ($data === null) {
+            $data = $this->data;
+        }
+        return array_slice($data, (($page - 1)*$size), $size, true);
     }
 
+    public function group_by($column = 'group') {
```
kroky, line 155: Can you refactor the code to group by a string version of the column name? I don't see a reason we group by a callable or support cases for array or object contacts. You can refactor calling code to ensure we are working with a standard set of data - either array of arrays or array of contact objects.

kroky:
Ah, yes, I didn't notice you pushed the updates. Looks good!

PR #1128, opened 2024-07-22
[FIX] I fixed a PHP Warning  added a condition to check if ->get('rec…

amaninyumu1:


Shadow243:
@amaninyumu1  Can you describe the scenario please !

amaninyumu1:
Hello @Shadow243 
The problem was that we were trying to use $this->get('recaptcha_config') without being able to check if it exists. So I added a condition to check if it exists and then use it
return '<div class="g-recaptcha" data-sitekey="'.$this->html_safe($this->get('recaptcha_config')['site_key']).'"></div>';

Shadow243 (changes requested):
According to https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/setup.php#L8 `$this->get('recaptcha_config')` will never be undefined, did you get an issue about it ?

The file https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/recaptcha.ini should be removed as well.

amaninyumu1:
Yes but in this scenario the error was that the "recaptcha_config" key is null so it was impossible to access array offset on value of type null which is why I checked if it exists. In case it exists then it is not null

Shadow243:
> The file https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/recaptcha.ini should be removed as well.

Remove this please

amaninyumu1:
I deleted this file https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/recaptcha.ini

Shadow243:
> I deleted this file https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/recaptcha.ini

Delete it from the projet not form the PR.

modules/recaptcha/modules.php
```diff
@@ -54,9 +54,8 @@ protected function output() {
  */
 class Hm_Output_recaptcha_form extends Hm_Output_Module {
     protected function output() {
-        if (!$this->get('router_login_state')) {
-            $rconf = $this->get('recaptcha_config');
-            return '<div class="g-recaptcha" data-sitekey="'.$this->html_safe($rconf['site_key']).'"></div>';
+        if (!$this->get('router_login_state') && $this->get('recaptcha_config')) {
```
Shadow243, line 57: use the condition like it's used here: https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/modules.php#L16C9-L17C55

you can put the same condition at this level too: Hm_Output_recaptcha_script(https://github.com/cypht-org/cypht/blob/master/modules/recaptcha/modules.php#L46)

modules/recaptcha/modules.php
```diff
@@ -54,9 +54,8 @@ protected function output() {
...
+            return '<div class="g-recaptcha" data-sitekey="'.$this->html_safe($this->get('recaptcha_config')['site_key']).'"></div>';
```
amaninyumu1, line 58: this error comes back when I remove the condition that I put
[Sun Jul 21 11:18:55.635757 2024] [php:warn] [pid 1948:tid 1952] [client ::1:7190] PHP Warning:  Trying to access array offset on value of type null in C:\\xampp\\htdocs\\my_cypht\\modules\\recaptcha\\modules.php on line 59

amaninyumu1:
if (!$this->get('router_login_state')) {
            $rconf = $this->get('recaptcha_config');
            if (is_array($rconf) && array_key_exists('site_key', $rconf) && $rconf['site_key']) {
                return '<div class="g-recaptcha" data-sitekey="'.$this->html_safe($this->get('recaptcha_config')['site_key']).'"></div>';
            }
        }

Shadow243:
I just added the same condition at the recaptcha script level to prevent the js file from being reloaded: `https://www.google.com/recaptcha/api.js` when we don't need it.

You can test to see if there will be an error (in console and in logs)

amaninyumu1:
yes it works

Issue #1113, opened 2024-07-09
Screen emails (messages from previously uncategorized emails are flagged)


marclaporte:
A new email reaches to us. It could be spam, a phishing attempt, a great collaboration opportunity, etc.

Let's add a way to identify / filter / temporarily hide emails from senders that are not in our various lists (blocked, in address book, collected, trusted, etc.)

Context:
- https://www.hey.com/features/the-screener/
- https://www.fastmail.help/hc/en-us/articles/8661280554511-Screening-emails

And please factor in the "reply to" as per https://github.com/cypht-org/cypht/issues/1115

Related Cypht code or issues:
- https://github.com/cypht-org/cypht/pull/816
- https://github.com/cypht-org/cypht/pull/721

PR #1132, opened 2024-07-25
implement-screen-email

christer77:
Fixed [Screen emails (messages from previously uncategorized emails are flagged)](https://github.com/cypht-org/cypht/issues/1113)

### DIFF
--- a/lib/repository.php
+++ b/lib/repository.php
@@ -55,7 +55,11 @@ trait Hm_Repository {
 
     public static function edit($id, $entity) {
         if (array_key_exists($id, self::$entities)) {
-            self::$entities[$id] = array_merge(self::$entities[$id], $entity);
+            if (is_array($entity)) {
+                self::$entities[$id] = array_merge(self::$entities[$id], $entity);
+            } else {
+                self::$entities[$id] = $entity;
+            }
             self::save();
             return true;
         }
--- a/modules/contacts/modules.php
+++ b/modules/contacts/modules.php
@@ -400,6 +400,56 @@ class Hm_Output_contacts_list extends Hm_Output_Module {
 }
 
 /**
+ * @subpackage contacts/handler
+ */
+class Hm_Handler_save_contact extends Hm_Handler_Module
+{
+    public function process()
+    {
+        list($success, $form ) = $this->process_form(array('email_address'));
+        if ($success) {
+            $contacts = $this->get('contact_store');
+            $contact_list = $contacts->getAll();
+            $existingEmails = array_map(function($contact){
+                return $contact->value('email_address');
+            },$contact_list);
+
+            $list_mails = array_unique(explode(",", $form['email_address']));
+
+            foreach ($list_mails as $addr) {
+                $addresses = process_address_fld($addr);
+                $newEmails = array_column($addresses, 'email');
+                if (!empty($newEmails)) {
+                    $newContacts = array_filter($newEmails, function ($email) use ($existingEmails) {
+                        return !in_array($email, $existingEmails);
+                    });
+                    $existingContacts = array_filter($existingEmails, function ($email) use ($newEmails) {
+                        return in_array($email, $newEmails);
+                    });
+                    if (!empty($newContacts)) {
+                        $newContacts = array_map(function ($email) {
+                            return ['source' => 'local', 'email_address' => $email, 'display_name' => $email, 'group' => 'Trusted Senders'];
+                        }, $newContacts);
+                        $contacts->add_contact($newContacts[0]);
+                    }
+                    if (!empty($existingContacts)) {
+                        $existingContacts = array_map(function ($email) {
+                            return ['source' => 'local', 'email_address' => $email, 'group' => 'Trusted Senders'];
+                        }, $existingContacts);
+                        foreach ($existingContacts as $key => $contact) {
+                            $contacts->update_contact($key, $contact);
+                        }
+                    }
+                }
+            }
+            $this->session->record_unsaved('Contacts added to  Trusted Contacts list');
+            Hm_Msgs::add('Contacts added to  Trusted Contacts list');
+        }
+
+    }
+}
+
+/**
  * @subpackage contacts/output
  */
 class Hm_Output_filter_autocomplete_list extends Hm_Output_Module {
--- a/modules/contacts/setup.php
+++ b/modules/contacts/setup.php
@@ -23,6 +23,8 @@ add_handler('compose', 'process_send_to_contact', true, 'contacts', 'save_user_d
 add_handler('compose', 'load_contacts', true, 'contacts', 'process_compose_form_submit', 'after');
 add_handler('compose', 'store_contact_message', true, 'contacts', 'load_contacts', 'after');
 
+add_handler('ajax_imap_folder_display', 'load_contacts', true, 'contacts', 'load_user_data', 'after');
+
 add_handler('ajax_imap_message_content', 'load_contacts', true, 'contacts', 'load_user_data', 'after');
 add_handler('ajax_imap_message_content', 'find_message_contacts', true, 'contacts', 'imap_message_content', 'after');
 add_output('ajax_imap_message_content', 'add_message_contacts', true, 'contacts', 'filter_message_headers', 'after');
@@ -31,6 +33,7 @@ add_handler('ajax_imap_message_content', 'store_contact_allow_images', true, 'co
 setup_base_ajax_page('ajax_add_contact', 'core');
 add_handler('ajax_add_contact', 'load_contacts', true, 'contacts', 'load_user_data', 'after');
 add_handler('ajax_add_contact', 'save_user_data', true, 'core', 'language', 'after');
+add_handler('ajax_add_contact', 'save_contact',  true);
 
 
 setup_base_ajax_page('ajax_autocomplete_contact', 'core');
@@ -81,8 +84,8 @@ return array(
         'contact_source' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
         'contact_type' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
         'contact_auto_collect' => FILTER_VALIDATE_BOOLEAN,
-        'enable_warn_contacts_cc_not_exist_in_list_contact' => FILTER_VALIDATE_INT
-
+        'enable_warn_contacts_cc_not_exist_in_list_contact' => FILTER_VALIDATE_INT,
+        'email_address' => FILTER_SANITIZE_FULL_SPECIAL_CHARS
     ),
     'allowed_get' => array(
         'contact_id' => FILTER_SANITIZE_FULL_SPECIAL_CHARS,
--- a/modules/core/message_list_functions.php
+++ b/modules/core/message_list_functions.php
@@ -206,7 +206,18 @@ function message_list_row($values, $id, $style, $output_mod, $row_class='') {
     if ($row_class) {
         $res .= ' '.$output_mod->html_safe($row_class);
     }
-    $res .= '">';
+    $data_uid = "";
+    if ($uids = explode("_", $id)) {
+        if (isset($uids[2])) {
+            $data_uid = 'data-uid="'. $uids[2] .'"';
+        }
+    }
+    if (!empty($data_uid)) {
+        $res .= '" '.$data_uid.'>';
+    } else {
+        $res .= '">';
+    }
+    
     if ($style == 'news') {
         $res .= '<td class="news_cell checkbox_cell">';
     }
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -857,7 +857,7 @@ function imap_refresh_oauth2_token($server, $config) {
  * @return int count of messages moved
  */
 if (!hm_exists('imap_move_same_server')) {
-function imap_move_same_server($ids, $action, $hm_cache, $dest_path) {
+function imap_move_same_server($ids, $action, $hm_cache, $dest_path, $screen_emails=false) {
     $moved = array();
     $keys = array_keys($ids);
     $server_id = array_pop($keys);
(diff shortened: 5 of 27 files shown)


marclaporte:
I sure look forward to this feature :-)

modules/imap/functions.php
```diff
@@ -859,19 +859,31 @@ function imap_refresh_oauth2_token($server, $config) {
  * @return int count of messages moved
  */
 if (!hm_exists('imap_move_same_server')) {
-function imap_move_same_server($ids, $action, $hm_cache, $dest_path) {
+    function imap_move_same_server($ids, $action, $hm_cache, $dest_path, $screen_emails=false) {
```
Shadow243, line 862: correct the indentation

modules/imap/hm-imap.php
```diff
@@ -2140,7 +2140,7 @@ public function get_first_message_part($uid, $type, $subtype=false, $struct=fals
          * @return array list of headers
          */
 
-        public function get_mailbox_page($mailbox, $sort, $rev, $filter, $offset=0, $limit=0, $keyword=false) {
+         public function get_mailbox_page($mailbox, $sort, $rev, $filter, $offset=0, $limit=0, $keyword=false, $trusted_senders=array()) {
```
Shadow243, line 2143: there is a  space before`public' but it shouldn't

modules/contacts/modules.php
```diff
@@ -397,6 +397,45 @@ protected function output() {
...
+                    if (!empty($newContacts)) {
+                        $newContacts = array_map(function ($email) {
+                            return ['source' => 'local', 'email_address' => $email, 'display_name' => $email, 'group' => 'Trusted Senders'];
+                        }, $newContacts);
+                        $contacts->add_contact($newContacts[0]);
+                    }
+                }
+            }
+            $this->session->record_unsaved('Contact.s added to  Trusted Contacts list');
+            Hm_Msgs::add('Contact.s added to  Trusted Contacts list');
```
Shadow243, line 433: is it `Contact.s` or `Contacts` ?

modules/contacts/setup.php
```diff
@@ -31,7 +33,7 @@
 setup_base_ajax_page('ajax_add_contact', 'core');
 add_handler('ajax_add_contact', 'load_contacts', true, 'contacts', 'load_user_data', 'after');
 add_handler('ajax_add_contact', 'save_user_data', true, 'core', 'language', 'after');
-
```
Shadow243, line 34: remove this unnecessary change

modules/imap/output_modules.php
```diff
@@ -1383,3 +1392,24 @@ protected function output() {
...
+/**
+ * Option to set the per page count for IMAP folder views
+ * @subpackage imap/output
+ */
+class Hm_Output_first_time_screen_emails_per_page_setting extends Hm_Output_Module {
+    protected function output() {
+        $settings = $this->get('user_settings', array());
+        $per_page = 20;
+        $reset = '';
+        if (array_key_exists('first_time_screen_emails', $settings)) {
```
Shadow243, line 1405: I'm getting this by default:
Screenshot 2024-07-31 at 17 23 48

rewrite the condition like this:
```
if (array_key_exists('first_time_screen_emails', $settings) && $settings['first_time_screen_emails']) {
            $per_page = $settings['first_time_screen_emails'];
}
```
to resolve and make it work correctly.

You will just add `&& $settings['first_time_screen_emails']` in your condition.

Shadow243 (changes requested):
the PR seems correct Thanks, just some small correction to be done

modules/imap/output_modules.php
Shadow243: @christer77 Will you have time to fix minor issues ?

christer77: Noted

modules/core/message_list_functions.php
```diff
@@ -206,7 +206,18 @@ function message_list_row($values, $id, $style, $output_mod, $row_class='') {
     if ($row_class) {
         $res .= ' '.$output_mod->html_safe($row_class);
     }
-    $res .= '">';
+    $data_uid = "";
+    if ($uids = explode("_", $output_mod->html_safe($id))) {
```
kroky, line 210: No need to html_safe this when you are using it in PHP context - like passing it to explode function. Use html_safe when you append output to a template as html content.

modules/contacts/modules.php
```diff
@@ -397,6 +397,45 @@ protected function output() {
...
+
+            $list_mails = array_unique(explode(",", $form['email_address']));
+
+            foreach ($list_mails as $addr) {
+                $addresses = process_address_fld($addr);
+                $newEmails = array_column($addresses, 'email');
+
+                if (!empty($newEmails)) {
+                    $newContacts = array_filter($newEmails, function ($email) use ($existingEmails) {
+                        return !in_array($email, $existingEmails);
```
kroky, line 421: What if you want to add an existing contact not present in Trusted Senders group as trusted sender? Shouldn't it edit the contact and assign to Trusted Senders group?

modules/imap/functions.php
```diff
@@ -857,19 +857,31 @@ function imap_refresh_oauth2_token($server, $config) {
...
     $imap = Hm_IMAP_List::connect($server_id, $cache);
     foreach ($ids[$server_id] as $folder => $msgs) {
         if (imap_authed($imap) && $imap->select_mailbox(hex2bin($folder))) {
-            if ($imap->message_action(mb_strtoupper($action), $msgs, hex2bin($dest_path[2]))) {
+            if ($screen_emails) {
                 foreach ($msgs as $msg) {
                     $moved[]  = sprintf('imap_%s_%s_%s', $server_id, $msg, $folder);
+                    $email = current(array_column(process_address_fld($imap->get_message_headers($msg)['From']), "email"));
+                    $uid = $imap->search('ALL', false, array(array('FROM', $email)));
+                    if ($imap->message_action(mb_strtoupper($action), $uid, hex2bin($dest_path[2]))) {
```
kroky, line 873: What is the purpose of this block? Move all emails from this sender to the destination folder? Is this the intended behavior?

modules/imap/functions.php
```diff
@@ -857,19 +857,31 @@ function imap_refresh_oauth2_token($server, $config) {
...
+                        $moved[]  = sprintf('imap_%s_%s_%s', $server_id, $msg, $folder);
```
kroky, line 874: I think $uid must be looped and $msg replaced with single uid values. Otherwise, you can just show a summary of the number of moved messages. In any case, it seems a bit dangerous to move all emails from a certain sender at once - there could be thousands.

modules/imap/handler_modules.php
```diff
@@ -252,18 +252,37 @@ class Hm_Handler_imap_process_move extends Hm_Handler_Module {
...
             }
             if (count($other_server_ids) > 0) {
                 $moved = array_merge($moved, imap_move_different_server($other_server_ids, $form['imap_move_action'], $dest_path, $this->cache));
 
             }
             if (count($moved) > 0 && count($moved) == count($msg_ids)) {
                 if ($form['imap_move_action'] == 'move') {
-                    Hm_Msgs::add('Messages moved');
+                    if ($screen) {
+                        Hm_Msgs::add('Email.s moved to Sreen email folder');
```
kroky, line 282: Please correct the message - dot seems wrong and also the folder is called "Screen emails" - probably worth making the name configurable by user?

modules/imap/handler_modules.php
```diff
@@ -804,7 +830,17 @@ public function process() {
             $imap = Hm_IMAP_List::connect($form['imap_server_id'], $cache);
             if (imap_authed($imap)) {
                 $this->out('imap_mailbox_page_path', $path);
-                list($total, $results) = $imap->get_mailbox_page(hex2bin($form['folder']), $sort, $rev, $filter, $offset, $limit, $keyword);
+                if (isset($this->request->get['screen_emails']) && hex2bin($form['folder']) == 'INBOX' && $this->module_is_supported("contacts")) {
+                    $contacts = $this->get('contact_store');
+                    $contact_list = $contacts->getAll();
```
kroky, line 835: Are all existing contact emails trusted by default or only those in Trusted Senders group? I am a bit confused by the usage of trusted senders group.

modules/imap/hm-imap.php
```diff
@@ -2160,6 +2160,16 @@ public function get_mailbox_page($mailbox, $sort, $rev, $filter, $offset=0, $lim
...
             }
+            if ($trusted_senders && is_array($trusted_senders)) {
+                $terms = array();
+                if ($keyword) {
+                    $terms[] = array('TEXT', $keyword);
+                }
+                foreach ($trusted_senders as $sender) {
+                    $terms[] = array('FROM', 'NOT '. $sender);
+                }
+                $uids = $this->search($filter, $uids, $terms);
```
kroky, line 2171: There should be only one search call - please combine this block with the keyword block above and execute one final search with the relevant terms.

modules/imap/functions.php
christer77: Targeted behaviour? No, but we found it interesting to do so for an address that we find not interesting.

modules/imap/functions.php
christer77: I see that fear comes back to complexity, what do you think? Should we leave this way of doing things (move all messages) or remove it?

modules/imap/functions.php
kroky: Then, is user informed that he would actually move many messages at once? Shouldn't this be a configuration or user choice somewhere in the UI?

modules/imap/functions.php
kroky: If you leave this code, please address the message suggestion I added above.

modules/imap/functions.php
christer77: > Then, is user informed that he would actually move many messages at once? Shouldn't this be a configuration or user choice somewhere in the UI?
I find this suggestion very interesting. I think that's how I'm going to do it.

modules/imap/functions.php
christer77: Noted

modules/imap/functions.php
christer77: move-message-in-screen-email

modules/imap/functions.php
christer77: > If you leave this code, please address the message suggestion I added above.

Suggestion applied

marclaporte:
@christer77 This branch has conflicts that must be resolved

modules/imap/handler_modules.php
christer77: It is for all existing contact emails(in address book, trusted, etc.)

modules/contacts/modules.php
kroky: I see what you do but my question remains - if an email coming from the form is not in contacts group - it is added with a group -> Trusted Senders. If an email IS in the contacts list, it is not edited at all - no group 'Trusted Senders' is added for that email. This sounds like a bug to me.

modules/imap/functions.php
kroky: I don't see it. moved array should contain the msg uids of the actual moved messages. When you bulk move many messages, you only add the original msg uid and not the uids of the moved messages. This is a bug.

lib/repository.php
```diff
@@ -55,9 +55,11 @@ public static function add($entity, $save = true) {
 
     public static function edit($id, $entity) {
         if (array_key_exists($id, self::$entities)) {
-            self::$entities[$id] = array_merge(self::$entities[$id], $entity);
-            self::save();
-            return true;
+            if (is_array(self::$entities[$id]) && is_array($entity)) {
```
kroky, line 58: If these are objects (e.g. Hm_Contacts), I think you should replace the object in the entities list - otherwise, the change would be discarded and not saved.

modules/imap/functions.php
```diff
@@ -857,19 +857,31 @@ function imap_refresh_oauth2_token($server, $config) {
...
     foreach ($ids[$server_id] as $folder => $msgs) {
         if (imap_authed($imap) && $imap->select_mailbox(hex2bin($folder))) {
-            if ($imap->message_action(mb_strtoupper($action), $msgs, hex2bin($dest_path[2]))) {
+            if ($screen_emails) {
                 foreach ($msgs as $msg) {
                     $moved[]  = sprintf('imap_%s_%s_%s', $server_id, $msg, $folder);
+                    $email = current(array_column(process_address_fld($imap->get_message_headers($msg)['From']), "email"));
+                    $uids = $imap->search('ALL', false, array(array('FROM', $email)));
+                    foreach ($uids as $uid) {
+                        $move = $imap->message_action(mb_strtoupper($action), $uid, hex2bin($dest_path[2]));
```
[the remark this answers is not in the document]
christer77: This is done on line 870, if we do it on line 874, we risk having an array that contains more values that are not actually expected at the javascript level. I'm going to say that in the array we put the messages that have been selected by the user, while on line 874 we move the messages that have not been selected.

lib/repository.php
christer77: It seems that I did not understand correctly

modules/imap/functions.php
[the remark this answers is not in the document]
kroky: Line 870 adds just the originally selected message to the moved list. You actually move all other messages from the same email address that are found in $uids array. These must go into moved array as well. Otherwise, what happens if user selects 1 message - it says 1 message moved but it actually moved many more from the same sender - it should report the actual number of messages that are moved.

lib/repository.php
kroky: Why do you add the is_array check? Most probably because the entity might not be an array but Hm_Content (object) in which case you should replace the object in the list. The purpose of this edit method is to update an entity in the list. If you ignore the non-array ones, you fail to deliver the purpose of the method and silently introduce bugs.

modules/imap/functions.php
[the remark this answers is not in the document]
christer77: Noted

lib/repository.php
```diff
@@ -55,9 +55,11 @@ public static function add($entity, $save = true) {
 
     public static function edit($id, $entity) {
         if (array_key_exists($id, self::$entities)) {
-            self::$entities[$id] = array_merge(self::$entities[$id], $entity);
-            self::save();
-            return true;
+            if ((is_array(self::$entities[$id]) || is_object(self::$entities[$id])) && (is_array($entity) || is_object($entity))) {
+                self::$entities[$id] = array_merge(self::$entities[$id], $entity);
```
kroky, line 59: Will array_merge work on objects? I think if it is an object, you should really replace the self::$entities[$id] one with the new object unless you are willing to do an object merge manually.

modules/imap/functions.php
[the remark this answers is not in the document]
kroky: Thanks!

lib/repository.php
christer77: I can convert Object  to Array. what do you think?

kroky: The Hm_Contact object has properties and methods, converting to array will loose some functionality on the object. I prefer if you replace the entity. Something like:
```
if ($entity instanceof Hm_Contact) {
  self::$entities[$id] = $entity;
} else {
  self::$entities[$id] = array_merge(self::$entities[$id], $entity);
}
```

christer77: I think I understand. Thanks

modules/imap/output_modules.php
```diff
@@ -1410,3 +1424,39 @@ protected function output() {
...
+    protected function output() {
+        $settings = $this->get('user_settings', array());
+        $checked = "";
+        if (array_key_exists('move_messages_in_screen_email', $settings) && $settings['move_messages_in_screen_email']) {
+            if ($settings['move_messages_in_screen_email']) {
+                $checked = "checked";
+            }
+        }
+        $res = '<tr class="general_setting"><td><label for="move_messages_in_screen_email">'.
+            $this->trans('Move all messages in Screen Email folder if dislike email').'</label></td><td><input class="form-check-input" type="checkbox" role="switch" id="move_messages_in_screen_email" name="move_messages_in_screen_email" '.$checked.' ></td></tr>';
```
marclaporte, line 1458: "Move all messages in Screen Email folder if dislike email"

Please make this clearer for a new user.

PR #1133, opened 2024-07-25
[ENH] Improve Unicode Handling in DOMDocument HTML Loading

kambereBr:
This PR addresses an issue with Unicode character handling in the reply and reply-all functionalities.

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -449,7 +449,8 @@ class HTMLToText {
 
     function __construct($html) {
         $doc = new DOMDocument();
-        $doc->loadHTML(html_entity_decode($html, ENT_QUOTES | ENT_HTML5, 'UTF-8'));
+        $html = mb_convert_encoding($html, 'HTML-ENTITIES', 'UTF-8');
+        $doc->loadHTML(htmlentities($html, ENT_QUOTES, 'UTF-8'));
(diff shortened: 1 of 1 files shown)


modules/core/message_functions.php
```diff
@@ -449,7 +449,7 @@ class HTMLToText {
 
     function __construct($html) {
         $doc = new DOMDocument();
-        $doc->loadHTML(html_entity_decode($html, ENT_QUOTES | ENT_HTML5, 'UTF-8'));
+        $doc->loadHTML(mb_convert_encoding($html, 'HTML-ENTITIES', 'UTF-8'));
```
Shadow243, line 452: use https://www.php.net/manual/en/function.htmlentities.php instead is the goal is to prevent HTML injection by converting special characters to HTML entities

Shadow243 (changes requested):
I can see here https://github.com/Shadow243/cypht/commit/51adb5202200f3058f04698eba7603cabd43a415 that it was `htmlentities`

modules/core/message_functions.php
kambereBr: Great, thank you for the guidance.

Shadow243:
@Baraka24 you reported this, can you confirm it's now fine ?

Baraka24:
@Shadow243 this fixes the issue as well. I have tested.

PR #1138, opened 2024-07-30
Remove development updates in production

josaphatim:
Removed development updates section in production


marclaporte:
This is to show the list of commits, right?

josaphatim:
> This is to show the list of commits, right?

That it. I thought I could be better to hide them

marclaporte:
Ok, let's see if anyone reacts :-)

This gave me another idea: https://github.com/cypht-org/cypht/issues/1143

PR #1139, opened 2024-07-30
Add copy param to RedirectFilterAction & correct imap server marges o…

Shadow243:
Related Task:[item111792](https://avan.tech/item111792) and related issue: [20](https://github.com/cypht-org/php-sieve-manager/issues/20)


modules/sievefilters/modules.php
```diff
@@ -1278,6 +1283,16 @@ function sieve_enabled_callback($val) { return $val; }
     }
 }
 
+/**
```
josaphatim, line 1286: @Shadow243 I think the goal is to have both redirect and forward here. I think there is no need to create a user setting for this.

Screenshot 2024-07-31 at 10 13 58

PR #1150, opened 2024-08-07
reset form on add_filter click action

Shadow243:
When you click edit filter on sieve_filters page if you cancel(close) the modal and then click `Add Filter` the form is keeping old data while editing.

In this PR we make sure on Add Filter we reset the form after the  modal is opened.


Baraka24:
Thanks for fixing this issue @Shadow243.

Issue #1149, opened 2024-08-07
"All" inbox is blank, need help


europacafe:
## 💬 Question
I added two accounts (same domain) without a problem. I can see each inbox mails. When creating those accounts, I did not check the "Hide From Combined Pages". However, when I click "All", no combined inbox mails are displayed.

![image](https://github.com/user-attachments/assets/8bde272f-7560-49ba-a7fa-38c5a02c009d)
![image](https://github.com/user-attachments/assets/a08feea9-c533-48a1-bea9-e3ae982530ce)

europacafe:
Thanks. I definitely have a lot of mails during the last 7 days. I even test sending a few new mails,
![image](https://github.com/user-attachments/assets/7847a19b-4e57-431e-8913-ca6b0f70af26)
![image](https://github.com/user-attachments/assets/5585824b-fcde-43e4-b93a-20bed8dab18a)

marclaporte:
Thank you @europacafe 

Which version number of Cypht?
Manual install? via Docker? via Yunohost? via Tiki?

Can you join us on https://gitter.im/cypht-org/community ?  We could do a screenshare session with one of the developers to pinpoint the cause.

We did a lot of massive changes for Cypht 2.0.0, make Cypht more feature-rich and easier to maintain/evolve. But we broke a few thing in the transition, and we are eager to clean them up :-)

marclaporte:
It's great that you are running Stalwart! Stalwart offers JMAP support, and Cypht now supports JMAP:
https://github.com/cypht-org/cypht/issues/180

And we are now working on Sieve-JMAP: https://github.com/cypht-org/cypht/issues/931

europacafe:
With docker image v2.2.0, this issue has been solved.
Thank you very much

PR #1152, opened 2024-08-09
Server supports custom headers

Shadow243:
### Description

Based on a discussion from the [Stalwart GitHub issue https://github.com/stalwartlabs/mail-server/discussions/477](https://github.com/stalwartlabs/mail-server/discussions/477), it appears that Stalwart's IMAP server only indexes standard headers defined in RFCs and does not support searching custom headers (such as `X-Auto-Bcc`) by default. This design choice is intended to optimize server performance by maintaining a manageable index size, as most users typically search by common fields like recipients, subjects, or message contents.

### Key Points
- **Custom Header Search Not Supported:** Stalwart currently does not support searching for custom headers (e.g., `X-Auto-Bcc`) because these headers are not indexed.

A `server_supports_custom_headers` function is implemented to determine if the server supports searching custom headers. This function sends a test command to the server and inspects the response to check for support.

[Related Issue: https://github.com/cypht-org/cypht/issues/1149](https://github.com/cypht-org/cypht/issues/1149)

Screenshot 2024-08-09 at 03 35 41

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -1099,7 +1099,7 @@ if (!class_exists('Hm_IMAP')) {
             if ($only_auto_bcc) {
                $fld .= ' HEADER X-Auto-Bcc cypht';
             }
-            if (!mb_strstr($this->server, 'yahoo') && $exclude_auto_bcc) {
+            if ($exclude_auto_bcc && !mb_strstr($this->server, 'yahoo') && $this->server_supports_custom_headers()) {
                $fld .= ' NOT HEADER X-Auto-Bcc cypht';
             }
             $esearch_enabled = false;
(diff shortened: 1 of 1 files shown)


marclaporte:
@Shadow243 I am fine to do a workaround for just one server, but please add a link to https://github.com/stalwartlabs/mail-server/discussions/477 in the source code so we remove the workaround if ever Stalwart no longer needs it.

Shadow243:
> @Shadow243 I am fine to do a workaround for just one server, but please add a link to [stalwartlabs/mail-server#477](https://github.com/stalwartlabs/mail-server/discussions/477) in the source code so we remove the workaround if ever Stalwart no longer needs it.

Done.

modules/imap/hm-imap.php
```diff
@@ -1099,7 +1099,7 @@ public function search($target='ALL', $uids=false, $terms=array(), $esearch=arra
             if ($only_auto_bcc) {
                $fld .= ' HEADER X-Auto-Bcc cypht';
             }
-            if (!mb_strstr($this->server, 'yahoo') && $exclude_auto_bcc) {
+            if ($this->server_supports_custom_headers() && !mb_strstr($this->server, 'yahoo') && $exclude_auto_bcc) {
```
kroky, line 1102: Note that this adds one more IMAP query to every search query which will have a performance impact. I suggest you reverse the order of checks here, so it is only called when needed - e.g. $exclude_auto_css && ! mb_strstr && $this->server_supports_custom_headers...

Shadow243: Done

marclaporte:
Potential regression: https://github.com/cypht-org/cypht/issues/1195

modules/imap/hm-imap.php
```diff
@@ -1099,7 +1099,7 @@ public function search($target='ALL', $uids=false, $terms=array(), $esearch=arra
             if ($only_auto_bcc) {
```
Shadow243, line 1099: I think the same conditions must have been here too

marclaporte:
A necessary follow up: https://github.com/cypht-org/cypht/pull/1651#issuecomment-3249394383

PR #1158, opened 2024-08-11
Fix default_settings configured from .env file

Shadow243:
Time zone and language variables defined in the .env file are not set by default by the application.

Related issue: [https://github.com/YunoHost-Apps/cypht_ynh/pull/69#issuecomment-2124295348](https://github.com/YunoHost-Apps/cypht_ynh/pull/69#issuecomment-2124295348)


index.php
```diff
@@ -30,15 +30,15 @@
...
-date_default_timezone_set( 'UTC' );
-
 require VENDOR_PATH.'autoload.php';
 /* get includes */
 require APP_PATH.'lib/framework.php';
 $environment = Hm_Environment::getInstance();
 $environment->load();
 
+/* set default TZ */
+date_default_timezone_set(env('DEFAULT_SETTING_TIMEZONE', 'UTC'));
```
kroky, line 40: Can you please use that from config and not env?

lib/config.php
```diff
@@ -103,7 +103,7 @@ public function get($name, $default = false) {
      * @return void
      */
     public function set_tz() {
-        date_default_timezone_set($this->get('timezone_setting', 'UTC'));
+        date_default_timezone_set($this->get('timezone_setting', env('DEFAULT_SETTING_TIMEZONE', 'UTC')));
```
kroky, line 106: Please use from config

lib/environment.php
```diff
@@ -102,7 +102,7 @@ function merge_config_files($folder_path) {
      * @return array The configuration array.
      */
     function config($file_name) {
-        $path = CONFIG_PATH.$file_name.'.php';
+        $path = APP_PATH.'config/'.$file_name.'.php';
```
kroky, line 105: CONFIG_PATH is defined in index.php, why don't you use it?

modules/core/output_modules.php
```diff
@@ -635,6 +635,7 @@ protected function output() {
             '    }'.
             '    return key;'.
             '};';
+        $res .= 'window.hm_default_timezone = "'.env('DEFAULT_SETTING_TIMEZONE').'";';
```
kroky, line 638: Please use config, not env.

modules/core/output_modules.php
```diff
@@ -1188,22 +1189,23 @@ class Hm_Output_timezone_setting extends Hm_Output_Module {
      */
     protected function output() {
         $zones = timezone_identifiers_list();
-        $settings = $this->get('user_settings', array());
+        $settings = $this->get('user_settings', array());
```
kroky, line 1192: There's an extra whitespace at the end.

modules/core/output_modules.php
```diff
@@ -1188,22 +1189,23 @@ class Hm_Output_timezone_setting extends Hm_Output_Module {
      */
     protected function output() {
         $zones = timezone_identifiers_list();
-        $settings = $this->get('user_settings', array());
+        $settings = $this->get('user_settings', array());        
         $reset = '';
+        // exit(var_dump($settings));
```
kroky, line 1194: Please remove debugging code from the final PR to get merged.

modules/core/output_modules.php
```diff
@@ -1188,22 +1189,23 @@ class Hm_Output_timezone_setting extends Hm_Output_Module {
...
         if (array_key_exists('timezone', $settings)) {
             $myzone = $settings['timezone'];
         }
         else {
-            $myzone = false;
+            $myzone = array_key_exists('default_setting_timezone', config('app')) ? config('app')['default_setting_timezone'] : 'UTC';
```
kroky, line 1199: Why wouldn't config app be defined? I don't think you should use config function but simply $this->config - it should be set in modules.

modules/core/output_modules.php
```diff
@@ -1188,22 +1189,23 @@ class Hm_Output_timezone_setting extends Hm_Output_Module {
...
         }
         $res = '<tr class="general_setting"><td><label for="timezone">'.
             $this->trans('Timezone').'</label></td><td><select class="w-auto form-select form-select-sm" id="timezone" name="timezone">';
         foreach ($zones as $zone) {
             $res .= '<option ';
             if ($zone == $myzone) {
                 $res .= 'selected="selected" ';
-                if ($zone != 'Africa/Abidjan') {
-                    $reset = '<span class="tooltip_restore" restore_aria_label="Restore default value"><i class="bi bi-arrow-repeat refresh_list reset_default_value_select"></i></span>';
+                if ($zone != config('app')['default_setting_timezone']) {
```
kroky, line 1207: Please use proper access to config like other modules do.

modules/core/site.js
```diff
@@ -1815,6 +1815,17 @@ var reset_default_value_checkbox = function() {
     }
 };
 
+var reset_default_timezone = function() {
+    var hm_default_timezone = window.hm_default_timezone;
+    var timezoneSelect = document.getElementById('timezone');
+    // Find the option that matches hm_default_timezone
+    for (var i = 0; i < timezoneSelect.options.length; i++) {
+        if (timezoneSelect.options[i].value === hm_default_timezone) {
+            timezoneSelect.options[i].selected = true;
```
kroky, line 1824: Isn't $('#timezone').val(hm_default_timezone) faster?

lib/environment.php
Shadow243: Copy that, I'l have to redeclare it also in tests/phpunit/bootstrap.php to use it withtout error.

Shadow243:
Changes applied.

Shadow243:
@kroky this on is ready for review.

PR #1171, opened 2024-08-14
[FIX] hm-imap.php PHP NOTICE Deprecated:  Creation of dynamic propert…

amaninyumu1:


modules/imap/hm-imap.php
```diff
@@ -143,7 +143,7 @@ class Hm_IMAP extends Hm_IMAP_Cache {
         public $con_error_num = 0;
 
         public $banner = '';
-
+        public $namespace_count = 0;
```
Shadow243, line 146: can you make it private ?

amaninyumu1: yes @Shadow243

PR #1176, opened 2024-08-15
Refresh lists after adding a account using New User Experience 

henochit:
## Pullrequest
This fix allows the refresh of the server list after adding an account through the new user experience wizard.

### DIFF
--- a/modules/nux/site.js
+++ b/modules/nux/site.js
@@ -57,7 +57,7 @@ var display_final_nux_step = function(res) {
                 }
             );
         } else {
-            Hm_Utils.redirect();
+            Hm_Utils.redirect('?page=servers');
         }
     }
 };


modules/nux/site.js
```diff
@@ -57,7 +57,7 @@ var display_final_nux_step = function(res) {
                 }
             );
         } else {
-            Hm_Utils.redirect();
+            Hm_Utils.redirect('?page=servers');
```
Shadow243, line 60: at https://github.com/cypht-org/cypht/blob/master/modules/core/site.js#L1711 we have window.location.href that is already redirecting to the current URL of the page that is being displayed in the browser, in this case it will be `www.domain.com?page= servers`, even if we did not specify this parameter.

Have you had a case? I can't reproduce a one that redirects to an another page than `servers`.

marclaporte:
Once merged to master, please backport to 2.x

modules/nux/site.js
henochit: Hi @Shadow243,
You’re right! After updating to the latest version of the master branch, I’m no longer able to reproduce the bug. Let me test various cases to ensure it's completely fixed.

Baraka24: I can not reproduce this too... I have tried several time. @henochit have you tried too so we know...

Baraka24: Isn't it related to the no redirect while adding an account from home page quick add section.... I got this:
servers_quick_add_section_from_home_page
@henochit ?

PR #1181, opened 2024-08-20
Should be off by default | Address the controversial part in read rec…

Shadow243:
I turned off this feature([(was: Cypht: MDN | DSN Missing in compose)](https://avan.tech/item89125)) as a default setting as  asked in 2024-05: https://github.com/cypht-org/cypht/issues/687

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -392,7 +392,8 @@ class Hm_Handler_load_smtp_servers_from_config extends Hm_Handler_Module {
         }
 
         $this->out('uploaded_files', get_uploaded_files($draft_id, $this->session));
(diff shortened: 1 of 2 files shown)


marclaporte:
Once merged to master, please backport to 2.x

modules/smtp/modules.php
```diff
@@ -407,6 +408,7 @@ public function process() {
             $this->out('compose_draft', $draft);
         }
         $this->out('smtp_compose_type', $compose_type);
+        $this->out('user_settings', $settings);
```
josaphatim, line 411: Instead of passing the whole user_settings you can just pass enable_compose_delivery_receipt_setting

Shadow243: Working on it now

Shadow243: @josaphatim done.

PR #1188, opened 2024-08-21
[Fix] Add error message for a missing database connexion

GedeonTS:
Added an alert message when the db is not connected

### DIFF
--- a/lib/db.php
+++ b/lib/db.php
@@ -143,6 +143,7 @@ class Hm_DB {
             return self::$dbh[$key];
         } catch (Exception $oops) {
             Hm_Debug::add($oops->getMessage());
+            Hm_Msgs::add('ERRUnable to connect to the database. Please check your configuration settings and try again.');
             self::$dbh[$key] = false;
             return false;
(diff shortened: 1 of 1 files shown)


lib/db.php
```diff
@@ -143,6 +143,7 @@ static public function connect($site_config) {
             return self::$dbh[$key];
         } catch (Exception $oops) {
             Hm_Debug::add($oops->getMessage());
+            Hm_Msgs::add('ERRUnable to connect to the database. Please check your configuration settings and try again.');
```
Shadow243, line 146: There can be several reasons and not necessarily the configuration issue. I feel like this error message only addresses one case.

Shadow243: Any thought @kroky ?

kroky: Since the debugging msg contains the actual error, I think we can have a relatively short user facing message that database connection cannot be established and site/app must be configured correctly before proceeding. The admin configuring the app will know where to look to find the debugging message and fix the issue.

PR #1190, opened 2024-08-22
[FIX]Profiles: Fixing undefined array key rmk on Profiles page

Baraka24:
Fixing: 
`
[Thu Aug 22 15:57:39.322175 2024] [php:warn] [pid 7452:tid 1204] [client ::1:13054] PHP Warning:  Undefined array key "rmk" in C:\\Apache24\\htdocs\\my_cypht\\modules\\profiles\\modules.php on line 289, referer: http://localhost/my_cypht/?page=profiles
[Thu Aug 22 15:57:39.322175 2024] [php:notice] [pid 7452:tid 1204] [client ::1:13054] PHP Deprecated:  mb_strlen(): Passing null to parameter #1 ($string) of type string is deprecated in C:\\Apache24\\htdocs\\my_cypht\\modules\\profiles\\modules.php on line 289, referer: http://localhost/my_cypht/?page=profiles
`
or
rmk

This happens when we just add a new server to Cypht while **Set this profile default** checkbox is checked.

### DIFF
--- a/modules/profiles/functions.php
+++ b/modules/profiles/functions.php
@@ -3,10 +3,11 @@
 if (!defined('DEBUG_MODE')) { die(); }
 
 if (!hm_exists('add_profile')) {
-    function add_profile($name, $signature, $reply_to, $is_default, $email, $server_mail, $smtp_server_id, $imap_server_id, $context) {
+    function add_profile($name, $signature, $reply_to, $is_default, $email, $server_mail, $smtp_server_id, $imap_server_id, $context, $remark = '') {
         $profile = array(
             'name' => $name,
             'sig' => $signature,
+            'rmk' => $remark,
             'smtp_id' => $smtp_server_id,
             'replyto' => $reply_to,
             'default' => $is_default,


modules/core/handler_modules.php
```diff
@@ -1128,7 +1128,7 @@ public function process() {
                         return;
                     }
 
-                    add_profile($profileName, $profileSignature, $profileReplyTo, $profileIsDefault, $email, $imapAddress, $this->smtp_server_id, $this->imap_server_id, $this);
+                    add_profile($profileName, $profileSignature, $remark, $profileReplyTo, $profileIsDefault, $email, $imapAddress, $this->smtp_server_id, $this->imap_server_id, $this);
```
Shadow243, line 1131: Where is $remark defined here ?

modules/profiles/functions.php
```diff
@@ -3,10 +3,11 @@
 if (!defined('DEBUG_MODE')) { die(); }
 
 if (!hm_exists('add_profile')) {
-    function add_profile($name, $signature, $reply_to, $is_default, $email, $server_mail, $smtp_server_id, $imap_server_id, $context) {
+    function add_profile($name, $signature, $remark = '', $reply_to, $is_default, $email, $server_mail, $smtp_server_id, $imap_server_id, $context) {
```
Baraka24, line 6: @Shadow243 $remark is defined here.

Baraka24: It is defined here: https://github.com/cypht-org/cypht/pull/1190/files#r1727773754

Shadow243: Yes but calling the fonction `add_profile` you are passing `$remark` but i can't see where it's coming from. You mean you can pass an undefined variable ?

Shadow243: See even the VSCode 
Screenshot 2024-08-22 at 23 34 20

Shadow243: I'm, sure, it's defined on the form, you will get it from https://github.com/cypht-org/cypht/blob/master/modules/core/handler_modules.php#L990 and https://github.com/cypht-org/cypht/blob/master/modules/core/handler_modules.php#L1016

Shadow243: > @Shadow243 $remark is defined here.

You can move param `$remark` at the end in the fonction definition, like that you dont need to pass it while calling. like that you will then remove it at line 1131.

PR #1191, opened 2024-08-22
[FIX] ensure the email address part is the only value captured when a…

Shadow243:
Cherry-picked from https://github.com/cypht-org/cypht/pull/1166


Shadow243:
Il merge all master to 2.x as they are all fixes.

PR #1193, opened 2024-08-22
Correct settings.py & tags.py selenium tests and uncomment it

Shadow243:
Correct broken selenium settings tests


config/app.php
```diff
@@ -1144,21 +1144,21 @@
     | Timezone for date displays
     | Defaults to UTC
     */
-    // 'default_setting_timezone' => env('DEFAULT_SETTING_TIMEZONE', 'UTC'),
+    'default_setting_timezone' => env('DEFAULT_SETTING_TIMEZONE', 'UTC'),
```
kroky, line 1147: I wonder why are all the default settings commented here - they are coming from .env and stock .env.example should have the correct values. I think we should uncomment all the defaults here and let the user choose defaults either in .env or configure per-user site settings.

.github/tests/.env
```diff
@@ -28,10 +28,10 @@ DEFAULT_SMTP_PORT=
 DEFAULT_SMTP_TLS=
 DEFAULT_SMTP_NO_AUTH=
 
-USER_CONFIG_TYPE=DB
-USER_SETTINGS_DIR=/tmp
-ATTACHMENT_DIR=/tmp
-APP_DATA_DIR=/tmp
+USER_CONFIG_TYPE=file
```
kroky, line 31: Are you sure the github tests are not working with db?

config/app.php
Shadow243: Copy that, can i add them via this PR: https://github.com/cypht-org/cypht/pull/1158 ?

kroky: Sure, thanks!

.github/tests/.env
Shadow243: Updated

.github/tests/setup.sh
```diff
@@ -28,6 +28,11 @@ setup_cypht() {
         sed -i 's/mysql/sqlite/' tests/phpunit/mocks.php
         sed -i "s/'host'/'socket'/" tests/phpunit/mocks.php
     fi
+
```
josaphatim, line 31: No need for these sed below as again as USER_CONFIG_TYPE is set to DB

.github/tests/setup.sh
```diff
@@ -132,6 +137,12 @@ setup_site() {
 		STATUS_ERROR
 		exit 1
 	fi
+    STATUS_TITLE "Setup required Directories"
```
josaphatim, line 140: When can also remove these below for the same reason

tests/selenium/folder_list.py
```diff
@@ -59,7 +59,7 @@ def show_folders(self):
 
     print("FOLDER LIST TESTS")
     test_runner(FolderListTests, [
-        'reload_folder_list',
+        # 'reload_folder_list',
```
josaphatim, line 62: Should this be commented again ?

tests/selenium/tags.py
```diff
@@ -11,17 +11,20 @@ def __init__(self):
...
-        # ((JavascriptExecutor)driver).executeScript("arguments[0].scrollIntoView(false);", element);
-
-
-        self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
-        self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
-        self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
-        self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
-        self.by_css('[data-source=".tags_folders"]').click()
+        element = self.by_css('[data-source=".tags_folders"]')
+        print("NNNN")
```
josaphatim, line 15: This debugging too

.github/tests/setup.sh
Shadow243: Even with USER_CONFIG_TYPE=DB we still need them, as you can see with this screenshort i tried to remove them to show you the result.
Screenshot 2024-08-29 at 01 58 15

tests/selenium/folder_list.py
Shadow243: It was commented, it's passing by sometimes no. this is unstable i must find a good way that will not be working and no working at an other time.

Shadow243:
tags.py activated and corrected too.

.github/tests/setup.sh
Shadow243: I removed all and keep ATTACHMENT_DIR only.

.github/tests/setup.sh
Shadow243: idem

PR #1196, opened 2024-08-25
Adding idle and polling mechanism to imap

Shadow243:
This PR will address the IDLE & Polling integration into cypht to work with webhook and start pushing realtime notifications. 

This is just a template for now, I still need to test and improve.

- Related Issue: [https://github.com/cypht-org/cypht/issues/178](https://github.com/cypht-org/cypht/issues/178)
- Related PR: [https://github.com/cypht-org/cypht/pull/1056](https://github.com/cypht-org/cypht/pull/1056)


console
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env php
+<?php
+define('SKIP_HM_DISPATCH', true);
+
+
+// Load the application dependencies and bootstrap the application
+require_once __DIR__ . '/index.php';
```
kroky, line 7: Load the dependencies here and don't include index.php. Treat console as a separate entry point in the application. index.php will be for web requests and this one for cli.

composer.json
```diff
@@ -38,6 +38,7 @@
         "bacon/bacon-qr-code": "^1.0.3 || ^2.0.0",
         "christian-riesen/base32": "^1.3.2",
         "composer": "^2.0.0",
+        "cypht/cli-service": "*",
```
kroky, line 41: Why do we define another package? It doesn't seem to be runnable without main cypht repo, so better stay as just a subdirecrory in this repo. No need for composer dependencies here.

kroky:
Please ping me when you add the ratchet lib and polling, so I can review again.

Shadow243:
Channels tested and fixed

Screenshot 2024-12-09 at 18 24 36
Screenshot 2024-12-09 at 19 19 48

Shadow243:
schedule:work also working correctly, i'm receiving Notification every minute as a planned:
Screenshot 2024-12-09 at 19 52 16

Shadow243:
@kroky  Could you review this for me again ? I'm currently working on integrating it with the UI, and the plan is to launch everything through the interface. The Ratchet integration is not implemented yet.

kroky:
I think we are heading in the right direction but I wonder if we really need all those classes - I see concepts like scheduler, jobs, queue manager, event notifications, etc. - since we are already using symfony components, aren't there libs to use for these implementations? I think we are implementing way too many things currently on ourselves and the code looks similar to other laravel or symfony components. Did you take inspiration from somewhere? If so, can't we use those libs as they will be better maintained and tested if we do so?
Also, there is no need to prefix everything with Hm_ in services - it is a new module and can be named normally.

marclaporte:
> I see concepts like scheduler, jobs, queue manager, event notifications, etc.

FYI, this is coming soon: https://gitlab.com/tikiwiki/tiki/-/merge_requests/5453/

kroky:
After a discussion with @Shadow243 about the reasoning behind the manual implementation of the new concepts, we decided to go ahead with this approach rather than introducing a ton of external dependencies which would require even more 3rd party libs. Some considerations from the discussion with Steven:

> 1. Existing Libraries: While there are Symfony and Laravel libraries for schedulers, jobs, queues, etc., many come with a lot of dependencies we don’t need. For example, Laravel’s illuminate/queue pulls in things like illuminate/database, illuminate/support, and ramsey/uuid, etc..., which can bloat the project unnecessarily.  
> 
> 2. ORM Compatibility: Since we are using Cypht without an ORM, libraries requiring Doctrine(ex: symfony/messenger) won’t integrate cleanly and could add unnecessary complexity.  
> 
> 3. Custom Implementations: Yes, our code resembles parts of Laravel and Symfony components because we’ve taken inspiration from established patterns. However, the custom implementations are tailored specifically to our needs, keeping things lightweight and avoiding dependencies we won’t use.  
> 
> 4. Package Maintenance: Many libraries we researched were abandoned or didn’t fit our ecosystem. Laravel’s ecosystem is advanced, but it often forces you to pull in extra packages that aren’t relevant to our use case.  
> 
> 5. Symfony Messenger and Event Dispatcher: I could combine symfony/messenger with symfony/event-dispatcher, symfony/contracts, and symfony/dependency-injection to manage these functionalities. However, I’d face a challenge with queueing since Symfony doesn’t have a dedicated queue component without internal depandacies(like ORM, etc...).  
> 
> 
> This approach ensures flexibility without overloading the project. If you know of lightweight, actively maintained libraries that fit our needs, let me know, and we can evaluate them.  
> 
> Most Important: Features like the scheduler, events, listeners, and queues are highly advanced in Symfony, just like in Laravel, and have interdependencies that could pose problems. Cypht, on the other hand, intentionally remains simple (no ORM, etc.), while Laravel and Symfony rely on more complex components to manage a large part of their functionalities.
> 
> E.g: composer require illuminate/queue requires
> 
> php: ^8.2
> illuminate/collections: ^11.0
> illuminate/console: ^11.0
> illuminate/container: ^11.0
> illuminate/contracts: ^11.0
> illuminate/database: ^11.0
> illuminate/filesystem: ^11.0
> illuminate/pipeline: ^11.0
> illuminate/support: ^11.0
> laravel/serializable-closure: ^1.3|^2.0
> ramsey/uuid: ^4.7
> symfony/process: ^7.0.3
> 
>

composer.json
```diff
@@ -35,22 +40,33 @@
...
-        "composer": "^2.0.0",
+        "php": ">=8.1",
         "ext-curl": "*",
         "ext-fileinfo": "*",
         "ext-iconv": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "ext-session": "*",
+        "composer": "^2.0.0",
```
marclaporte, line 51: Please keep this list alphabetically sorted.

marclaporte:
We will soon release Cypht 2.5.0 so this MR is too big/risky to be part of it.

But after, we must decide if this is suitable for Cypht 2.6.x or rather 3.0.x

Shadow243:
> We will soon release Cypht 2.5.0 so this MR is too big/risky to be part of it.
> 
> But after, we must decide if this is suitable for Cypht 2.6.x or rather 3.0.x

I share that opinion.

Shadow243:
> We will soon release Cypht 2.5.0 so this MR is too big/risky to be part of it.
> 
> But after, we must decide if this is suitable for Cypht 2.6.x or rather 3.0.x

This PR will be suitable for 3.0.x

Issue #1153, opened 2024-08-09
Sieve - can't make it work


europacafe:
## 💬 Question
Mail server: stalwart + snappymail webmail ---> tls sieve work (mail.mydomain.com, ssl/tls, 4190)
Cypht 2.1.0 + stalwart --> tls sieve setting does not work, can't save server config

When opening a mail content, it always reports "Sieve: invalid config host for ....my mailbox name.."
![cypht_sieve_invalid_config](https://github.com/user-attachments/assets/72f28639-d098-4a71-bdf1-4c43d28a3481)

On stalwart mail server, below is the sieve config
![stalwart_sieve_tls_config](https://github.com/user-attachments/assets/d91123bb-190c-4133-88b8-2287a073c628)

On snappymail webmail, the sieive config works
![snappymail_sieve_config](https://github.com/user-attachments/assets/320d895f-c9db-42e0-bfd2-0ccbff6d0d26)

On cypht env file, below is the config:
![cypht_sieve_env_config](https://github.com/user-attachments/assets/0821f0ce-0f2a-488b-b857-ecab47ac9de4)

When I tried to enable sieve on one of my IMAP accounts as below, after saving the config, the screen froze for over 10 secs and came back with an offline label and it didn't save.
![image](https://github.com/user-attachments/assets/7113de0d-a7c1-4191-a442-8a6c8d378171)
I'm not sure what I did wrong. (I enable sieve on Settings/Site/Enable sieve filter)

josaphatim:
@europacafe thank you for reporting this. What if you remove the protocol in sieve host input `mail.brhofen.com:4190` ?

europacafe:
I did try that too. Not working as well.

marclaporte:
Please try Cypht master

europacafe:
How? (with docker compose)

marclaporte:
> How? (with docker compose)

@jonocodes This is not yet coded, right?

marclaporte:
Please try this: https://github.com/cypht-org/cypht/pull/1135/files (just manually modify the file)

europacafe:
just changed to `$server = Hm_IMAP_List::get($imap_server_id, true);`, restarted cypht container
same behavior for both tls://mail.mydomain.com:4190 and mail.mydomain.com:4190

Shadow243:
@europacafe have you checked that the server firewall gives access to 4190 on your server ?

europacafe:
[...]

PR #1199, opened 2024-08-27
Correct sieve connection when using scheme

Shadow243:
When using a scheme while connecting to Sieve, we encounter an error because the scheme is not being utilized. As a result, when TLS is active, there is no way to connect to Sieve unless you deactivate it.

- Related issue: [https://github.com/cypht-org/cypht/issues/1153](https://github.com/cypht-org/cypht/issues/1153)


modules/smtp/hm-smtp.php
```diff
@@ -60,6 +60,9 @@ class Hm_SMTP {
     private $config;
     private $server;
     private $starttls;
+    private $supports_tls;
```
Shadow243, line 63: These are not really related to sieve corrections, just some deprecations that i've been getting while correcting.
Screenshot 2024-08-27 at 05 20 41

modules/imap/functions.php
```diff
@@ -1495,16 +1495,15 @@ function forward_dropdown($output,$reply_args) {
 if (!hm_exists('parse_sieve_config_host')) {
 function parse_sieve_config_host($host) {
     $url = parse_url($host);
-    $host = $url['host'] ?? $url['path'];
+    if(!isset($url['host'])) {
+        $host = $url['path'];
+    }
     $port = $url['port'] ?? '4190';
-    $scheme = $url['scheme'] ?? 'tcp://';
-    $tls = $scheme === 'tls';
```
Shadow243, line 1501: With migadu we don't add `tls://` at the begininng but with Stalwart if it's implicit we must add it.

josaphatim:
@Shadow243 can you please rebase

Personally I'm not convinced with this fix. I have tested it with Postale and Migadu which do not work. @kroky can you please give your opinion ?

Shadow243:
> @Shadow243 can you please rebase
> 
> 
> 
> Personally I'm not convinced with this fix. I have tested it with Postale and Migadu which do not work. @kroky can you please give your opinion ?

https://github.com/cypht-org/cypht/pull/1199#discussion_r1733438371

Shadow243:
> @Shadow243 can you please rebase
> 
> I have tested it with Postale and Migadu which do not work. @kroky can you please give your opinion ?

There is a small change on the User interface, here some screenshorts that will show you:
Screenshot 2024-09-12 at 16 29 05

I added a checkbox that will be used to enable or disable TLS, instead of relying on the Sieve host starting with 'tls://'. With Migadu, even when TLS mode is set to true, we use the host without 'tls://'. However, with Stalwart, 'tls://' is required. In the older version, we were removing it before sending the request to the server.
Screenshot 2024-09-12 at 16 33 53

kroky:
I think we need these changes - relying on tls scheme is not enough, it is better to specify host and port separately and then turn on or off the tls via a checkbox. So, in terms of UI it is ok. Just check if the changed method signatures (connect_to_imap_server) is OK when used throghout the app. I don't see other problems but I haven't test this code.

josaphatim:
Thanks @kroky
@Shadow243 Then you need also to refactor IMAP_AUTH_SIEVE_CONF_HOST found in .env

Shadow243:
> Thanks @kroky @Shadow243 Then you need also to refactor IMAP_AUTH_SIEVE_CONF_HOST found in .env

@josaphatim it's now done. just a concern, if i add for example a wrong sieve url i'l getting time out error, it coming from henrique-borba/php-sieve-manager: `readResponse` function.

modules/sievefilters/hm-sieve.php
```diff
@@ -2,10 +2,10 @@
 <?php
 
 class Hm_Sieve_Client_Factory {
-    public function init($user_config = null, $imap_account = null)
+    public function init($user_config = null, $imap_account = null, $sieve_tls=false)
```
josaphatim, line 5: No need to add $sieve_tls parameter as it is already in $imap_account

.env.example
```diff
@@ -135,7 +136,7 @@ DEFAULT_SETTING_GITHUB_LIMIT=20
 DEFAULT_SETTING_GITHUB_SINCE='-1 weeks'
 DEFAULT_SETTING_INLINE_MESSAGE=false
 DEFAULT_SETTING_ENABLE_KEYBOARD_SHORTCUTS=1
-DEFAULT_SETTING_ENABLE_SIEVE_FILTER=false
+DEFAULT_SETTING_ENABLE_SIEVE_FILTER=true
```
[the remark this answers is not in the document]
Shadow243: No, i'l change

modules/sievefilters/hm-sieve.php
Shadow243: $imap_account has not that tls param I can add it thought as it's an array

config/app.php
```diff
@@ -135,10 +135,18 @@
     /*
     |
     | The hostname/IP address and port sieve is listening on. Example: example.org:4190
```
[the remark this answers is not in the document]
Shadow243: @josaphatim this comment is correct.

Shadow243:
@josaphatim corrections applied.

PR #1202, opened 2024-08-28
Check for null in quota-root. Some servers yield null values here

### LINKED ISSUE #1201: None value returned when checking for quota on dovecot imap server
## 🐛 Bugreport
When checking for quota, my imap server (dovecot) is returning none from the get_quota_root() call in handler_modules.php.

Unclear whether this causes malfunction, found while looking through the logs for another issue that was stopping things anyway. But the logs get cluttered with error messages and stack traces.

Found on commit 6644dd4a98d27b1250010cfb82aa7fd3c3752789

indridieinarsson:
### Issues
- fixes #1201

### Checklist
should be fairly straightforward, very unlikely to cause issues. 

### How2Test

- [ ] Check that imap servers with user quotas are displayed correctly.
- [ ] Check that imap servers without quotas, but for which quota check is still performed for some reasons, are bailed out silently, instead of logging an error message to the logs.

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -757,13 +757,15 @@ class Hm_Handler_imap_folder_expand extends Hm_Handler_Module {
             $imap = Hm_IMAP_List::connect($form['imap_server_id'], $cache);
             if (imap_authed($imap)) {
                 $quota_root = $imap->get_quota_root($folder ? $folder : 'INBOX');
-                $quota = $imap->get_quota($quota_root[0]['name']);
-                if ($quota) {
-                    $current = floatval($quota[0]['current']);
-                    $max = floatval($quota[0]['max']);
-                    if ($max > 0) {
-                        $this->out('quota', ceil(($current / $max) * 100));
-                        $this->out('quota_max', $max / 1024);
+                if ($quota_root) {
+                    $quota = $imap->get_quota($quota_root[0]['name']);
+                    if ($quota) {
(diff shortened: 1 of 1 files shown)


modules/imap/handler_modules.php
```diff
@@ -757,13 +757,15 @@ public function process() {
...
             if (imap_authed($imap)) {
                 $quota_root = $imap->get_quota_root($folder ? $folder : 'INBOX');
-                $quota = $imap->get_quota($quota_root[0]['name']);
-                if ($quota) {
-                    $current = floatval($quota[0]['current']);
-                    $max = floatval($quota[0]['max']);
-                    if ($max > 0) {
-                        $this->out('quota', ceil(($current / $max) * 100));
-                        $this->out('quota_max', $max / 1024);
+                if ($quota_root) {
```
Baraka24, line 760: his condition ensures that $quota_root is not null or false. However, there's still a potential issue if $quota_root is an empty array. The check if ($quota_root) will pass for an empty array, but accessing $quota_root[0]['name'] would still result in the "Trying to access array offset on null" warning. 
Can you check that $quota_root contains elements before accessing $quota_root[0]['name'] like:
`if ($quota_root && isset($quota_root[0]['name']))`.

Baraka24 (commented):
Thanks @indridieinarsson  so much for this great PR, if you can kindly make the changes from the review I'll be gladly to merge it.

Issue #1129, opened 2024-07-22
Incorrect Blocking and Adding Behavior for Email Address in Contacts List


Baraka24:
## 🐛 Bugreport

When blocking the email example@gmail.com, the success message incorrectly states "Sender Unblocked" instead of "Sender Blocked". Additionally, after blocking the email, if you try to add the same email to the trusted contacts list, it gets added but remains blocked. Ideally, there should be an alert stating, "This sender has been blocked and cannot be added to the trusted contacts list."

### Version & Environment
Cypht Version: master

### Steps to reproduce
1. Log in to Cypht.
2. Open a message from example@gmail.com and in its footer section Block the email .
3. Observe the success message which says "Sender Unblocked" instead of "Sender Blocked".
4. Attempt to add the same email example@gmail.com to the trusted contacts list.
5. Observe that the email gets added to the trusted contacts list but remains blocked.

### Possible Solution:

- [ ] Correct the success message to "Sender Blocked" when an email is blocked.
- [ ] Implement an alert that prevents adding a blocked email to the trusted contacts list, stating, "This sender has been blocked and cannot be added to the trusted contacts list."

PR #1213, opened 2024-08-30
[FIX] block_list page: reviewing unblock and block actions

christer77:
* Malfunctions were observed when unblocking senders in the block list.
* Fixing display view of blocked senders or domains
* Related [issue](https://github.com/cypht-org/cypht/issues/1129)

### DIFF
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -325,7 +325,7 @@ class Hm_Handler_sieve_unblock_sender extends Hm_Handler_Module {
         }
 
         foreach ($this->user_config->get('imap_servers') as $idx => $mailbox) {
-            if ($idx == $this->request->post['imap_server_id']) {
+            if ($idx == $form['imap_server_id']) {
                 $imap_account = $mailbox;
             }
         }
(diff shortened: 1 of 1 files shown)


modules/sievefilters/modules.php
```diff
@@ -462,7 +462,7 @@ public function process() {
                 Hm_Msgs::add('ERRIMAP Mailbox select error');
                 return;
             }
-            $msg_header = $imap->get_message_headers($form['imap_msg_uid']);
+            $msg_header = $imap->get_message_headers($this->request->post['imap_msg_uid']);
```
kroky, line 465: Can you please switch the usage of request->post variables directly in this whole method to $form array as these should have already been filtered appropriately before using in code. This will be in sync with all the other places Cypht is using process_form to process incoming variables.

modules/sievefilters/modules.php
```diff
@@ -359,7 +359,7 @@ public function process() {
             $unblock_sender = false;
             if ($current_script != '') {
                 $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[1]);
-                $blocked_list = json_decode(base64_decode($base64_obj));
+                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)));
```
kroky, line 362: Do we really want to replace all stars here? Maybe you have an incorrect json syntax because of a star but it is either in the beginning or the end of the script? I think we should be more careful with replacing here - probably use a regex for the exact problem you are trying to solve.

modules/sievefilters/modules.php
christer77: Notes. Thanks

Issue #1216, opened 2024-08-31
Can't send emails in 3019c66


PaulTGG:
## 🐛 Bugreport
Clicking the send button on the compose screen doesn't send the email.  (It appears that the email is saved on the server as a draft, but the message isn't actually sent.)

### Version & Environment

Rev: 3019c66

### Steps to reproduce
1. Install Cypht.
2. Compose an email.
3. Hit the Send button.

marclaporte:
Reportedly fixed by https://github.com/cypht-org/cypht/pull/1219

PR #1219, opened 2024-09-03
[FIX]Compose page: clicking on Send button can't send email

Baraka24:
[Related issue](https://github.com/cypht-org/cypht/issues/1216)

### DIFF
--- a/modules/smtp/site.js
+++ b/modules/smtp/site.js
@@ -419,29 +419,6 @@ var force_send_message = function() {
     }
 }
 
-var check_cc_exist_in_contacts_list = function(e) {
-        var compose_cc = $(".compose_cc").val().trim();
-        var list_cc = null;
-        var list_cc_not_exist_in_my_contact = [];
-        if (compose_cc.length > 0) {
-            list_cc = compose_cc.split(",");
-            var list_html = "<ol>";
-            list_cc.forEach(cc => {
-                cc = cc.trim().split(" ");
-                if (! list_emails.includes(cc.slice(-1)[0])) {
-                    list_cc_not_exist_in_my_contact.push(cc.slice(-1)[0])
-                    list_html += `<li>${cc.slice(-1)[0]}</li>`;
-                }
-            });
-            list_html += "</ol>";
-
-            if (list_cc_not_exist_in_my_contact) {
-                return list_html;
-            }
-        }
-    return "";
-};
-
 $(function () {
     if (!hm_is_logged()) {
         return;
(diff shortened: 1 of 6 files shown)


modules/smtp/site.js
```diff
@@ -512,14 +512,14 @@ $(function () {
 
             // if contact_cc not exist in contact list for user
             var checkInList = "";
-            if (list_emails) {
+            if (typeof list_emails !== "undefined") {
```
Shadow243, line 515: Migrate this fonctionnality into contact module.

Shadow243: it is related to the contact module, by leaving it here you create a dependency of the smtp module to the contact module

Baraka24: Done

Shadow243:
If you disable: `contacts,local_contacts,ldap_contacts`, you will gt this error:
```
Uncaught ReferenceError: check_cc_exist_in_contacts_list is not defined
    at HTMLButtonElement.<anonymous> (site.js:491:31)
    at HTMLButtonElement.La (cash.min.js:26:306)
```

Even in site.js from smpt you have to make sure check_cc_exist_in_contacts_list is defined before calling it, if not defined then overpass it.

modules/smtp/site.js
```diff
@@ -510,16 +487,16 @@ $(function () {
                 modalContentHeadline = "Your subject and body are empty!";
             }
 
-            // if contact_cc not exist in contact list for user
-            var checkInList = "";
-            if (list_emails) {
-                checkInList = check_cc_exist_in_contacts_list(e);
+            if (typeof check_cc_exist_in_contacts_list === 'function') {
```
Shadow243, line 490: This solution will not solve the issue, the goal is to be able to send an email even though module contact is no enabled, I think you should check if the module is enabled, if yes then active the feature, if no then we must be able to send an email.

Shadow243, line 490: It's works, Just an idea about how this can be refactored, I think there is a way to check if the module is enabled, we can check like that .

modules/smtp/site.js
```diff
@@ -510,16 +487,16 @@ $(function () {
...
+                var checkInList = check_cc_exist_in_contacts_list(e);
+
+                // if contact_cc not exist in contact list for user
                 if (checkInList) {
                     modalContentHeadline = "Adress mail not exist in your contact liste";
```
Shadow243, line 495: `list` no `liste`

Shadow243:
Here is an idea that you can implement, you create a function at `https://github.com/cypht-org/cypht/blob/master/modules/core/output_modules.php#L599` that will be checking if the module exists. ex: `hm_module_is_supported(module as param)`.In this function you will be checking in all modules that can be accessible there. if true then .....

You now have to replace your condition with this function. Also` check_cc_exist_in_contacts_list` you don't need to pass param: `e` because we are not using it.

Shadow243:
@josaphatim any suggestion before i merge this ?

Issue #1224, opened 2024-09-06
Stay alert to other potential regressions which arrived in Cypht 2.2.0 related to Unicode Support (Replace Standard PHP String Functions with Multibyte Counterparts)


marclaporte:
https://github.com/cypht-org/cypht/pull/1217 by @indridieinarsson
is a partial revert in 2 of the 57 files changed in https://github.com/cypht-org/cypht/pull/1051 by @kambereBr

There could be others, so let's stay alert.

@jasonmunro Can you think of places we should specifically be careful about?

indridieinarsson:
Some tips on this here:
https://stackoverflow.com/questions/10360764/there-are-simple-way-to-get-a-character-from-multibyte-string-in-php
and here :
https://stackoverflow.com/questions/3666306/how-to-iterate-utf-8-string-in-php

marclaporte:
Some work is happening: https://github.com/cypht-org/cypht/pull/1230

marclaporte:
All the related reported regressions have been solved, so closing this.

Issue #1222, opened 2024-09-05
Addressfields with Multibyte Characters in Displayname are cutted


MatthiasDomroes:
## 🐛 Bugreport
While processing Header-Field (e.g From) by using module/core/message_functions.php::process_address_fld($fld) address is cutted:
Raimund Matthias Domrös <matthias@domroes.net>
is converted to 
Raimund Matthias Domrös <matthias@domroes.ne>
The last t is omitted!

Error is caused by function addr_split. The function is coded as iteration over bytes but mb_strlen (counts charactes!) is used to calculate $max (line 500) .

### Version & Environment
Release 2.2.0 / file message_functions.php of master is unchanged.

marclaporte:
Thank you @MatthiasDomroes

marclaporte:
Perhaps related: https://github.com/cypht-org/cypht/issues/1224

kroky:
Should be solved by https://github.com/cypht-org/cypht/pull/1230

PR #1217, opened 2024-09-01
When reading literal_data from mail body from IMAP, use byte count instead of string-length count.

indridieinarsson:
## Pullrequest
This is a first attempt to fix issue #1119. 
I'm not very proficient in php, and not knowledgeable about imap, so for the love of all that is good and holy, review and test this properly.

Take a look at this chunk from hm-imap-base.php:
> elseif ($line[$i] == '{') {
>                 $end = mb_strpos($line, '}');
>                 if ($end !== false) {
>                     $literal_size  = mb_substr($line, ($i + 1), ($end - $i - 1));
>                 }
>                 $lit_result = $this->read_literal($literal_size, $max, $current_size, $line_length);
>                 $chunk = $lit_result[0];
> 

If I understand this writing correctly: https://www.rfc-editor.org/rfc/rfc3501#section-4.3, the number in curly braces that is being read into $literal_size, is the number of bytes in the email body, which is then sent over the wire right away.
Therefore, all string-size calculations in `read_literal()` should use the `strsize` rather than `mb_strsize` command. My theory is that the issue #1119 is caused by this: the number of bytes read are underestimated since `mb_strsize` is smaller than `strsize` for real multibyte strings. We then end up trying to read strings from the server after the server has sent the entire message, causing the code to hang. 

### Issues

- Might possibly fix #1119 

### Checklist
This touches the code that reads the email body, pretty central for an email client. This PR will change behaviour in cases when the body of the message contains multibyte characters (ÞþæÆöÖðÐáÁóÓ etc.).

PR #1229, opened 2024-09-08
[FIX] Fix multibyte string handling for accurate byte and character operations

kambereBr:
This PR fixes multibyte string handling for accurate byte and character operations, preventing string processing issues.

### Issues
https://github.com/cypht-org/cypht/issues/1222
https://github.com/cypht-org/cypht/issues/1224


kroky:
https://github.com/cypht-org/cypht/pull/1230 should address the issues. Please don't revert to single-byte string functions on all those places. Check this comment for more info: https://github.com/cypht-org/cypht/pull/1217#issuecomment-2338082183

Issue #1119, opened 2024-07-15
FastCGI timeout (Not docker related)


indridieinarsson:
## 🐛 Bugreport

I am running cypht using docker compose. Building the docker image from source locally. 
Previous version ran without the issue described here (master branch ca. month ago). After updating to the most recent master branch, the issue began.

Cypht comes up as regular, and is usable (there are some issues regarding the Everything folder not displaying all accounts, but that was unchanged between versions).
After some usage, usually just a few minutes, cypth becomes unreachable, with the browser displaying just: "504 Gateway Time-out
nginx/1.26.1". Sometimes, browser says "404 page not found".

In the docker logs, I get the following:
```
<< fastCGI error is repeated a lot >>
cypht-1  | 172.22.0.2 - - [15/Jul/2024:11:17:56 +0000] "POST / HTTP/1.1" 504 569 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36"
cypht-1  | 127.0.0.1 -  15/Jul/2024:11:18:10 +0000 "POST /index.php" 200
cypht-1  | 172.22.0.2 - - [15/Jul/2024:11:18:10 +0000] "POST / HTTP/1.1" 200 78 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36"
cypht-1  | 127.0.0.1 -  15/Jul/2024:11:18:14 +0000 "GET /index.php" 200
cypht-1  | 127.0.0.1 - - [15/Jul/2024:11:18:14 +0000] "GET / HTTP/1.1" 200 510848 "-" "curl/8.7.1"
cypht-1  | 127.0.0.1 -  15/Jul/2024:11:18:26 +0000 "POST /index.php" 200
cypht-1  | 172.22.0.2 - - [15/Jul/2024:11:18:26 +0000] "POST / HTTP/1.1" 200 21209 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36"
cypht-1  | 127.0.0.1 -  15/Jul/2024:11:18:26 +0000 "POST /index.php" 200
[...]

PR #1230, opened 2024-09-09
fix multibyte handling in imap literals and address splitting, improve unit tests

kroky:
fix imap::read_literal to work on 7-bit bytes rather than multibyte strings, fix imap unit tests and add tests for recent bugs: accidentally reading next lines of literals and not splitting addresses correctly


kroky:
Supersedes: https://github.com/cypht-org/cypht/pull/1217

indridieinarsson:
Worth mentioning that this PR should also fix issue #1119

Issue #1165, opened 2024-08-13
Emails can't be moved to a new location anymore via drag and drop


jacob-js:
Seen on the latest master, holding a message via the messages list and moving it around in an attempt to drop it into a new folder does not do anything.
Maybe some code from https://github.com/cypht-org/cypht/pull/834 has been displaced?

marclaporte:
Thank you @jacob-js for the report. This is a good one for @GedeonTS to explore Cypht code :-)

PR #1233, opened 2024-09-10
[FIX] Move email: Enable drag-and-drop feature to move messages betwe…

GedeonTS:
https://github.com/cypht-org/cypht/issues/1165


modules/core/site.js
```diff
@@ -2127,7 +2127,6 @@ if(tableBody && !hm_mobile()) {
         sort: false,
         group: 'messages',
         ghostClass: 'drag_target',
-        draggable: ':not(.inline_msg)',
```
jacob-js, line 2130: @GedeonTS how is removing this line supposed to fix the issue? That property ensures that emails can still be read in inline mode when requested instead of triggering the drag action.

GedeonTS: That line is the one that breaks the feature, what do you suggest? @jacob-js

jacob-js: @GedeonTS you're basically removing this fix: https://github.com/cypht-org/cypht/pull/1034
I suggest first investigating the true cause of the issue, and then we can implement a proper fix.

GedeonTS: After more investigation I confirm that the true cause of the issue is https://github.com/cypht-org/cypht/pull/1034 @jacob-js , I suggest to revert it  and find another alternative for it

marclaporte: ok, so please work together to find another solution

jacob-js:
I'll do one more refactoring and this will be ready for a merge.

PR #1236, opened 2024-09-11
Added flatpickr for better date/time picker

josaphatim:
Added https://flatpickr.js.org/ for a beautiful and single design browser cross date picker.


marclaporte:
@josaphatim Let's try to align Tiki and Cypht dependencies:
- https://gitlab.com/tikiwiki/tiki/-/merge_requests/4879
- https://dev.tiki.org/Date-and-Time-Picker
- https://avan.tech/item6551

josaphatim:
> @josaphatim Let's try to align Tiki and Cypht dependencies:
> 
> * https://gitlab.com/tikiwiki/tiki/-/merge_requests/4879
> * https://dev.tiki.org/Date-and-Time-Picker
> * https://avan.tech/item6551

Great! As Flatpickr is not longer maintained I think to go with https://github.com/t1m0n/air-datepicker as it is the only supporting date and time. What do you think about this ?

marclaporte:
@jacob-js please advise.

jacob-js:
Cypht can consume the same component used in Tiki. It would only require getting the bundle of the file [datetime-picker.ce.js](https://gitlab.com/tikiwiki/tiki/-/blob/master/src/js/vue-widgets/datetime-picker/src/datetime-picker.ce.js) somewhere accessible so it can be installed in Cypht.

kroky:
Note that we need to distribute the package if we want to allow users to install standalone cypht. They can't randomly link to a remote server lib or download it manually. It should be bundled either as part of the dependencies (composer) or in the source itself. Since that's a vue component that requires a build process, it is better to build it externally and bundle as a package.

jacob-js:
If we were to use it, nothing would prevent Tiki from distributing the package. It's a Vue web component, so it doesn't depend on a Vue environment to execute.

kroky:
What do you mean Tiki to distribute it? How do we handle dependencies/upgrades? How do we link to the distribution package in Cypht?

jacob-js:
[vue-widgets/datetime-picker](https://gitlab.com/tikiwiki/tiki/-/tree/master/src/js/vue-widgets/datetime-picker?ref_type=heads) is a package which can be published to npm and accessible from there. Thus, dependencies and upgrades can be managed.

kroky:
https://gitlab.com/tikiwiki/tiki/-/tree/master/src/js/vue-widgets/datetime-picker?ref_type=heads is part of a source tree of Tiki in a particular branch. I would like to avoid bundling such packages into 3rd party software like Cypht as maintaining these in the future will be tricky - e.g. Cypht has multiple branches, each depend on a particular Tiki version of the package, someone updates something in the Tiki package - they don't even know that something external depends on it, etc.

If we bundle as a package, it should be composer. Cypht has no node dependencies and installing via npm is not something the current installer can do.

Why don't we bundle the vuedatetimepicker lib as a composer package and make cypht rely on that + versioning?

jacob-js:
I don't necessarily mean to fetch the package directly from Tiki. We could extract it from Tiki, bundle it as a standalone package, and then re-inject it into Tiki via npm and into Cypht via CDN. The component itself isn't restricted to Tiki domains. In fact, this applies to everything in https://gitlab.com/tikiwiki/tiki/-/tree/master/src/js/vue-widgets.

kroky:
Let's do it like this then, I like the idea!

josaphatim:
@jacob-js Can you make this happen ?

jacob-js:
> @jacob-js Can you make this happen ?

Yes, let's wait until Tiki 28 is released, then we'll make this happen.

josaphatim:
I think Cypht development must be independent from Tiki.
I think we can't wait a Tiki release to do something in Cypht.

marclaporte:
I don't know if it's a viable option here, but here is what we used for JS many dependencies before the new build system: https://asset-packagist.org/

josaphatim:
@mercihabam any update here ?

PR #1239, opened 2024-09-12
Enable keyboard shortcut selenium test

josaphatim:


josaphatim:
Pipeline will be fixed after [setting tests merged](https://github.com/cypht-org/cypht/pull/1193) as it depends on site settings page

Issue #1159, opened 2024-08-12
Preview of message when viewing message list


europacafe:
## 🚀 Feature
It would be nice if cypht has an option to turn on and off for "Preview" message. When turned on, allows an option to show 1 or 2 line of mail body when viewing the message list

### Design, Layout, Architecture
Here is the sample "preview" from yahoo webmail (highlighted in yellow)
![image](https://github.com/user-attachments/assets/8004889a-c36a-43d9-bbfd-d3c5a8b0d6d3)

marclaporte:
@europacafe 

I like it. And also, I'd like an option on mouse over. Sometimes, I want to skim through the emails for urgent things without marking them as read.

For the next 50-100 commits, we will focus on bug fixes and [paper cut bugs](https://en.wikipedia.org/wiki/Paper_cut_bug) but after, we'll come back to enhancements like this.

Please ping me if it's forgotten.

marclaporte:
@europacafe Cypht 2.4.0 will be released any day now and we solved many bugs, so we can now balance bug fixing and enhancements.

Thus, @christer77 has started working on your suggestion.

europacafe:
Thanks for the continuous enhancement. Looking forward to it.

christer77:
hello @europacafe

can you close this issue if resolved please.

marclaporte:
This will land in Cypht 2.5.0 (in coming weeks)

This was the commit:
https://github.com/cypht-org/cypht/pull/1241

PR #1241, opened 2024-09-13
[ENH]message list: Preview of message when viewing message list

christer77:
[Issue: Preview of message when viewing message list](https://github.com/cypht-org/cypht/issues/1159)

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -223,6 +223,11 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
             $msg['subject'] = '[No Subject]';
         }
         $subject = $msg['subject'];
+        $preview_msg = "";
+        if (isset($msg['preview_msg'])) {
+            $preview_msg = $msg['preview_msg'];
+        }
+       
         if ($parent_list == 'sent') {
             $icon = 'sent';
             $from = $msg['to'];
@@ -298,7 +303,7 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
             $res[$id] = message_list_row(array(
                     array('checkbox_callback', $id),
                     array('icon_callback', $flags),
-                    array('subject_callback', $subject, $url, $flags, $icon),
+                    array('subject_callback', $subject, $url, $flags, $icon, $preview_msg),
                     array('safe_output_callback', 'source', $source),
                     array('safe_output_callback', 'from'.$nofrom, $from, null, str_replace(array($from, '<', '>'), '', $msg['from'])),
                     array('date_callback', $date, $timestamp),
@@ -314,7 +319,7 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
                     array('checkbox_callback', $id),
                     array('safe_output_callback', 'source', $source, $icon),
                     array('safe_output_callback', 'from'.$nofrom, $from, null, str_replace(array($from, '<', '>'), '', $msg['from'])),
-                    array('subject_callback', $subject, $url, $flags),
+                    array('subject_callback', $subject, $url, $flags, null, $preview_msg),
                     array('date_callback', $date, $timestamp, $is_snoozed),
                     array('icon_callback', $flags)
                 ),
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -822,6 +822,7 @@ class Hm_Handler_imap_folder_page extends Hm_Handler_Module {
         $offset = 0;
         $msgs = array();
         $list_page = 1;
+        $include_preview = $this->user_config->get('active_preview_message_setting', false);
 
         list($success, $form) = $this->process_form(array('imap_server_id', 'folder'));
         if ($success) {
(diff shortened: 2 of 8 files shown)


modules/imap/handler_modules.php
```diff
@@ -852,6 +852,14 @@ public function process() {
...
                     $msg['server_name'] = $details['name'];
                     $msg['folder'] = $form['folder'];
+                    $uid = $msg['uid'];
+                    $part = true;
+                    $max = 87;
+                    $msg_struct = $imap->get_message_structure($uid);
+                    $struct = $imap->search_bodystructure($msg_struct, array('imap_part_number' => $part));
+                    $msg_struct_current = array_shift($struct);
+                    $msg_text = $imap->get_message_content($uid, $part, $max, $msg_struct_current);
+                    $msg['preview_msg'] = $msg_text;
```
Shadow243, line 862: It's displaying the `$subject` content, so why did you declare `$preview_msg` instead of just using `$subject`?

christer77: It's not a subject content, it is the content of the message in preview mode.

can you see this capture of this [issue](https://github.com/cypht-org/cypht/issues/1159#issue-2459881679)

Shadow243:
@christer77 take this comment: https://github.com/cypht-org/cypht/issues/1159#issuecomment-2297722189 into consideration too

christer77:
> @christer77 take this comment: [#1159 (comment)](https://github.com/cypht-org/cypht/issues/1159#issuecomment-2297722189) into consideration too

well noted

modules/imap/handler_modules.php
kroky, line 862: The way you do it has several problems:
1. Too many internal calls to imap method.
2. Too many calls to imap server for each message - it will be rather slow for bigger lists of messages. Might slow down the normal list pages by 2-4 times.
3. I think we should seriously do this conditionally based on a setting that user turns on and off for specific pages (imap folder pages, combined pages).
4. The code to get the preview should be moved out of here as this only works for single imap pages. We have many other places listing messages (search results, combined pages - all, sent, unread, etc.)
5. It is best to have a setting, pass it to the imap get_mailbox_page or get_message_list methods and have it query the text body as well if required. This will ensure we get all the content in a single call to the imap server. Please test this with 50 messages at once and you will notice the difference. Check if the IMAP protocol doesn't have something built-in - like getting part of the text body of the message or getting only the text body, not the whole structure. Parsing the bodystructure is an expensive operation. If we really need to do this for each message separately in list-view, we should consider aggressive caching.

modules/imap/functions.php
```diff
@@ -291,7 +296,7 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
             $res[$id] = message_list_row(array(
                     array('checkbox_callback', $id),
                     array('icon_callback', $flags),
-                    array('subject_callback', $subject, $url, $flags, $icon),
+                    array('subject_callback', array("subject" => $subject, "preview_msg" => $preview_msg), $url, $flags, $icon),
```
kroky, line 299: subject_callback is under your control - please add a separate parameter there for preview_msg instead of using subject and preview in an array - it is harder to understand what $vals[0] mean in the callback. Better use separate named parameters.

modules/imap/handler_modules.php
christer77: Noted

modules/imap/functions.php
christer77: Noted

modules/imap/handler_modules.php
```diff
@@ -849,9 +849,19 @@ public function process() {
...
                     $msg['folder'] = $form['folder'];
+                    // $uid = $msg['uid'];
+                    // $part = true;
+                    // $max = 87;
+                    // $msg_struct = $imap->get_message_structure($uid);
+                    // $struct = $imap->search_bodystructure($msg_struct, array('imap_part_number' => $part));
+                    // $msg_struct_current = array_shift($struct);
+                    // $msg_text = $imap->get_message_content($uid, $part, $max, $msg_struct_current);
+                    // // $msg['preview_msg'] = "xxxxxxx";
+                    // $msg['preview_msg'] = $msg_text;
```
kroky, line 864: Please remove debugging code.

modules/imap/hm-imap.php
```diff
@@ -829,6 +831,17 @@ public function get_message_list($uids, $raw=false) {
                                 }
                             }
                         }
+                        elseif ($vals[$i] == 'BODY[0.1') {
+                            $content = '';
+                            $i++;
+                            $i++;
```
kroky, line 837: Seems strange but you probably tested this? Why do you increment so many times?

modules/smtp/modules.php
```diff
@@ -2027,7 +2027,7 @@ function get_outbound_msg_detail($post, $draft, $body_type) {
     }
     if (array_key_exists('compose_in_reply_to', $post)) {
         $in_reply_to = $post['compose_in_reply_to'];
-        $draft['draft_in_reply_to'] = $post['compose_in_reply_to'];
+        $draft['draft_in_reply_to'] = $pzost['compose_in_reply_to'];
```
kroky, line 2030: Typo?

modules/imap/hm-imap.php
```diff
@@ -761,7 +761,7 @@ public function poll() {
          * @param bool $raw flag to disable decoding header values
          * @return array list of headers and values for the specified uids
          */
-        public function get_message_list($uids, $raw=false) {
+        public function get_message_list($uids, $raw=false, $include_preview = false) {
```
kroky, line 764: Implementation is much better than previous time.  Please respect "include_preview" option - i.e. don't fetch BODY below if preview is not requested. Add a user config setting to fetch previews - boolean - yes/no. Pass it to this method, so unless user wants previews, it won't fetch the body.

christer77: well noted

modules/imap/hm-imap.php
christer77: That's right, it's just to allow us to access the message index while our main "for" loop is iterating.

Issue #1068, opened 2024-06-05
Always have send button enabled on compose page


josaphatim:
## 🐛 Bugreport
There is a 30 second save interval in the draft page during which the form data is saved in draft. For people with slow internet connection, it takes too long and the send button is disabled during this process. It's frustrating waiting for it to finish to finally send the form.

### Version & Environment
All versions are concerned

### Steps to reproduce
1. Go to compose page
2. Wait 30 seconds
3. You will see the send button disabled

marclaporte:
@GedeonTS Can you reproduce this issue?

marclaporte:
Without an SMTP server configured, Cypht should know right away sending won't be possible (unless via JMAP or EWS)

@josaphatim please advise.

IrAlfred:
Hello @josaphatim 

This issue seems to have already been resolved. If you still experience it, please feel free to reopen the thread.

PR #1246, opened 2024-09-14
[FIX] Make the send button enabled on compose page

GedeonTS:
https://github.com/cypht-org/cypht/issues/1068

### DIFF
--- a/modules/smtp/site.js
+++ b/modules/smtp/site.js
@@ -524,7 +524,9 @@ $(function () {
             }
 
             function handleSendAnyway() {
-                smtp_delete_draft($('.compose_draft_id').val());
+                if($('.compose_draft_id').val() === 0){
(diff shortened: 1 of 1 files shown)


modules/smtp/site.js
```diff
@@ -137,8 +137,6 @@ var save_compose_state = function(no_files, notice) {
         return;
     }
 
-    $('.smtp_send_placeholder').prop('disabled', true);
-    $('.smtp_send_placeholder').addClass('disabled_input');
```
kroky, line 141: If you remove this feature, you should also take care of the "undisabling" code below...

Another thing to check - is it OK to submit the form while draft is saved as ajax request? What happens to the draft? I suppose it will be left on the IMAP server as compose_draft_id will be still empty when submitting the compose form. In that case, isn't it better to submit another ajax request before submitting the form - only if the previous one hasn't completed and make the draft be deleted?

PR #1251, opened 2024-09-22
[FIX]setup_database.php: Check if the required MySQL extensions are loaded

Baraka24:
Having: 
`
Attempting to connect to database ... (1/10)
Attempting to connect to database ... (2/10)
Attempting to connect to database ... (3/10)
Attempting to connect to database ... (4/10)
Attempting to connect to database ... (5/10)
Attempting to connect to database ... (6/10)
Attempting to connect to database ... (7/10)
Attempting to connect to database ... (8/10)
Attempting to connect to database ... (9/10)
Attempting to connect to database ... (10/10)
Unable to connect to database
`
while running `$ php ./scripts/setup_database.php results` on a fresh Ubuntu 22.04.

### DIFF
--- a/scripts/setup_database.php
+++ b/scripts/setup_database.php
@@ -27,6 +27,39 @@ $bad_driver = "Unsupported db driver: {$db_driver}";
 
 // NOTE: these sql commands could be db agnostic if we change the blobs to text
 
+// Check if the required extensions for the configured DB driver are loaded
+if ($db_driver == 'mysql') {
+    $required_extensions = ['mysqli', 'mysqlnd', 'pdo_mysql'];
+    $missing_extensions = [];
+
+    foreach ($required_extensions as $extension) {
+        if (!extension_loaded($extension)) {
+            $missing_extensions[] = $extension;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


Yannick243:
Hello @Baraka24 , can you add an if check depending on the `DB_DRIVER` configured ? 
For example, if someone prefers to use another driver like `postgres`, they should still see the error, even if they don’t want to use MySQL, as the code below is currently restricted to MySQL.

scripts/setup_database.php
```diff
@@ -27,6 +27,39 @@
 
 // NOTE: these sql commands could be db agnostic if we change the blobs to text
 
+// Check if the required extensions for the configured DB driver are loaded
+if ($db_driver == 'mysql') {
+    $required_extensions = ['mysqli', 'mysqlnd', 'pdo_mysql'];
```
josaphatim, line 32: @Baraka24 can you please create a function to remove duplicate codes share by these 2 if statements ?

Baraka24: [PR](https://github.com/cypht-org/cypht/pull/1390)

Issue #1144, opened 2024-08-04
Improve performance for most common operations (view next email, delete email, move email to another folder, etc.)


marclaporte:
As of Cypht 2.2.0, the feature set is getting pretty impressive.

However, performance is slower than Gmail or Roundcube. To illustrate, there is a video of deleting emails in Roundcube/Kolab here: https://github.com/cypht-org/cypht/issues/435

Let's explore how this could be improved.

jacob-js:
Performance is slow because each operation makes a separate request to the backend service and the IMAP server, even though this is unnecessary in most cases. The most important actions are fetching the message list and fetching the message content. These could be combined, so that once the message list is displayed, the user doesn't need to make another request to view the message content or move to the next message.

Also, actions such as `delete, copy, and move` should not reload the message list. The optimal solution would be to manage the message state via JavaScript. Instead of requesting a new list from the server, these actions would update the display by modifying the current state.

I'll explore this approach and see how fast it becomes.

jacob-js:
I avoided combining the two requests because it results in a longer request, causing the user to wait even longer for the list of messages to appear. Instead, I worked on https://github.com/cypht-org/cypht/pull/1252 which improves the message caching to prevent multiple fetches of the same message when requested and ensure that viewing the next email gets done instantly after a delete, move, or archive operation.
I'll leave the PR for review. Afterward, we can work on preventing some actions from reloading the page when performed, as this is another performance issue.

kroky:
#1252 looks good - just check if Tiki-master is OK with latest code - not sure about the paths of the new module import.

jacob-js:
> #1252 looks good - just check if Tiki-master is OK with latest code - not sure about the paths of the new module import.

Tiki-master isn't OK because those modules are not included in the final site.js. But, this PR (https://github.com/cypht-org/cypht/pull/1264) should fix.

kroky:
Alright, thanks!

PR #1252, opened 2024-09-24
Improve messages caching to ensure a message isn't fetched more than once when viewing its content

jacob-js:
https://github.com/cypht-org/cypht/issues/1144


modules/core/site.js
```diff
@@ -280,6 +280,14 @@ var Hm_Ajax_Request = function() { return {
                 for (var name in res.folder_status) {
                     Hm_Folders.unread_counts[name] = res.folder_status[name]['unseen'];
                     Hm_Folders.update_unread_counts();
+                    const messages = new Hm_MessagesStore(name, Hm_Utils.get_url_page_number());
```
josaphatim, line 283: Hello @jacob-js shouldn't we move the Hm_MessagesStore definition to core module ? Also on production I'm getting error
`site.js?v=somerandomchars:1282 Uncaught ReferenceError: Hm_MessagesStore is not defined`

jacob-js: It’s not a good idea to move it to core. The core module is being misused, which is why you even find a reference to `Hm_MessagesStore` there. Core should not handle displaying the messages list or the message content; it should only contain things that can’t go into any other modules or are used by many, if not all, modules.

That’s my 2¢, unless I misunderstand your reasoning behind introducing the core module.

jacob-js: The error in production should have been already addressed by https://github.com/cypht-org/cypht/pull/1264

kroky: It really depends on the usage. Originally, we find message functions in core module as there are several modules that display messages - imap, the obsolete and removed pop3, jmap (which is working on top of imap currently), feeds (which are again treated as kind of messages, so Cypht can aggregate and show everyhting and other aggregate folders containing multiple sources). Thus, this message store makes sense to be in core module but I am not sure it was designed to work with anything outside imap module and maybe that's why Merci thinks it shouldn't be moved. What happens if you add an rss feed to Everything/All menu entry and browse that? Is the Hm_MessagesStore handling that fine?

josaphatim: Hm_MessagesStore is called in core module. Once the imap module disabled it will raise an error as its definition won't be found. Also the places were it is used doesn't only concern the IMAP module

jacob-js: Victor just made a very important point: `Hm_MessagesStore` works only with IMAP, so it makes sense for it to stay there. However, the usage found in core is incorrect, as it will break with other source types. This raises the question of whether we should extend its capabilities to work with all sources, requiring a move to the core module, or have a separate store for each source.

kroky: It makes sense to extend it to work with other message types and move to core - since it handles the local storage cache, retrieving next items, etc. - these are actions relevant for all message types.

jacob-js: This is happening in https://github.com/cypht-org/cypht/pull/1266

PR #1254, opened 2024-09-25
[FIX] modules.php PHP Warning Undefined array key of redirect uri and…

amaninyumu1:


modules/wordpress/modules.php
```diff
@@ -243,8 +243,8 @@ class Hm_Handler_setup_wordpress_connect extends Hm_Handler_Module {
     public function process() {
         $details = wp_connect_details($this->config);
         if (!empty($details)) {
-            $oauth2 = new Hm_Oauth2($details['client_id'], $details['client_secret'], $details['redirect_uri']);
-            $this->out('wp_auth_url', $oauth2->request_authorization_url($details['auth_url'], 'global', 'wp_authorization'));
+            $oauth2 = new Hm_Oauth2($details['client_id'], $details['client_secret'], $details['redirect_uri'] ?? null);
```
josaphatim, line 246: @amaninyumu1 shouldn't this be **client_uri** ? As there is no redirect_uri in wordpress config file (Just wondering I'm not sure)

PR #1257, opened 2024-09-28
Added code to raise exception on database errors

### LINKED ISSUE #1017: Raise exceptions on error
## 🗣 Suggestion

This conversation started here: https://github.com/cypht-org/cypht/issues/977#issuecomment-2098904257

Concerning PDO error modes [php.watch/versions/8.0/PDO-default-errmode](https://php.watch/versions/8.0/PDO-default-errmode).

It appears php 7.x defaulted to make them silent, but 8.x is making them throw exceptions. This would have very different behaviors in the way cypht uses them since it is not explicitly setting a mode. So I recommend setting it explicitly.

That being said, I agree with the decision to have it default to throw exceptions. I came upon this when trying to figure out why some database operations were not working while I was developing. cypht seems to capture and hide the actual errors as you can see here:
https://github.com/cypht-org/cypht/blob/2cf7a0ae88215e9ebc10935a0910290cec17fc5d/scripts/create_account.php#L43-L50

Unless there is another way to expose these errors (perhaps via a debug mode log) you cant see what the issue is, which is needed in order to fix it.

In my testing I was able to turn the above into a single line:
```php
	$auth->create($user, $pass);
```

I explicitly ask for exceptions to happen when setting up the connection:
```php
self::$dbh[$key]->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
```

That way the (error) logs are actionable.

josaphatim:
### Issues
- fixes #1017

### DIFF
--- a/lib/db.php
+++ b/lib/db.php
@@ -94,17 +94,23 @@ class Hm_DB {
         if (!$type) {
             $type = self::execute_type($sql);
         }
-        $sql = $dbh->prepare($sql);
-        if (!$sql || !$sql->execute($args)) {
+        try {
+            $sql = $dbh->prepare($sql);
+            if (!$sql || !$sql->execute($args)) {
+                return false;
+            }
+            if ($type == 'modify' || $type == 'insert') {
+                return $sql->rowCount();
+            }
+            if ($all) {
+                return $sql->fetchAll(PDO::FETCH_ASSOC);
+            }
+            return $sql->fetch(PDO::FETCH_ASSOC);
+        } catch (PDOException $oops) {
+            Hm_Msgs::add('ERRDatabase error. Please try again.');
+            Hm_Debug::add($oops->getMessage());
             return false;
         }
-        if ($type == 'modify' || $type == 'insert') {
-            return $sql->rowCount();
-        }
-        if ($all) {
-            return $sql->fetchAll(PDO::FETCH_ASSOC);
-        }
-        return $sql->fetch(PDO::FETCH_ASSOC);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lib/db.php
```diff
@@ -139,6 +145,7 @@ static public function connect($site_config) {
         try {
             self::$dbh[$key] = new PDO($dsn, self::$config['db_user'], self::$config['db_pass']);
             self::$dbh[$key]->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
+            self::$dbh[$key]->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
```
Shadow243, line 148: If exceptions are not handled properly, detailed error messages could be displayed to users, exposing sensitive information like SQL queries or database structure. This can lead to security vulnerabilities.

lib/db.php
```diff
@@ -94,17 +94,23 @@ static public function execute($dbh, $sql, $args, $type = false, $all = false) {
...
+            if ($type == 'modify' || $type == 'insert') {
+                return $sql->rowCount();
+            }
+            if ($all) {
+                return $sql->fetchAll(PDO::FETCH_ASSOC);
+            }
+            return $sql->fetch(PDO::FETCH_ASSOC);
+        } catch (PDOException $oops) {
+            Hm_Msgs::add('ERRDatabase error. Please try again.');
+            Hm_Debug::add($oops->getMessage());
```
Shadow243, line 111: if we keep `self::$dbh[$key]->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);` we could then Log the error for later inspection like this: `error_log($e->getMessage());` or use `PDO::ERRMODE_WARNING`

This can be a security issue in production.

lib/db.php
kroky: That's correct but we still prefer working in exception mode and catch exceptions in relevant places. We have more flexibility this way and errors are not silently ignored.

PR #1265, opened 2024-10-02
fix(other): screen email: fix adding to trust contact while the email is blo…

christer77:
This PR resolves [issue's](https://github.com/cypht-org/cypht/issues/1129) points 4 and 5

### DIFF
--- a/modules/imap/site.js
+++ b/modules/imap/site.js
@@ -1356,19 +1356,80 @@ var add_email_in_contact_trusted = function(list_email) {
     }
 };
 
+
+function get_list_block_sieve() {
+    sessionStorage.removeItem('list_blocked');
+    var detail = Hm_Utils.parse_folder_path(hm_list_path());
+    var list_blocked_senders = [];
+    var page = hm_page_name();
+    if (page == 'message_list') {
+        Hm_Ajax.request(
+            [
+                { name: 'hm_ajax_hook', value: 'ajax_list_block_sieve' },
+                { name: 'imap_server_id', 'value': detail.server_id},
+            ],
+            function (res) {
+                if (res.ajax_list_block_sieve) {
+                    sessionStorage.setItem('list_blocked', res.ajax_list_block_sieve);
+                }
+            }
+        );
+    }
+};
+
 $('.screen-email-unlike').on("click", function() { imap_screen_email(); return false; });
 
 $('.screen-email-like').on("click", function() {
+    var list_blocked_senders = (sessionStorage.getItem('list_blocked') !== null) ? JSON.parse(sessionStorage.getItem('list_blocked')) : [];
     var list_email = [];
+    var list_msg_uid = [];
+    var email_existing_in_blocked_senders = [];
     $('input[type=checkbox]').each(function() {
         if (this.checked && this.id.search('imap') != -1) {
             let email = $('.'+ this.id +' .from').attr("data-title")
             if (email = email.trim()) {
                 list_email.push(email);
+                if (list_blocked_senders.length > 0) {
+                    list_blocked_senders.forEach((sender, index) => {
+                        if (sender === email) {
+                            email_existing_in_blocked_senders.push(email);
+                            list_msg_uid.push($(this).parent().parent().attr("data-uid"));
+                            delete list_blocked_senders[index];
+                        }
+                    });
+                }
             }
         }
     });
-    add_email_in_contact_trusted(list_email); return false;
+
+    if (email_existing_in_blocked_senders) {
+        var list_html = "<ol>";
+        email_existing_in_blocked_senders.forEach(sender => {
+            sender = sender.trim();
+            list_html += `<li>${sender}</li>`;
+        });
+        list_html += "</ol>";
+        const modal = new Hm_Modal({
+            modalId: 'emptySubjectBodyModal',
+            title: 'Warning',
+            btnSize: 'sm'
+        });
+
+        var modalContentHeadline = "Adress mail exist in your Block list";
+        modal.addFooterBtn(hm_trans('Add Emails to Trust contact'), 'btn-warning', handleAddEmail);
+        modal.setContent(modalContentHeadline + list_html + `<p>${hm_trans('If you add these, all will be unblocked.<br>Are you sure you want to add this in your Trust contact?')}</p>`);
+        modal.open();
+        function handleAddEmail() {
+            list_msg_uid.forEach(function(msg_uid) {
+                block_unblock_sender(msg_uid, Hm_Utils.parse_folder_path(hm_list_path()), 'sender', 'unblocked');
+            });
+            modal.hide();
+            add_email_in_contact_trusted(list_email);
+        };
+    } else {
+        add_email_in_contact_trusted(list_email);
+    }
+    return false;
 });
 
 $(document).on('click', '[data-bs-dismiss="modal"]', function() {
--- a/modules/sievefilters/functions.php
+++ b/modules/sievefilters/functions.php
@@ -284,6 +284,22 @@ if (!hm_exists('get_sieve_client_factory')) {
     }
 }
 
+if (!hm_exists('prepare_sieve_script ')) {
+    function prepare_sieve_script ($script, $index = 1, $action = "decode")
+    {
+        $blocked_list = [];
+        if ($script != '') {
+            $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[$index]);
+            if ($action == "decode") {
+                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)));
+            } else {
+                $blocked_list = json_encode(base64_decode($base64_obj));
+            }
+        }
+        return $blocked_list;
+    }
+}
+
 if (!hm_exists('get_domain')) {
     function get_domain($email)
     {
@@ -424,8 +440,7 @@ if (!hm_exists('get_blocked_senders_array')) {
             $blocked_senders = [];
             $current_script = $client->getScript('blocked_senders');
             if ($current_script != '') {
-                $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[1]);
-                $blocked_list = json_decode(base64_decode($base64_obj));
+                $blocked_list = prepare_sieve_script ($current_script);
                 if (!$blocked_list) {
                     return [];
                 }
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -30,10 +30,10 @@ class Hm_Handler_sieve_edit_filter extends Hm_Handler_Module {
         try {
             $client = $factory->init($this->user_config, $imap_account, $this->module_is_supported('nux'));
             $script = $client->getScript($this->request->post['sieve_script_name']);
-            $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[1]);
-            $this->out('conditions', json_encode(base64_decode($base64_obj)));
-            $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[2]);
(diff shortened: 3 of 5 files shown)


Shadow243:
@christer77 please rebase to fix failing tests.

modules/imap/site.js
```diff
@@ -1408,19 +1408,78 @@ var add_email_in_contact_trusted = function(list_email) {
...
+
+var list_blocked_senders = [];
+get_liste_block_sieve(Hm_Utils.parse_folder_path(hm_list_path()));
+
+function get_liste_block_sieve(detail) {
+    var page = hm_page_name();
+    if (page == 'message_list') {
+        Hm_Ajax.request(
+            [
+            { name: 'hm_ajax_hook', value: 'ajax_liste_block_sieve' },
```
Shadow243, line 1421: correct the indentation here

modules/sievefilters/modules.php
```diff
@@ -1344,3 +1344,62 @@ public function process() {
...
+    public function process() {
+        list($success, $form) = $this->process_form(array('imap_server_id'));
+
+        if (!$success) {
+            return;
+        }
+
+        foreach ($this->user_config->get('imap_servers') as $idx => $mailbox) {
+            if ($idx == $this->request->post['imap_server_id']) {
+                $imap_account = $mailbox;
```
Shadow243, line 1362: Add `break;` after this line to exit the loop once the matching server ID is found.

modules/sievefilters/modules.php
```diff
@@ -1344,3 +1344,62 @@ public function process() {
...
+            }
+        }
+
+        $factory = get_sieve_client_factory($this->config);
+        try {
+            $client = $factory->init($this->user_config, $imap_account);
```
Shadow243, line 1368: `$imap_account` might be undefined can you initialize it with `null` even ?

modules/sievefilters/setup.php
```diff
@@ -127,6 +134,7 @@
         'sieve_detail_display' => array(FILTER_UNSAFE_RAW, false),
         'imap_extensions_display' => array(FILTER_UNSAFE_RAW, false),
         'script_details' => array(FILTER_UNSAFE_RAW, FILTER_REQUIRE_ARRAY),
+        'ajax_liste_block_sieve' => array(FILTER_UNSAFE_RAW, FILTER_REQUIRE_ARRAY),
```
Shadow243, line 137: array(FILTER_UNSAFE_RAW, FILTER_REQUIRE_ARRAY) ??

modules/imap/site.js
```diff
@@ -1318,19 +1318,78 @@ var add_email_in_contact_trusted = function(list_email) {
...
 $('.screen-email-like').on("click", function() {
     var list_email = [];
+    var list_msg_uid = [];
+    var email_existing_in_blocked_senders = [];
     $('input[type=checkbox]').each(function() {
         if (this.checked && this.id.search('imap') != -1) {
             let email = $('.'+ this.id +' .from').attr("data-title")
             if (email = email.trim()) {
                 list_email.push(email);
+                if (list_blocked_senders) {
```
Shadow243, line 1352: Can you do `if(list_blocked_senders.lenght > 0)' as it's a array

christer77: okay

modules/imap/site.js
```diff
@@ -1318,19 +1318,78 @@ var add_email_in_contact_trusted = function(list_email) {
...
     var list_email = [];
+    var list_msg_uid = [];
+    var email_existing_in_blocked_senders = [];
     $('input[type=checkbox]').each(function() {
         if (this.checked && this.id.search('imap') != -1) {
             let email = $('.'+ this.id +' .from').attr("data-title")
             if (email = email.trim()) {
                 list_email.push(email);
+                if (list_blocked_senders.lenght > 0) {
+                    Object.keys(list_blocked_senders).forEach(key => {
```
Shadow243, line 1353: Since list_blocked_senders is an array (not an object), using Object.keys is unnecessary. Instead, you can iterate directly over the array.
```
list_blocked_senders.forEach((sender, index) => {
        if (sender === email) {
            //your code here
        }
    });
    ```

File: modules/sievefilters/modules.php [hunk 7]
```diff
@@ -1344,46 +1346,3 @@ public function process() {
         }
     }
 }
-
-/**
- * @subpackage sievefilterstoggle/handler
- */
-class Hm_Handler_sieve_toggle_script_state extends Hm_Handler_Module {
```
Shadow243, line 1351: You are removing this but i can't see the corresponding line removed in setup.php

Shadow243 (changes requested):
Some updates to be done.

modules/imap/site.js
```diff
@@ -1301,31 +1301,90 @@ var imap_screen_email = function() {
...
-      Hm_Ajax.request(
-        [
-          { name: 'hm_ajax_hook', value: 'ajax_add_contact' },
-          { name: 'email_address', value: list_email.join(',') },
-        ],
-        function (res) {
-          window.location.reload();
-        }
-      );
+        Hm_Ajax.request(
```
Shadow243, line 1304: What did you change here, seams to be indentation. make sure you remove changes like this.

christer77: okay

Shadow243: @christer77 Please check this image to see what I’m seeing on my side. No changes were made, except for a space you added before `Hm_Ajax.request(` on the line following `if (list_email) {`.
Screenshot 2024-11-28 at 21 26 15

modules/sievefilters/setup.php
```diff
@@ -124,6 +129,8 @@
         'ajax_sieve_block_domain',
         'ajax_sieve_block_change_behaviour',
         'ajax_sieve_toggle_script_state',
+        'ajax_liste_block_sieve',
```
marclaporte, line 132: liste -> list

modules/imap/site.js
```diff
@@ -1313,19 +1313,78 @@ var add_email_in_contact_trusted = function(list_email) {
     }
 };
 
+var list_blocked_senders = [];
+get_liste_block_sieve(Hm_Utils.parse_folder_path(hm_list_path()));
```
marclaporte, line 1317: liste -> list

please grep the whole code base to correct

modules/sievefilters/setup.php
christer77: hello marc; I did not grasp the preoccupation. What is the request?

marclaporte: liste is a word in French.

We want source code in English: https://en.wiktionary.org/wiki/list

modules/sievefilters/modules.php
```diff
@@ -1379,11 +1379,67 @@ public function process() {
...
-            
+
             Hm_Msgs::add("Script $state");
         } catch (Exception $e) {
             Hm_Msgs::add("ERRSieve: {$e->getMessage()}");
         }
         $this->out('success', $success);
     }
 }
+class Hm_Handler_liste_block_sieve_script extends Hm_Handler_Module {
```
marclaporte, line 1390: Still some more liste -> list

Please use something like https://en.wikipedia.org/wiki/Grep

modules/sievefilters/modules.php
```diff
@@ -1379,11 +1379,67 @@ public function process() {
...
     }
 }
+class Hm_Handler_list_block_sieve_script extends Hm_Handler_Module {
+    public function process() {
+        list($success, $form) = $this->process_form(array('imap_server_id'));
+        if (!$success) {
+            return;
+        }
+
+        foreach ($this->user_config->get('imap_servers') as $idx => $mailbox) {
```
kroky, line 1397: Accessing internal representation of user config in unrelated modules is not recommended. Better use Hm_Imap_List::get() method to get the details by server ID.

modules/sievefilters/modules.php
```diff
@@ -1379,11 +1379,67 @@ public function process() {
...
+                $client->putScript(
+                    'blocked_senders',
+                    ''
+                );
+            }
+
+            $blocked_senders = [];
+            $current_script = $client->getScript('blocked_senders');
+            if ($current_script != '') {
+                $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $current_script, 0)[1]);
```
kroky, line 1424: Can you abstract the code pertaining to sieve filters retrieval parts of it and replacing/saving parts of it into a separate method/function/class? I believe a lot of these base64/json encoding/decoding calls are repeated in this modules file and can be abstracted.

modules/sievefilters/modules.php
```diff
@@ -1379,11 +1379,67 @@ public function process() {
...
+            $email_sender = str_replace('*', '', $email_sender);
+        }
+
+        $factory = get_sieve_client_factory($this->config);
+        try {
+            $client = $factory->init($this->user_config, $imap_account);
+            $scripts = $client->listScripts();
+
+            if(array_search('blocked_senders', $scripts, true) === false) {
+                $client->putScript(
```
kroky, line 1415: Why do you put an empty script here when you are going to check for non-empty script later in the code?

modules/imap/site.js
```diff
@@ -1313,19 +1313,78 @@ var add_email_in_contact_trusted = function(list_email) {
     }
 };
 
+var list_blocked_senders = [];
+get_list_block_sieve(Hm_Utils.parse_folder_path(hm_list_path()));
```
kroky, line 1317: I believe recent master has a dispatch/routing module that has other ways to execute code on page load and not on file load like you did it here. If in doubt, ask @mercihabam about it.

modules/sievefilters/modules.php
christer77: noted

modules/sievefilters/modules.php
kroky: This doesn't seem resolved.

modules/sievefilters/functions.php
```diff
@@ -284,6 +284,18 @@ function get_sieve_client_factory($site_config)
     }
 }
 
+if (!hm_exists('decode_blocked_senders_script')) {
+    function decode_blocked_senders_script($script)
```
kroky, line 288: This is nice, thanks! However, I still see 6 occurrences of base64_decode in sievefilters/modules.php - they should all be abstracted. Try to think about common operations that are repeated and abstract them in functions. Even better would be to have a meaningful functions doing one action at a time and calling others when necessary. I see 4 operations that are repeatedly doing the same thing - getting a script from cypht, decoding it, encoding the contents back and saving it back to cypht.

modules/sievefilters/functions.php
```diff
@@ -284,6 +284,22 @@ function get_sieve_client_factory($site_config)
     }
 }
 
+if (!hm_exists('encode_or_decode_blocked_senders_script')) {
+    function encode_or_decode_blocked_senders_script($script, $index = 1, $action = "decode")
```
kroky, line 288: That's better. Can you just rename the method to be more consistent with its usage? Something like prepare_sieve_script and action could be read/write or direction=straight/reverse or something more obvious. encode and decode is not good here as it has twofold meaning - base64 decoding but json encoding or base64 encoding but json decoding.

modules/imap/site.js
```diff
@@ -1356,19 +1356,80 @@ var add_email_in_contact_trusted = function(list_email) {
...
+                block_unblock_sender(msg_uid, Hm_Utils.parse_folder_path(hm_list_path()), 'sender', 'unblocked');
+            });
+            modal.hide();
+            add_email_in_contact_trusted(list_email);
+        };
+    } else {
+        add_email_in_contact_trusted(list_email);
+    }
+    return false;
 });
```
mercihabam, line 1433: @christer77  This should not be called when the module "sievefilters" is disabled. Please move it to its appropriate module.

Issue #1115, opened 2024-07-09
Improve detection and protection against business email compromise (BEC) like CEO Fraud, etc.


marclaporte:
"The second form is [name and email spoofing](https://www.barracuda.com/support/glossary/email-spoofing), where the attacker uses both the CEO’s name and their correct sender address. In this form of the attack, the attacker typically uses a reply-to address that is different than the sender address, so that your response to the email will go to them."
Source: https://www.barracuda.com/support/glossary/ceo-fraud

We expose an alternate "reply to" here https://github.com/cypht-org/cypht/pull/781 but we can do better. Related: https://github.com/cypht-org/cypht/issues/1113

See also: https://www.fbi.gov/how-we-can-help-you/scams-and-safety/common-scams-and-crimes/business-email-compromise

Danelif:
I found this [https://www.egress.com/blog/phishing/how-can-i-stop-ceo-fraud](https://www.egress.com/blog/phishing/how-can-i-stop-ceo-fraud) interresting

marclaporte:
@Danelif What is your next step here?

Danelif:
@marclaporte After meeting with @christer77 we discussed that cypht might check the emails box to compare a suspicious incoming email with previous emails and if a correspondence of 90% is found, let's turn this to spam. For example, if I receive an email from d****@evaludata.com and in my email record I have d***@evoludata.com this mean the incoming email is fake and we spam it directly

marclaporte:
Ok, please explore this.

We already have a feature to screen emails: https://github.com/cypht-org/cypht/issues/1113 so this addresses part of the issue.

Please keep in mind that mail servers often have other methods to filter spam (typically with content).

christer77:
Hello @marclaporte 

What's next, please?

marclaporte:
Thank you @christer77 for https://github.com/cypht-org/cypht/pull/1269

PR #1269, opened 2024-10-04
feat: Improved detection and protection against business email compromise (BEC) such as CEO fraud

christer77:
[issue: Improve detection and protection against business email compromise (BEC) like CEO Fraud, etc.
](https://github.com/cypht-org/cypht/issues/1115)

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -227,7 +227,7 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
         if (isset($msg['preview_msg'])) {
             $preview_msg = $msg['preview_msg'];
         }
-       
+    
         if ($parent_list == 'sent') {
             $icon = 'sent';
             $from = $msg['to'];
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -814,7 +814,12 @@ class Hm_Handler_imap_folder_page extends Hm_Handler_Module {
         $offset = 0;
         $msgs = array();
         $list_page = 1;
+        $include_content_body = false;
         $include_preview = $this->user_config->get('active_preview_message_setting', false);
+        $ceo_use_detect_ceo_fraud = $this->user_config->get('ceo_use_detect_ceo_fraud_setting', false);
+        if ($include_preview || $ceo_use_detect_ceo_fraud) {
+            $include_content_body = true;
+        }
 
         list($success, $form) = $this->process_form(array('imap_server_id', 'folder'));
         if ($success) {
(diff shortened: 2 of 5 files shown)


modules/imap/handler_modules.php
```diff
@@ -854,7 +854,21 @@ public function process() {
...
+                    $part = true;
+                    $max = 87;
+                    $msg_struct = $imap->get_message_structure($uid);
+                    $struct = $imap->search_bodystructure($msg_struct, array('imap_part_number' => $part));
+                    $msg_struct_current = array_shift($struct);
+                    $msg_text = $imap->get_message_content($uid, $part, $max, $msg_struct_current);
+                    // remove line 858 to 863 when https://github.com/cypht-org/cypht/pull/1241 is merged
+                    
+                    $isValide = $this->validationMessage($msg['to'], $msg['subject'], $msg_text, $imap, $this->cache, $form['imap_server_id']);
+                    if ($isValide) {
```
marclaporte, line 867: English: valid
French: valide

modules/imap/handler_modules.php
```diff
@@ -868,6 +882,72 @@ public function process() {
             $this->out('do_not_flag_as_read_on_open', $this->user_config->get('unread_on_open_setting', false));
         }
     }
+    public function validationMessage($email, $subject, $msg, $imap, $cache, $imap_server_id) {          
+        // 1. Check Suspicious Phrases or Requests
```
marclaporte, line 886: Phrase in French is a false friend to English phrase:
https://en.wiktionary.org/wiki/phrase

"phrase" in French translates "sentence" in English.

Since we may not be checking full sentences, you may want to use "suspicious terms"

christer77: Exactly

modules/imap/handler_modules.php
```diff
@@ -840,16 +845,28 @@ public function process() {
...
                 } else {
-                    list($total, $results) = $imap->get_mailbox_page(hex2bin($form['folder']), $sort, $rev, $filter, $offset, $limit, $keyword, null, $include_preview);
+                    list($total, $results) = $imap->get_mailbox_page(hex2bin($form['folder']), $sort, $rev, $filter, $offset, $limit, $keyword, null, $include_content_body);
                 }
                 foreach ($results as $msg) {
                     $msg['server_id'] = $form['imap_server_id'];
                     $msg['server_name'] = $details['name'];
                     $msg['folder'] = $form['folder'];
+                    $uid = $msg['uid'];
+                    // echo "<pre>"; var_dump("cxxxxxxxxxxxxxxxx, ", $msg); die;
```
marclaporte, line 857: Please clean

modules/imap/handler_modules.php
```diff
@@ -862,6 +879,80 @@ public function process() {
...
+        }
+        return false;
+    }
+    private function extractAmountFromEmail($emailBody) {
+        $pattern = '/\b\d+(?:,\d+)?\.?\d*\s*(?:USD|dollars?|US\$?|EUR|euros?|€|JPY|yen|¥|GBP|pounds?|£|CAD|CAD\$|AUD|AUD\$)/i';
+    
+        preg_match_all($pattern, $emailBody, $matches);
+    
+        if (count($matches[0]) > 0) {
+            return $matches[0][0];
```
kroky, line 953: Shouldn't you look/test all the amounts found, not only the first one?

modules/imap/handler_modules.php
```diff
@@ -840,16 +845,28 @@ public function process() {
...
+                    list($total, $results) = $imap->get_mailbox_page(hex2bin($form['folder']), $sort, $rev, $filter, $offset, $limit, $keyword, null, $include_content_body);
                 }
                 foreach ($results as $msg) {
                     $msg['server_id'] = $form['imap_server_id'];
                     $msg['server_name'] = $details['name'];
                     $msg['folder'] = $form['folder'];
+                    $uid = $msg['uid'];
+
+                    if ($ceo_use_detect_ceo_fraud) {
+                        $isValid = $this->validationMessage($msg['to'], $uid, $form['folder'], $msg['subject'], $msg['preview_msg'], $imap, $this->cache, $form['imap_server_id']);
```
kroky, line 859: Use one method for one job. If this is the validation method, then return the result of the validation and then decide if you want to move that email or not. This way, you won't have to pass imap, cache and other unrelated variables to the validation method.

modules/imap/handler_modules.php
```diff
@@ -840,16 +845,28 @@ public function process() {
...
+                        if (! $isValid) {
+                            $msg = [];
```
kroky, line 861: You will have to change the total as well if you silently remove messages from the search results.

modules/imap/handler_modules.php
```diff
@@ -862,6 +879,80 @@ public function process() {
             $this->out('do_not_flag_as_read_on_open', $this->user_config->get('unread_on_open_setting', false));
         }
     }
+    public function validationMessage($email, $uid, $current_folder, $subject, $msg, $imap, $cache, $imap_server_id,) {
```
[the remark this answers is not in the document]
christer77: Well noted

modules/imap/handler_modules.php
```diff
@@ -840,16 +845,40 @@ public function process() {
...
+                            );
+                            imap_move_same_server($server_ids, "move", $this->cache, [null, null, $dest_folder]);
+                            $msg = [];
+                            $total--;
+                        }
+                    }
+                    if (! $include_preview && isset($msg['preview_msg'])) {
+                        $msg['preview_msg'] = "";
+                    }
                     $msgs[] = $msg;
```
kroky, line 879: Please refrain from adding empty messages to the list. If you skip the message, simply don't add it to the list and update the total to match. Then, we won't need to change calling code to check if the msg is empty or not.

modules/imap/functions.php
```diff
@@ -212,122 +212,121 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
...
-        }
-        $is_snoozed = !empty($msg['x_snoozed']) && hex2bin($msg['folder']) == 'Snoozed';
-        if ($is_snoozed) {
-            $snooze_header = parse_snooze_header('X-Snoozed: '.$msg['x_snoozed']);
-            $date = $snooze_header['until'];
-            $timestamp = strtotime($date);
-        } else {
-            if ($list_sort == 'date') {
-                $date_field = 'date';
+        if ($msg) {
```
kroky, line 215: Please return this function to the previous state - we shouldn't have empty messages in the list.

modules/imap/handler_modules.php
christer77: Oh yes!

thanks.

modules/imap/functions.php
```diff
@@ -295,10 +295,6 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
             $icon = false;
         }
 
-        //if (in_array('draft', $flags)) {
-        //    $url = '?page=compose&list_path='.sprintf('imap_%s_%s', $msg['server_id'], $msg['folder']).'&uid='.$msg['uid'].'&imap_draft=1';
-        //}
```
kroky, line 300: This seems some debugging code. @christer77 can you please review the diff thoroughly again, make sure you commit only what you want to commit and let me know to do a final review here?

christer77: okay

Issue #763, opened 2023-09-11
Expose IMAP capabilities for sharing folders, and delegating responsibility for an account (ex.: administrative assistant)


marclaporte:
https://doc.dovecot.org/configuration_manual/shared_mailboxes/
https://docs.iredmail.org/mailbox.sharing.html

marclaporte:
Please use https://www.virtualmin.com/documentation/email/ as a reference implementation.

marclaporte:
The feature is done. Documentation is next.

PR #1270, opened 2024-10-04
[ENH] Implement Expose IMAP capabilities for sharing folders

Shadow243:
Related issue: https://github.com/cypht-org/cypht/issues/763
Screenshot 2024-10-10 at 03 11 09


modules/core/output_modules.php
```diff
@@ -1645,7 +1645,99 @@ class Hm_Output_content_section_end extends Hm_Output_Module {
      * Closes the main tag opened in Hm_Output_content_section_start
      */
     protected function output() {
-        return '</div></main>';
+        $share_folder_modal = '<div class="modal fade" id="shareFolderModal" tabindex="-1" aria-labelledby="shareFolderModalLabel" aria-hidden="true">';
```
kroky, line 1648: This class is a section_end - please put all that modal code in one or more separate classes and include those in the setup.php file where appropriate.

kroky:
All the rest looks good, thanks!

modules/core/site.js
```diff
@@ -58,6 +58,18 @@ $.fn.fadeOutAndRemove = function(timeout = 600) {
     return this;
 };
 
+$.fn.modal = function(action) {
```
josaphatim, line 61: Hello @Shadow243 
Shouldn't we mix this with Hm_Modal ?

Shadow243: I thought about this idea then I actually left modal to reflect the behavior of bootstrap especially since it is an implementation of Bootstrap modal

Shadow243: @josaphatim  If I use JS to generate dynamically the modal I will then not be able to write phpunit tests, So I favored the code that I can easily test.

modules/core/output_modules.php
Shadow243: new class added. @kroky i hope this PR is ready.

kroky (approved):

marclaporte:
I look forward to seeing how this behaves on various platforms, especially Dovecot, Cyrus and Stalwart.

PR #1277, opened 2024-10-08
refactor(other): promote JMAP & EWS in the UI

Bienvenumoringa:
[Related task](https://avan.tech/item112643)

Modify the text in the help section to explain that Cypht supports the JMAP (RFC8621) protocol in addition to IMAP and SMTP, and make JMAP available by default in standalone Cypht installations or those integrated with Tiki.

image

### DIFF
--- a/modules/nux/modules.php
+++ b/modules/nux/modules.php
@@ -543,7 +543,7 @@ class Hm_Output_nux_dev_news extends Hm_Output_Module {
 class Hm_Output_nux_help extends Hm_Output_Module {
     protected function output() {
         return '<div class="nux_help mt-3 col-lg-6 col-md-12 col-sm-12"><div class="card"><div class="card-body"><div class="card_title"><h4>'.$this->trans('Help').'</h4></div>'.
-            $this->trans('Cypht is a webmail program. You can use it to access your E-mail accounts from any service that offers IMAP, or SMTP access - which most do.').' '.
+            $this->trans('Cypht is a webmail aggregator client. You can use it to access your E-mail accounts from any server or service that offers any of the main protocols: IMAP/SMTP, Exchange Web Services (EWS) or the newest protocol: JMAP (RFC8621).').' '.
         '</div></div></div>';
     }
 }


modules/nux/modules.php
```diff
@@ -80,6 +81,9 @@ public function process() {
         if (data_source_available($modules, 'smtp')) {
             $smtp_servers = count(Hm_SMTP_List::dump(false));
         }
+        if (data_source_available($modules, 'jmap')) {
+            $jmap_servers = count(Hm_IMAP_List::dump(false));
```
Shadow243, line 85: Here you will get imap servers to, i mean it will combine imap ad jmap. If soo then we don't need to add an other link for jmap.

Shadow243: With the first idea it was already well started, it was rather necessary to apply the filter here: https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L64 without forgetting in this case to do it for this too: https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L75

Shadow243:
Screenshot 2024-10-22 at 15 49 52
You don't to repeat it after SMTP. it's already okay after IMAP as it's an` imap` alternative

Bienvenumoringa:
> Screenshot 2024-10-22 at 15 49 52
> You don't need to repeat it after SMTP. it's already okay after IMAP as it's an` imap` alternative, like this: Screenshot 2024-10-22 at 19 07 46
> 
> you can use a cndition like this:
> 
> ```
> $res .= sprintf(
>                         $this->trans('You have %d %s%s'),
>                         $server_data[$proto],
>                         mb_strtoupper($proto_dsp),
>                         ($proto === 'imap' ? ' / JMAP ' . $server_data['jmap'] . ($server_data[$proto] > 1 ? ' sources' : ' source') : ($server_data[$proto] > 1 ? ' sources' : ' source'))
>                     );
> ```
> 
> and filter from here:
> 
> ```
> if (data_source_available($modules, 'imap')) {
>             $imap_servers = count(array_filter(Hm_IMAP_List::dump(false), function($v) { return !array_key_exists('type', $v);}));
>             $jmap_servers = count(array_filter(Hm_IMAP_List::dump(false), function($v) { return array_key_exists('type', $v) && $v['type'] == 'jmap'; }));
>         }
> ```
> 
> ```
> $this->out('nux_server_setup', array(
>             'imap' => $imap_servers,
>             'jmap' => $jmap_servers,
>             'feeds' => $feed_servers,
>             'smtp' => $smtp_servers,
>             'profiles' => $profiles
>         ));
> ```

apply changers

modules/nux/modules.php
```diff
@@ -68,11 +68,13 @@ public function process() {
         $smtp_servers = NULL;
         $feed_servers = NULL;
         $profiles = NULL;
+        $jmap_servers = NULL;
 
         $modules = $this->config->get_modules();
 
         if (data_source_available($modules, 'imap')) {
-            $imap_servers = count(Hm_IMAP_List::dump(false));
+            $imap_servers = count(array_filter(Hm_IMAP_List::dump(false), function($v) { return !array_key_exists('type', $v);}));
```
Shadow243, line 76: To get imap you can just do:
'Hm_IMAP_List::dump(false) - $jmap_servers' instead of filtering again

modules/nux/modules.php
```diff
@@ -592,12 +597,12 @@ protected function output() {
                 $res .= sprintf(' <a href="?page=servers#%s_section">%s</a>', $proto, $this->trans('Add'));
             }
             else {
-                if ($server_data[$proto] > 1) {
```
Shadow243, line 595: You will need this if check to support the s for exemple source and sources

Shadow243 (changes requested):
One more change and we are done

modules/nux/modules.php
Bienvenumoringa:
```
It is already resolved here ($server_data[$proto] > 1 ? 'sources' : 'source') and here ($server_data[$proto] > 1 ? 'sources' : 'source').
```

Shadow243: It should be: You have 1 IMAP source(sources if many) | You have 1 JMAP source(sources if many)
Screenshot 2024-11-04 at 07 18 35

Bienvenumoringa: image

Shadow243:
Check your logs, there is an Fatal error `PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, int given in`

Bienvenumoringa:
> Check your logs, there is an Fatal error `PHP Fatal error: Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, int given in`

It okay @Shadow243

Shadow243:
> It okay @Shadow243

There is a calculation error, try locally especially with several servers connected and correct, then ping me

Shadow243:
We wil no longer need to add at sestion `Welcome to Cypht` because it will come with https://github.com/cypht-org/cypht/pull/1278 but we can update sestion `Help` may be add `EWS` too

kroky:
Oops, yes, #1278 already included separation of imap, jmap, ews and smtp servers, so no need to do it here.

Shadow243:
@marclaporte I added `EWS` too

marclaporte:
Here is the sentence now:
Cypht is a webmail program. You can use it to access your E-mail accounts from any service that offers IMAP, or SMTP access - which most do. And Cypht also supports the newest protocol: JMAP (RFC8621) and EWS.').

 It is a little convoluted and some may not know what EWS.

I propose:

Cypht is a webmail aggregator client. You can use it to access your E-mail accounts from any server or service that offers any of the main protocols: IMAP/SMTP, Exchange Web Services (EWS) or the newest protocol: JMAP (RFC8621).

PR #1284, opened 2024-10-16
Disable duplicate redis extension from Github actions

Shadow243:
Getting error `PHP Warning:  Module "redis" is already loaded in Unknown on line 0` at: https://github.com/cypht-org/cypht/actions/runs/11374213670/job/31642437449?pr=1275#step:8:13

I added those 2 lines to make sure we don't have any `redis` before we install at step: `Set up PHP`

### DIFF
--- a/.github/workflows/Test-Build.yml
+++ b/.github/workflows/Test-Build.yml
@@ -69,6 +69,7 @@ jobs:
           tools: phpunit, composer
           ini-values: cgi.fix_pathinfo=1
         env:
+          update: true
           fail-fast: true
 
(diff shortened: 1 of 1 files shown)


Shadow243:
> @Shadow243 this is an upstream issue described here: [shivammathur/setup-php#874](https://github.com/shivammathur/setup-php/issues/874) - please apply the proper fix they suggest.

Copy that @kroky , changes at: https://github.com/shivammathur/setup-php/issues/874#issuecomment-2418027500 applied

PR #1288, opened 2024-10-19
Fix some page handlers issues

jacob-js:


kroky:
Looks good but can you fix the failing test as well?

jacob-js:
> Looks good but can you fix the failing test as well?

Sure, I'm looking into it.

PR #1266, opened 2024-10-02
Implement browser history API based navigation, preventing page reloads between routings

jacob-js:
https://github.com/cypht-org/cypht/issues/1144

Checklist:
- [x] The sidebar item's active state updates when navigating
- [x] Ensure handlers for refresh links are applied so the browser doesn’t reload on click
- [x] Sorting and filtering work
- [x] Fix combined folder content not being loaded upon navigation
- [x] Can display flagged contents
- [x] Cronjob to reload active folder's content in the background
- [x] Refresh the list during navigation; at least in the background.
- [x] FIX: Navigating to a route, the data from a previous route gets fetched.
- [x] Fix unit tests relative to the js files output
- [x] ~~When clicking on a message link while contents are still being fetched in the background, the requests should be aborted so the navigation is performed faster~~ (Aborting the requests on the client doesn't help because, once the requests are sent, the server processes them synchronously, blocking other eventual requests. A solution could be processing the requests server-side asynchronously.)
- [x] Can requests be processed asynchronously? Let's keep track of this here: https://github.com/cypht-org/cypht/issues/1279

PR #1289, opened 2024-10-19
Correct imap buttons on servers page & trigger change sieve_mode_tls checkbox on edit server

Shadow243:
Follow up to https://github.com/cypht-org/cypht/pull/1199 & https://github.com/cypht-org/cypht/pull/1266

### DIFF
--- a/modules/core/js_modules/route_handlers.js
+++ b/modules/core/js_modules/route_handlers.js
@@ -17,6 +17,7 @@ function applyServersPageHandlers() {
     if (window.githubServersPageHandler) githubServersPageHandler();
     if (window.nasaServersPageHandler) nasaServersPageHandler();
     if (window.smtpServersPageHandler) smtpServersPageHandler();
+    if (window.imapServersPageHandler) imapServersPageHandler();
     if (window.wpServersPageHandler) wpServersPageHandler();
 }
 
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -1901,6 +1901,9 @@ var fillImapData = function(details) {
     if (details.sieve_config_host) {
         $('#srv_setup_stepper_imap_sieve_host').val(details.sieve_config_host);
         $("#srv_setup_stepper_enable_sieve").trigger("click", false);
+        $('#srv_setup_stepper_imap_sieve_mode_tls')
+                            .prop('checked', details.tls)
+                            .trigger('change');
     }
 
     if(details.tls) {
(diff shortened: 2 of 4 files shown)


jacob-js:
@Shadow243 I am skeptical as to why the pipeline succeeded here while containing mistakes to the test, showing up in follow-up pull requests. Does the pipeline run on cached content?

Shadow243:
> @Shadow243 I am skeptical as to why the pipeline succeeded here while containing mistakes to the test, showing up in follow-up pull requests. Does the pipeline run on cached content?

It's weird, this test doesn't work sometimes, I even opened another PR https://github.com/cypht-org/cypht/pull/1290 to fix this, with the same fix you applied my test didn't pass. let's see if it won't break anymore.

jacob-js:
> > @Shadow243 I am skeptical as to why the pipeline succeeded here while containing mistakes to the test, showing up in follow-up pull requests. Does the pipeline run on cached content?
> 
> It's weird, this test doesn't work sometimes, I even opened another PR #1290 to fix this, with the same fix you applied my test didn't pass. let's see if it won't break anymore.

As for the test, it is stable. Reintroducing the line `self.wait_for_navigation_to_complete()` worked for you as well, but you encountered a separate test failure that was fixed with this [change](https://github.com/cypht-org/cypht/pull/1288/commits/afbaff02a1b581b810234ba1e275f05faddaeaf5#diff-015513616f5a3e387c6d1e06e82a29d9be33debc8de739fcf998719ab2d62280R87).

It is still confusing that the pipeline sometimes runs against old content (seemingly from the master branch) rather than the actual branch state.

Shadow243:
> > > @Shadow243 I am skeptical as to why the pipeline succeeded here while containing mistakes to the test, showing up in follow-up pull requests. Does the pipeline run on cached content?
> > 
> > 
> > It's weird, this test doesn't work sometimes, I even opened another PR #1290 to fix this, with the same fix you applied my test didn't pass. let's see if it won't break anymore.
> 
> As for the test, it is stable. Reintroducing the line `self.wait_for_navigation_to_complete()` worked for you as well, but you encountered a separate test failure that was fixed with this [change](https://github.com/cypht-org/cypht/pull/1288/commits/afbaff02a1b581b810234ba1e275f05faddaeaf5#diff-015513616f5a3e387c6d1e06e82a29d9be33debc8de739fcf998719ab2d62280R87).
> 
> It is still confusing that the pipeline sometimes runs against old content (seemingly from the master branch) rather than the actual branch state.

Super, i then close #1290.

PR #1290, opened 2024-10-19
Correct folders selenium test

Shadow243:
follow up to https://github.com/cypht-org/cypht/pull/1289


Shadow243:
https://github.com/cypht-org/cypht/blob/master/tests/selenium/pages.py#L122 is failing sometimes.

PR #1292, opened 2024-10-21
[FIX]: Fixed the duplicate close button for inline messages and ensured it works properly for non-inline messages

GedeonTS:
Screenshot from 2024-10-21 09-03-57

### DIFF
--- a/modules/imap/site.js
+++ b/modules/imap/site.js
@@ -778,8 +778,7 @@ var imap_setup_message_view_page = function(uid, details, list_path, callback) {
                         .append(msgResponse.msg_text)
                         .append(msgResponse.msg_parts);
         document.title = $('.header_subject th').text();
-        $('.header_subject th').append('<i class="bi bi-x-lg close_inline_msg"></i>');
-        $('.close_inline_msg').on("click", function() { msg_inline_close(); });
+
 
         $('.reply_link, .reply_all_link, .forward_link').each(function() {
             $(this).data("href", $(this).attr("href")).removeAttr("href");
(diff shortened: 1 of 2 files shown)


modules/imap/site.js
```diff
@@ -881,8 +881,10 @@ var imap_setup_message_view_page = function(uid, details, list_path, callback) {
                         .append(msgResponse.msg_text)
                         .append(msgResponse.msg_parts);
         document.title = $('.header_subject th').text();
-        $('.header_subject th').append('<i class="bi bi-x-lg close_inline_msg"></i>');
-        $('.close_inline_msg').on("click", function() { msg_inline_close(); });
+        if ($('.header_subject th').find('.bi.bi-star-half.account_icon').length === 0) {
+            $('.header_subject th').append('<i class="bi bi-x-lg close_inline_msg"></i>');
+            $('.close_inline_msg').on("click", function() { window.history.back(); });
```
Shadow243, line 886: there is a function `msg_inline_close` that can be used to close the inline, if you decide to not use you should then remove. But for me i would prefer to use the function instead of removing.

modules/inline_message/site.js
```diff
@@ -123,8 +128,11 @@ var capture_subject_click = function() {
         var uid = msg_details[0];
         var list_path = msg_details[1];
         var inline_msg_loaded_callback = function() {
-            $('.header_subject th').append('<i class="bi bi-x-lg close_inline_msg"></i>');
-            $('.close_inline_msg').on("click", function() { msg_inline_close(); });
+            if ($('.header_subject th').find('i.bi.bi-x-lg.close_inline_msg').length === 0) {
```
Shadow243, line 131: This is very specific `find('i.bi.bi-x-lg.close_inline_msg')` do we have many  tag i with class close_inline_msg ?

GedeonTS: Fixed!

PR #1299, opened 2024-10-24
Add a privacy settings section with the external images whitelist config

jacob-js:


modules/core/functions.php
```diff
@@ -623,3 +623,17 @@ function check_file_upload($request, $key) {
...
+
+function privacy_setting_callback($val, $key, $mod) {
+    $setting = Hm_Output_privacy_settings::$settings[$key];
+    $user_setting = $mod->user_config->get($key . "_setting");
+    $update = $mod->request->post['update'];
+
+    if ($update) {
+        // TODO: FIX - the updated setting doesn't get saved.
+        $val = implode($setting['separator'], array_filter(array_merge(explode($setting['separator'], $user_setting), [$val])));
+        $mod->user_config->set($key . '_setting', $val);
```
jacob-js, line 635: Hi @kroky, could you have a look at this and let me know if you find something that might be blocking the updated settings from being saved?

kroky: Did you check the Settings -> Save page? config.php defines crypt_state function that checks if cypth is running in a single server mode and auth is IMAP - only then the settings are actually saved when they are set. In all other cases, they are added to the session and stored after user provides their password on the save page which is used as an encryption key to store the settings on the disk or db. Does this help?
If you tried this and the privacy setting is still not saved, then maybe you are hitting an issue with the size of the value?

jacob-js: The value is not even updated in the session. It works fine from the settings page, but it seems impossible to update a setting outside of the settings page.

> Did you check the Settings -> Save page?

The save page seems to recognize the updated record. Maybe simply, because of the call to: `$mod->session->record_unsaved('Privacy settings updated');`

kroky: Yes, record_unsaved tells it to recognize the change even though there is no change. Please check if the callback has all the parameters passed - I see that process_site_setting calls the callback but might not pass the handler module that you need in order to get the user config. Maybe it doesn't pass the right value as well?

jacob-js: That could be a thing, let me check that out.

jacob-js: I figured I had to explicitly update the user data in the session, though it’s still unknown why `config->set()` doesn't work by itself.

kroky: Fair enough!

PR #1311, opened 2024-11-01
Removed forget/save actions

josaphatim:
This PR removes server forget/save buttons in favor of edit button.
The context is that many fields are being added on server page (lately sieve_tls) and others may come. The forget/save cannot handle these in the future. Edit button instead allows users to edit all together SMTP and IMAP + Sieve with all the fields.


kroky:
Yes, I followed the same approach with the upcoming EWS integration.

PR #1313, opened 2024-11-02
Tweak the sidebar for a modern look and feel

jacob-js:
Further enhancements could be made. If anyone would like to help, the following points need attention:

- [ ] Content such as the changes notice button, debug mode, and logout button are misplaced.
- [ ] The link to the home page should not be in the sidebar as an item, instead, it should be accessible via the app logo.
- [ ] The hide folders icon down in the sidebar, should not be present on mobile viewports. On desktop screens, it has to be replaced by something similar to the menu toggle button on mobile screens.


josaphatim:
@jacob-js thank for this. But can you fix the logo on dark themes (such as darkly theme) as it is not visible ?

jacob-js:
@josaphatim This PR doesn't make the sidebar perfect, please feel free to change whatever you think should be changed for the greater good, including the logo on dark themes!

Issue #1314, opened 2024-11-02
.env.example assigns DISABLE_ORIGIN_CHECK twice


Denperidge:
## 🐛 Bugreport
Not exactly a bug, but the current version the .env.example file has DISABLE_ORIGIN_CHECK assigned twice: on [line 32](https://github.com/cypht-org/cypht/blob/abdd8aab4284967bf676a3f8fdbc37c793120f51/.env.example#L32) & [line 78](https://github.com/cypht-org/cypht/blob/abdd8aab4284967bf676a3f8fdbc37c793120f51/.env.example#L78)

PR #1316, opened 2024-11-03
[FIX].env.example: DISABLE_ORIGIN_CHECK assigns twice

Baraka24:
[Related issue](https://github.com/cypht-org/cypht/issues/1314)


.env.example
```diff
@@ -29,8 +29,6 @@ USER_CONFIG_TYPE=file
 USER_SETTINGS_DIR=/var/lib/hm3/users
 ATTACHMENT_DIR=/var/lib/hm3/attachments
 
-DISABLE_ORIGIN_CHECK=false
```
Shadow243, line 32: Set the other one to false by default

Issue #1300, opened 2024-10-25
When sending an email, the SMTP ID field should not repeat identical information (it clutters the interface and adds no value)


marclaporte:
For Cypht profiles, I often use an email address as a display name.

The field shows data coming from 3 fields. If the data is different, it's fine to show all 3. But if there is some repetition, just show each unique value once.

![2024-10-25_103856](https://github.com/user-attachments/assets/ee1e9a83-4f01-4279-9392-9f5785ffe6bd)

PR #1326, opened 2024-11-06
[ENH] Remove duplicate SMTP display names in dropdown

DKing2222:
## Pullrequest
Enhancement to remove duplicate entries if there is some repetition in the Compose SMTP ID dropdown menu during email composition by only displaying unique values.

### Issues

https://github.com/cypht-org/cypht/issues/1300


modules/smtp/modules.php
```diff
@@ -1591,7 +1591,7 @@ function smtp_server_dropdown($data, $output_mod, $recip, $selected_id=false) {
                     $res .= 'selected="selected" ';
                 }
                 $res .= 'value="'.$output_mod->html_safe($vals['id']).'">';
-                $res .= $output_mod->html_safe(sprintf("%s - %s", $vals['user'], $vals['name']));
+                $res .= $output_mod->html_safe(($vals['user'] == $vals['name']) ? $vals['user'] : sprintf("%s - %s", $vals['user'], $vals['name']));
```
jacob-js, line 1594: @DKing2222 did you test? The statement you should have modified is a few lines above (at line 1584) which outputs all the server SMTP profiles. The alternative is used when a server does not have a profile!

DKing2222: I tested with and without a profile because even without a profile configured there was a repetition which as claimed does not bring any value

PR #1331, opened 2024-11-11
make sieve data format automatic on connect

Shadow243:
Related comment from @marclaporte : https://avan.tech/item119173#threadId81209

To support both Cypht 1.4 (which uses `tls://` prefixes for Sieve hostnames) and newer versions (which omit `tls://`), we’ve implemented an automated way to determine the correct Sieve hostname format using `Nux_Quick_Services`. Here's how it works:

1. **Automatic Host Detection**: The code checks `Nux_Quick_Services` to retrieve the correct Sieve host for each service. For example, for `migadu`, it will use `domain.com:port`.
   
2. **Version Compatibility**: If the running version is Cypht 1.4, it will prepend `tls://` to the hostname (e.g., `tls://domain.com:port`). For newer versions, it uses the plain hostname format (`domain.com:port`).

3. **Seamless Integration**: This ensures compatibility with both Cypht 1.4 and newer versions without manual configuration changes, reducing user setup errors and simplifying support.

### DIFF
--- a/modules/sievefilters/hm-sieve.php
+++ b/modules/sievefilters/hm-sieve.php
@@ -2,10 +2,16 @@
 <?php
 
 class Hm_Sieve_Client_Factory {
-    public function init($user_config = null, $imap_account = null)
+    public function init($user_config = null, $imap_account = null, $is_nux_supported = false)
     {
         if ($imap_account && ! empty($imap_account['sieve_config_host'])) {
-            list($sieve_host, $sieve_port) = parse_sieve_config_host($imap_account['sieve_config_host']);
+            // Check if module nux is enabled and if it is, get the sieve host from the services
+            if($is_nux_supported && $sieve_config = get_sieve_host_from_services($imap_account['server'])) {
+                $sieve_host = $sieve_config['host'];
+                $sieve_port = $sieve_config['port'];
+            } else {
+                list($sieve_host, $sieve_port) = parse_sieve_config_host($imap_account['sieve_config_host']);
+            }
             $client = new PhpSieveManager\ManageSieve\Client($sieve_host, $sieve_port);
             $client->connect($imap_account['user'], $imap_account['pass'], $imap_account['sieve_tls'], "", "PLAIN");
             return $client;
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1576,7 +1576,7 @@ if (!hm_exists('connect_to_imap_server')) {
 
                 include_once APP_PATH.'modules/sievefilters/hm-sieve.php';
                 $sieveClientFactory = new Hm_Sieve_Client_Factory();
-                $client = $sieveClientFactory->init(null, $server);
+                $client = $sieveClientFactory->init(null, $server, $context->module_is_supported('nux'));
 
                 if (!$client && $show_errors) {
                     Hm_Msgs::add("ERRFailed to authenticate to the Sieve host");
--- a/modules/nux/modules.php
+++ b/modules/nux/modules.php
@@ -709,4 +709,8 @@ class Nux_Quick_Services {
         }
         return array();
     }
+
+    static public function get() {
+        return self::$services;
+    }
 }
--- a/modules/sievefilters/functions.php
+++ b/modules/sievefilters/functions.php
@@ -106,7 +106,7 @@ if (!hm_exists('get_mailbox_filters')) {
     {
         $factory = get_sieve_client_factory($site_config);
         try {
-            $client = $factory->init($user_config, $mailbox);
+            $client = $factory->init($user_config, $mailbox, in_array(mb_strtolower('nux'), $site_config->get_modules(true), true));
             $scripts = [];
             foreach ($client->listScripts() as $script) {
                 if (mb_strstr($script, 'cypht')) {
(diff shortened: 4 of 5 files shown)


modules/sievefilters/hm-sieve.php
```diff
@@ -2,10 +2,16 @@
 <?php
 
 class Hm_Sieve_Client_Factory {
-    public function init($user_config = null, $imap_account = null)
+    public function init($user_config = null, $imap_account = null, $is_nux_supported = false)
     {
         if ($imap_account && ! empty($imap_account['sieve_config_host'])) {
-            list($sieve_host, $sieve_port) = parse_sieve_config_host($imap_account['sieve_config_host']);
+            // Check if module nux is enabled and if it is, get the sieve host from the services
+            if($is_nux_supported && $sieve_config = get_sieve_host_from_services($imap_account['server'])) {
```
kroky, line 9: Quite easier approach would be to store sieve config as part of imap server list and just use it here. Even if it is a quick service setup by nux module, make sure you always store the sieve host in the imap config. Then, you won't need to distinguish if it is a nux quick service one or normal imap.

Shadow243: Well noted. However, in this context where @marclaporte  wanted that for instances with already configured servers, we handle the data structure change introduced in [this PR](https://github.com/cypht-org/cypht/pull/1199), it was necessary to find a way to connect Sieve using the structure according to the current version. For example, if in the current instance the format is `tls://`, then Sieve must automatically adapt to that.

I therefore relied on this logic to retrieve the data format from the NUX module, as it was also changed in [this PR](https://github.com/cypht-org/cypht/pull/1199).

The idea was that NUX would always provide the correct format regardless of the version. However, if I store these settings in an IMAP list, how can we avoid having to edit and reconfigure the servers on older instances? The goal is to format them automatically without user intervention.

kroky: I believe we need a migration technique for this problem and other similar problems we are facing. Probably worth working on that in another PR as this one is urgent but the general idea is this:
- whenever a configuration format changes, we add a migration that will run on the old deployments
- we need to make sure this migration is run at each deployment - probably add it as part of the scripts/config_gen.php
- we can use configuration updates in a migration but also database schema updates (like the sqlite one)

Shadow243: Copy that. I add a task for this

Shadow243: > Quite easier approach would be to store sieve config as part of imap server list and just use it here. Even if it is a quick service setup by nux module, make sure you always store the sieve host in the imap config. Then, you won't need to distinguish if it is a nux quick service one or normal imap.

@kroky trying to implement here: https://github.com/cypht-org/cypht/pull/1490/commits/15f4447f01410a17be51a5509598edfca32d48d8

I noticed that there's one scenario where we still need to use the Nux module, even though we're now storing Sieve configs as part of IMAP. We now have the ability to enable or disable Sieve from the settings, even when the module is enabled via the .env file.

The specific case is: when connecting a server while Sieve is disabled in the settings, we're not storing the Sieve config simply because the form doesn't show the Sieve inputs anymore. In this case, we still need the Nux module.

kroky: I don't like that approach as it breaks module separation of concerns and including functions from other modules is undesirable at the latest. Modules should be self-contained as much as possible and if they depend on other modules, then this should be handled during configuration or cypht initialization.
With Sieve, we apparently have this dependency - sieve module depends on imap module. If imap is not enabled, sieve should also be not enabled. We can have imap enabled and sieve disabled in config .env - then, this means we don't show any sieve inputs and we don't try to guess any sieve settings automatically. If users need sieve, it should be enabled via the env config.
Then, storing it as part of the imap config must be done in sieve module. Imap module shouldn't be aware of sieve in the perfect world. Sieve module can interfere with imap functions and storage as it depends on that module, thus we are sure imap is enabled (and loaded). I want to avoid requiring other modules function files and stuff like that. Modules should be self-contained.
Thus, I think the way to do is:
- make sieve module store sieve config in imap
- skip recehcking and reinitialziation from nux module on each imap check - that would be a waste of resources and also error-prone
- if sieve is enabled later after server is initialized, user must go to the imap form and re-save it - actually review the sieve config and save it in order to activate. That's a normal configuration step. If you want to make it easier for the user, you can display a note somewhere on the home page that sieve module is enabled but sieve configuration is not complete. That would be a much better approach and keep things where they should live than doing it automatically with the intra-module communication.

Shadow243: > * if sieve is enabled later after server is initialized, user must go to the imap form and re-save it - actually review the sieve config and save it in order to activate. That's a normal configuration step. If you want to make it easier for the user, you can display a note somewhere on the home page that sieve module is enabled but sieve configuration is not complete. That would be a much better approach and keep things where they should live than doing it automatically with the intra-module communication.

added at https://github.com/cypht-org/cypht/pull/1490/commits/18aea3e82694b9a81fc4e02008f0885d1ff77329

Shadow243: > * make sieve module store sieve config in imap

According tp `Hm_IMAP_List::add($imap_list)` or `Hm_IMAP_List::edit($server_id, $imap_list)` in `connect_to_imap_server` function, we are already storing sieve configs in imap.

kroky: I meant something else - we need to make modules more self-contained and reduce dependencies. Sieve config is stored in imap module but it is done in other modules than sievefilters one. We need to do the actual storing and retrieval in sievefilters module, not other modules as this breaks encapsulation. I've added more info to Cypht developer documentation: https://github.com/cypht-org/cypht-website/commit/9ed2ce0c5c5bdaeb140624f4142d04fbea7c12d8

PR #1339, opened 2024-11-12
Fixed tag display content

Shadow243:
Tags show message is no more working

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -733,62 +733,66 @@ function sort_by_internal_date($a, $b) {
  * @return array
  */
 if (!hm_exists('merge_imap_search_results')) {
-function merge_imap_search_results($ids, $search_type, $session, $hm_cache, $folders = array('INBOX'), $limit=0, $terms=array(), $sent=false) {
+function merge_imap_search_results($ids, $search_type, $session, $hm_cache, $folders=array('INBOX'), $limit=0, $terms=array(), $sent=false, $combined=false) {
     $msg_list = array();
     $connection_failed = false;
-    $sent_results = array();
     $status = array();
-    foreach($ids as $index => $id) {
+    $sent_results = array();
+    foreach ($ids as $id) {
         $cache = Hm_IMAP_List::get_cache($hm_cache, $id);
         $imap = Hm_IMAP_List::connect($id, $cache);
         if (imap_authed($imap)) {
             $server_details = Hm_IMAP_List::dump($id);
-            $folder = $folders[$index];
+            if ($combined) {
+                $folders = array_unique(array_merge($folders, $imap->get_special_use_mailboxes()));
+            }
             if ($sent) {
                 $sent_folder = $imap->get_special_use_mailboxes('sent');
                 if (array_key_exists('sent', $sent_folder)) {
                     list($sent_status, $sent_results) = merge_imap_search_results($ids, $search_type, $session, $hm_cache, array($sent_folder['sent']), $limit, $terms, false);
                     $status = array_merge($status, $sent_status);
                 }
+            }
+            foreach ($folders as $folder) {
                 if ($folder == 'SPECIAL_USE_CHECK') {
                     continue;
                 }
-            }
-            if ($imap->select_mailbox($folder)) {
-                $status['imap_'.$id.'_'.bin2hex($folder)] = $imap->folder_state;
-                if (!empty($terms)) {
-                    foreach ($terms as $term) {
-                        if (preg_match('/(?:[^\x00-\x7F])/', $term[1]) === 1) {
-                            $imap->search_charset = 'UTF-8';
-                            break;
+                if ($imap->select_mailbox($folder)) {
+                    $status['imap_'.$id.'_'.bin2hex($folder)] = $imap->folder_state;
+                    if (!empty($terms)) {
+                        foreach ($terms as $term) {
+                            if (preg_match('/(?:[^\x00-\x7F])/', $term[1]) === 1) {
+                                $imap->search_charset = 'UTF-8';
+                                break;
+                            }
                         }
-                    }
-                    if ($sent) {
-                        $msgs = $imap->search($search_type, false, $terms, array(), true, false, true);
+                        if ($sent) {
+                            $msgs = $imap->search($search_type, false, $terms, array(), true, false, true);
+                        }
+                        else {
+                            $msgs = $imap->search($search_type, false, $terms);
+                        }
                     }
                     else {
-                        $msgs = $imap->search($search_type, false, $terms);
+                        $msgs = $imap->search($search_type);
                     }
-                }
-                else {
-                    $msgs = $imap->search($search_type);
-                }
-                if ($msgs) {
-                    if ($limit) {
-                        rsort($msgs);
-                        $msgs = array_slice($msgs, 0, $limit);
-                    }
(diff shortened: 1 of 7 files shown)


modules/imap/functions.php
```diff
@@ -743,7 +743,8 @@ function merge_imap_search_results($ids, $search_type, $session, $hm_cache, $fol
         $imap = Hm_IMAP_List::connect($id, $cache);
         if (imap_authed($imap)) {
             $server_details = Hm_IMAP_List::dump($id);
-            $folder = $folders[$index];
+            $folder = $folders[$index] ?? $folders[0];
```
Shadow243, line 746: We are no longer sending 2 request, i meant getting folder using loop $ids is not correct, the param should no longer be an array, any idea @kroky ?

modules/tags/modules.php
```diff
@@ -103,13 +103,13 @@ class Hm_Handler_imap_tag_content extends Hm_Handler_Module {
...
+        $tag_id = $this->request->post['tag_id'];
         if ($ids && $tag_id) {
             $limit = $this->user_config->get('tag_per_source_setting', DEFAULT_TAGS_PER_SOURCE);
             $date = process_since_argument($this->user_config->get('tag_since_setting', DEFAULT_TAGS_SINCE));
             $folder = bin2hex('INBOX');
             if (array_key_exists('folder', $this->request->post)) {
-                $folder = $this->request->post['folder'];
+                $folder = bin2hex($this->request->post['folder']);
             }
             list($status, $msg_list) = merge_imap_search_results($ids, 'ALL', $this->session, $this->cache, array(hex2bin($folder)), $limit, array(array('SINCE', $date), array('HEADER X-Cypht-Tags', $tag_id)));
```
jacob-js, line 114: @Shadow243 Please enlighten me, why is the INBOX folder needed here?

Shadow243: We need to send a folder, and by default it is INBOX, however in this change here https://github.com/cypht-org/cypht/blob/master/modules/core/js_modules/Hm_MessagesStore.js#L167 you send tag_id as folder but it is an error, folders are for example: INBOX, Sent, Draft, etc...

I tried to remove the folder parameter but what if we need it to filter in another folder? So I left INBOX hoping that in the future we will go further to other folders

Note: t`ag_id` is different from a `folder`

modules/imap/functions.php
Shadow243: but I had an idea, if we send several folders at the same time especially folders it is already a table. like we want to display everything (send, received, etc...) which has as tag "Demo" for example

modules/tags/modules.php
jacob-js: As no explicit folder is needed here, we should keep the content of Hm_MessagesStore unchanged, that way we get an API that works for almost all lists of messages without varying properties, thus, assigning the `tag_id` to the folder property is the better way to go.

Also, does this fix the retrieval of the message rows?

Shadow243: We can't assign tag_id into folder. Like I said tag_id is different of folder.

feel free to push something on this PR if you have an idea.

jacob-js: Maybe you don't want to understand my point and prefer to stick to your opinion, but in the context where `tag_id` is being used, it wouldn't be unacceptable to assign its value to a property named `folder`.

Shadow243: This is exactly what I will understand, if I send tag_id in folder, this line will no longer have any meaning:
```
if (array_key_exists('folder', $this->request->post)) {
                $folder = $this->request->post['folder'];
```
remember in `$this->request->post['folder']` we have the tag_id, what's your thoughts ?

jacob-js: We can remove that statement because we don't use it.

Shadow243: Well noted, But if we remove the folder parameter, we will no longer have a filter, we will only search in INBOX and if I have an email from the Sent folder then it will never be displayed, but by passing folder separately from the tag_id we will add the possibility of filtering, ideally this PR should rather be modified and add to the user a possibility of sending the folder filter (INBOX, Draft, Sent, etc...)

Shadow243: Here's how the query worked before:
Screenshot 2024-11-16 at 14 23 38

jacob-js: A tag is a filter by itself; implementing additional folder filters on top of it creates unnecessary clutter in this case. It's like saying, 'I want the list of messages from `Everything`, but then show me only messages from a specific folder.' I'm fairly certain the client doesn't use such filtering.

Shadow243: Here is the answer hidden in what you just wrote, we can send 'Everything' this filter already exists. in merge_imap_search_results when it is the tag we must return all "INBOX,Sent, Drafts, etc...

jacob-js: Let me rephrase to help you understand what I said:

Imagine someone requests a list of messages (e.g., Archive), and on top of that filter, they add, _Please show me only messages from the folder `Sent`._ Would that add any value?

Shadow243: I understood your point of view and it even reminds me that I had not taken into account the fact that we must display everything (from the whole folders once it is tagged), I have just put the PR in draft because I am in the process of broadening the reflection, I suggest that you let me finish modifying then I will notify you

Shadow243: @jacob-js Here is the result I was looking for, I display the all email that have been tagged from folders all folders using `get_special_use_mailboxes()`. In this screenshort the first email is in "Sent" and the rest in 'INBOX'.
Screenshot 2024-11-16 at 16 56 27

modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -164,7 +165,7 @@ class Hm_MessagesStore {
                     break;
                 case 'tag':
                     hook = "ajax_imap_tag_data";
-                    folder = getParam('tag_id');
+                    tag_id = getParam('tag_id');
```
jacob-js, line 168: Please revert here.

Shadow243: That is the tag_id why revert ?

jacob-js: So, from our entire conversation last Saturday, it seems you didn’t grasp the key points. I recommend revisiting and carefully reviewing all the chats.

modules/imap/functions.php
```diff
@@ -733,62 +733,66 @@ function sort_by_internal_date($a, $b) {
  * @return array
  */
 if (!hm_exists('merge_imap_search_results')) {
-function merge_imap_search_results($ids, $search_type, $session, $hm_cache, $folders = array('INBOX'), $limit=0, $terms=array(), $sent=false) {
+function merge_imap_search_results($ids, $search_type, $session, $hm_cache, $folders = array('INBOX'), $limit = 0, $terms = array(), $sent = false, $combined = false) {
```
jacob-js, line 736: This function is already too complex, with eight parameters making its purpose unclear. This complexity will likely cause issues in the future, if not already, as larger functions are more prone to breaking in multiple parts of the application. It would be better to refactor it to ensure it serves a clear and focused purpose.

Shadow243: @jacob-js This PR doesn't make the tag perfect, please feel free to change whatever you think should be changed for the greater good, including the `merge_imap_search_results`!

PR #1344, opened 2024-11-12
Sieve socket creation failed

christer77:
This PR resolve **_Sieve: Socket creation failed_** error.

The first constant is that, since the appearance of this error, the **_BLOCK SENDER_** button is always disabled.

Related screen:

BLOCK SENDE DISABLED-SIEVE-ERROR

To reproduce this error, just open message inline mode or in new tab

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1525,7 +1525,10 @@ function parse_sieve_config_host($host) {
         $host = $url['path'];
     }
     $port = $url['port'] ?? '4190';
-    return [$host, $port];
+    $scheme = $url['scheme'] ?? 'tcp://';
+    $tls = $scheme === 'tls';
+    // echo "<pre>"; var_dump($url, $host); die;
+    return [$host, $port, $tls];
 }}
 
 if (!hm_exists('connect_to_imap_server')) {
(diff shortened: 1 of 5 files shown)


marclaporte:
Related: https://github.com/cypht-org/cypht/pull/1331

modules/imap/functions.php
```diff
@@ -1523,9 +1523,13 @@ function parse_sieve_config_host($host) {
     $url = parse_url($host);
     if(!isset($url['host'])) {
         $host = $url['path'];
+    } else {
+        $host = $url['host'];
     }
     $port = $url['port'] ?? '4190';
-    return [$host, $port];
+    $scheme = $url['scheme'] ?? 'tcp://';
```
Shadow243, line 1530: I'm not sure this will work with stalwart too, try and ping me

christer77: I'll get back to you after the test

Shadow243:
Related PR: https://github.com/cypht-org/cypht/pull/1199

modules/imap/functions.php
Shadow243: https://github.com/cypht-org/cypht/issues/1153

Issue #776, opened 2023-09-29
Review the whole process/code/documentation for Cypht to work for Google and Microsoft's email offerings. (was: Error Ocurred when attempting gmail oauth2 setup)


Terramoto:
I'm running the docker version, i've:

- setup the oauth2.ini file in app_data path specified in hm3.ini
- php script/config_gen.php

The file is detected during the config_gen.php, i then proceed to the interface:

- pick up the gmail option from the drop-down list. 
- Setup the email address and name,
- Proceed to the Google authentication, permissions page, 
- Check emails and calendars access permissions and then it redirects me to the webmail url but signed out. I type in login details but a "Error Ocurred" is displayed and the url reverts to the default webmail url path.

I've notice the redirect URL sends two scopes:
?state=nux_authorization&code=(Authorization-Code)&scope=https://mail.google.com/%20https://www.googleapis.com/auth/contacts

I've tried removing the second scope and this time i got the "Error Ocurred" message but it didn't signed me off....

marclaporte:
Thank you @Terramoto for this report.

Can you try with debug mode activated as per https://cypht.org/install.html ?

Thanks!

Terramoto:
> Thank you @Terramoto for this report.
> 
> Can you try with debug mode activated as per https://cypht.org/install.html ?
> 
> Thanks!

I'm afraid it doesn't seem to work, i'm using the docker version provided in the README:
https://hub.docker.com/r/sailfrog/cypht-docker

```
# ln -s /usr/local/share/cypht /var/www/html/mail-debug
ln: /var/www/html/mail-debug: No such file or directory
```

I've tried to set the display_errors on in the PHP.ini file but that hasn't made much of a difference.

marclaporte:
@rodriguezny How can we activate errors in Cypht Docker?

rodriguezny:
Did you try to visualize the container log ?

rodriguezny:
Thanks @Terramoto for the report.
The notices are due to php version, no  need to be worried about that, but the log doesn't show the error that causes the "error occurred" message when attempting gmail auth2 message.
Adding a Gmail account seems broken, I don't have precision, give us some time to investigate on the issue on cypht without docker and cypht docker version, will come back to you later.

Did you try to setup gmail auth2 on cypht outside docker (no docker version) ? If no, you can try it and provide us feedback waiting for us to come back to you with the fix/solution ?

marclaporte:
[...]

PR #1278, opened 2024-10-09
feat(frontend/backend): add support for Exchange Web Services (EWS)

kroky:
Feature request: https://github.com/cypht-org/cypht/issues/247

Work in progress:
1. Add Mailbox bridge for imap and similar modules to work with protocols other than IMAP - e.g. EWS (WIP).
2. Implement EWS methods using php-ews package and raw SOAP calls.
3. Review and write tests.

PR #1348, opened 2024-11-14
fix(cypht): error ocurred when attempting microsoft's email oauth2 setup

christer77:
Related [issue](https://github.com/cypht-org/cypht/issues/776)

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -853,6 +853,9 @@ function imap_refresh_oauth2_token($server, $config) {
         elseif ($server['server'] == 'imap-mail.outlook.com') {
             $details = $oauth2_data['outlook'];
         }
+        elseif ($server['server'] == 'imap-mail.office365.com') {
+            $details = $oauth2_data['office365'];
+        }
         if (!empty($details)) {
             $oauth2 = new Hm_Oauth2($details['client_id'], $details['client_secret'], $details['client_uri']);
             $result = $oauth2->refresh_token($details['refresh_uri'], $server['refresh_token']);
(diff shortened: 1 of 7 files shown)


modules/imap/functions.php
```diff
@@ -1093,7 +1093,7 @@ function clear_existing_reply_details($session) {
  */
 if (!hm_exists('imap_authed')) {
 function imap_authed($imap) {
-    return is_object($imap) && ($imap->get_state() == 'authenticated' || $imap->get_state() == 'selected');
+    return is_object($imap) && ($imap->get_state() == 'authenticated' || $imap->get_state() == 'selected' || $imap->get_state() == 'connected');
```
kroky, line 1096: If oauth2 returns connected instead of authenticated, let's handle that in the imap get_state() method. We have this refactored in the upcoming https://github.com/cypht-org/cypht/pull/1278, so we need a reliable way to check if imap server connection is ready to use or not. Instead of scattering this check throughout the modules, let's check it only in one place.

modules/nux/site.js
```diff
@@ -53,11 +53,11 @@ var display_final_nux_step = function(res) {
                 {'name': 'imap_server_id', value: res.nux_server_id},
                 {'name': 'imap_service_name', value: res.nux_service_name}],
                 function () {
-                    Hm_Utils.redirect();
+                    Hm_Utils.redirect('?page=servers');
```
kroky, line 56: Why do you redirect to a specific page? Isn't the nux module also having a separate url that uses the server forms? redirect without any argument should actually redirect to the same page.

PR #1356, opened 2024-11-16
[ENH]DragAndDrop: add drag and drop functionality on touchscreens

GedeonTS:
[Task](https://avan.tech/item120201)


jacob-js:
Kudos @GedeonTS, it looks good! I made a small fix in my last commit to address the issue of drag-and-drop working intermittently.

Issue #1322, opened 2024-11-05
Docker image broken - GLOB_BRACE undefined


indridieinarsson:
## 🐛 Bugreport
It seems the docker image is broken. 
PR #1288 introduces usage of the global constant `GLOB_BRACE`.

Building a docker image is successful, however, when starting it, it fails with :
`cypht-1  | PHP Fatal error:  Uncaught Error: Undefined constant "GLOB_BRACE" in /usr/local/share/cypht/scripts/config_gen.php:193
`
This is a known issue with the php alpine docker image, which does not support the functionality configured with this constant : https://github.com/docker-library/php/issues/719

PR #1360, opened 2024-11-17
fix(docker): replace php:8.1-fpm-alpine image by php:8.1-fpm-bookworm: alpine doesn't support GLOB_BRACE

IrAlfred:
## Replace php:8.1-fpm-alpine docker base image by php:8.1-fpm-bookworm

This merge request replaces the php:8.1-fpm-alpine image with php:8.1-fpm-bookworm in our Docker setup. The change addresses the challenges caused by the minimalistic nature of the Alpine-based image, which often lacks key features and extensions required for smooth development and runtime environments.

* [Related issue](https://github.com/cypht-org/cypht/issues/1322)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,38 +1,42 @@
-FROM php:8.1-fpm-alpine
+FROM php:8.1-fpm-bookworm
 
-WORKDIR "/usr/local/share/cypht"
+WORKDIR /usr/local/share/cypht
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV COMPOSER_CACHE_DIR=/tmp/composer_cache
+ENV COMPOSER_HOME=/tmp/composer_home
 
 RUN set -e \
-    && apk add --no-cache \
-    supervisor nginx composer sqlite freetype libpng libjpeg-turbo \
-    php-session php-fileinfo php-dom php-xml libxml2-dev php-xmlwriter php-tokenizer \
-    && apk add --no-cache --virtual .build-deps \
+    && apt-get update && apt-get install -y \
+    supervisor nginx sqlite3 libfreetype6-dev libpng-dev libjpeg-dev \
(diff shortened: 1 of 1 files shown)


jacob-js:
@IrAlfred, why shouldn't we consider changing the PHP image we're using? The issue arises due to the minimalistic nature of the Alpine image, which omits certain features. While it does reduce the Docker image size, is it worth the development effort required to work around the missing extensions?

IrAlfred:
> @IrAlfred, why shouldn't we consider changing the PHP image we're using? The issue arises due to the minimalistic nature of the Alpine image, which omits certain features. While it does reduce the Docker image size, is it worth the development effort required to work around the missing extensions?

I've thought about it for a while... I agree with switching the base image. Which image would you recommend we use ?

indridieinarsson:
> > @IrAlfred, why shouldn't we consider changing the PHP image we're using? The issue arises due to the minimalistic nature of the Alpine image, which omits certain features. While it does reduce the Docker image size, is it worth the development effort required to work around the missing extensions?
> 
> I've thought about it for a while... I agree with switching the base image. Which image would you recommend we use ?

wouldn't php:8.1-fpm-bookworm be the one to use? Same php version, but running on the most up-to-date debian. Same publisher as the alpine-based image in use now.

jacob-js:
Thank you @indridieinarsson for the suggestion! @IrAlfred please try this one and keep us informed of the outcome.

IrAlfred:
Thank you @jacob-js , I work on it

IrAlfred:
Please review @jacob-js and @indridieinarsson

mercihabam:
@indridieinarsson, please share some feedback as soon as you can!

indridieinarsson:
@mercihabam  @IrAlfred. Not really an insightful comment, but I did test the PR, and it works. Great job.

Issue #1350, opened 2024-11-15
Cypht marking all emails as read


PaulTGG:
## 🐛 Bugreport
The latest Cypht is marking all messages per account as being read, if you open an email (not inline), and perform an action on it (such as delete).

### Version & Environment

Rev: 02182bb

OS:  []

### Steps to reproduce
1. Install Cypht
2. Open an email with an account that has some unread messages in it.
3. Delete the email.  The other messages in that accounts inbox will be marked as read.

PaulTGG:
More info: I just noticed that reading one email marks all the other emails in that account as being read.  No specific action required in the test I just did.

christer77:
Thanks for this.

I'm getting started

christer77:
Hello @PaulTGG 

I must say that I tested both cases, but I can't reproduce issue. In short, unread messages remain as such.

PaulTGG:
Hi @christer77, I just tried again, with a fresh Debian install, and a fresh install of the latest Cypht master.  As soon as I opened one email (in this case, it was not inline), and went back to my combined inbox, all the other emails for that account were marked as read.  (I confirmed this with a separate email app.)

marclaporte:
@PaulTGG @christer77 please coordinate a screenshare on https://gitter.im/cypht-org/community

christer77:
Hi @PaulTGG,

Can you confirm [this](https://github.com/cypht-org/cypht/pull/1376)

PaulTGG:
@christer77 Looks good to me, thanks!

PR #1376, opened 2024-11-21
[FIX]Cypht marking all emails as read

christer77:
Related [issue](https://github.com/cypht-org/cypht/issues/1350)

Related [RFC](https://www.rfc-editor.org/rfc/rfc3501)


christer77:
> @christer77 please add the rfc link in the PR description

okay

Issue #1355, opened 2024-11-16
Display raw not working


ulfgebhardt:
## 🐛 Bugreport

Display raw not working.

When I click on `Raw` in the top menu of an email to see headers and alike, I am redirected to the startpage.
It is not possible to display an raw email.

### Version & Environment

Rev: [v2.4.0]

OS:  [chromium, alpine]

### Mitigation

There is a `Show Source` which opens a new window with what I need. not as convenient, but it serves its purpose.

### Steps to reproduce
1. Open Mail
2. Click `Raw`
3. Get redirected - no raw email
4. ...
5. Profit

ulfgebhardt:
> @ulfgebhardt, thank you for your report! The `raw` action should be removed, as the `show source` action provides all the necessary data and offers additional functionalities, such as Download and Copy to Clipboard.

I like the `raw` method way more then the show source - thats your choice in the end, but the raw thing is very useful

jacob-js:
> but the raw thing is very useful

Please explain!

ulfgebhardt:
> > but the raw thing is very useful
> 
> Please explain!

Assume you have an email like this and you are unsure if its legit or whatever is wrong with it and you want to investigate
![image](https://github.com/user-attachments/assets/2b63e421-b821-46b9-ba87-df95a9040692)

You can just switch to `raw` mode and see whats happening - inline & fast - no need to switch tabs, no further hustle (also notice the nice clean interface with very little colors) - once your done you can switch back to `normal` and keep reading.
![image](https://github.com/user-attachments/assets/ba0724c9-4299-429e-b098-44763a058e16)

In terms of usability the proposed solution is a regression.
> Open Email -> Raw -> Normal
> Open Email -> Show Source -> Switch Tab -> Close Tab

The same workflow applies to images - see #1367 - with inline images you don't need to download them and fiddle them out of the downloads folder, but just can look at them - no hustle

mercihabam:
The reasoning here is that the page used for opening a message functions as a reading board, not a debugging window. Most users don’t even know what "raw" refers to in the context of emails, so it doesn’t make sense to keep it within the message template frame. Additionally, opening the source in a new tab, where it occupies the entire screen, makes it easier for users to focus on the content.

PR #1377, opened 2024-11-21
[REF] output_modules.php Remove the unnecessary Raw action displayed …

amaninyumu1:
[Related Issue](https://github.com/cypht-org/cypht/issues/1355)


modules/imap/output_modules.php
```diff
@@ -376,9 +376,6 @@ protected function output() {
             if ($msg_part === '0') {
                 $txt .= ' | <a class="normal_link hlink msg_part_link normal_link" data-message-part="" href="#">'.$this->trans('normal').'</a>';
```
jacob-js, line 377: We don't have any use for this anymore. Please remove it as well!

amaninyumu1: it's done

PR #1380, opened 2024-11-23
refactor(backend): db session locking to support exclusive mode

Shadow243:
Related task [item89272](https://avan.tech/item89272)


lib/session_db.php
```diff
@@ -169,4 +185,94 @@ public function db_start($request) {
...
+     */
+    private function acquire_lock($key) {
+        $lock_name = 'session_lock_' . substr(hash('sha256', $key), 0, 51);        
+        $query = '';
+        $params = [];
+
+        switch ($this->db_driver) {
+            case 'mysql':
+                $query = 'SELECT GET_LOCK(:lock_name, :timeout)';
+                $params = [':lock_name' => $lock_name, ':timeout' => $this->lock_timeout];
```
kroky, line 202: We should implement the lock_timeout in other drivers as well - this is important for concurrent requests. You can use a polling mechanism with intervals of 500ms to 1 second.

scripts/setup_database.php
```diff
@@ -86,8 +86,11 @@
 if (strcasecmp($session_type, 'DB')==0) {
     printf("Creating database table hm_user_session ...\n");
 
-    if ($db_driver == 'mysql' || $db_driver == 'sqlite') {
+    if ($db_driver == 'mysql') {
         $stmt = "{$create_table} hm_user_session (hm_id varchar(255), data longblob, date timestamp, primary key (hm_id));";
+    } elseif($db_driver == 'sqlite') {
+        //0 means unlocked, 1 means locked 
+        $stmt = "{$create_table} hm_user_session (hm_id varchar(255), data longblob, lock INTEGER DEFAULT 0, date timestamp, primary key (hm_id));";
```
kroky, line 93: How would existing users update their databases? I think we should introduce some kind of a migration mechanism or handle this in the code.

lib/session_db.php
Shadow243: @kroky I just updated, I just did a test with mysql. I'm waiting to see if there are any corrections from you so far before confirming with `pgsql` too but it should work for `pgsql` too

Shadow243:
@kroky I tested with postgresql and mysql

I got a new error with postgres:
```
CREATE TABLE
psql:/home/runner/work/cypht/cypht/tests/phpunit/data/schema_postgres.sql:10: ERROR:  type "longblob" does not exist
LINE 1: ... EXISTS hm_user_session (hm_id varchar(255), data longblob,
```

lib/session_db.php
kroky: The polling looks good.

scripts/setup_database.php
kroky: The way you did it is a bit fragile and would require careful maintenance of the existing db schema as a php array which might be a problem for programmers. I'd suggest we use a migration-kind of mechanism where we have initial schema and then migration files for each change we do - much like Laravel, Rails and other similar frameworks work. The migrations can be simple sql files in a special directory where the main schema resides. Then a simple script can be added to migrate - checking which ones are already executed and executing the rest. We should have a table to track the executed migrations, so some kind of ID or migration file name should be recorded in that table.

kroky:
It seems postgres is trying to use the schema file for mysql and not the one for postgres - check the calling code and why does it try that.

scripts/setup_database.php
Shadow243: oh i see. I'l work on that.

Shadow243:
> It seems postgres is trying to use the schema file for mysql and not the one for postgres - check the calling code and why does it try that.

Previously, we had a single file. I added a separate one specific to PostgreSQL, using the bytea type for the data field instead of longblob.

Shadow243:
@kroky The current function effectively generates a migration file with a timestamp-based name, ensuring the correct order of migrations. While this implementation works as expected, it could be improved by adding more flexibility. Once the PR: https://github.com/cypht-org/cypht/pull/1196 is approved, we can further enhance this functionality by integrating Symfony Console, making migration management more flexible and accessible directly from the command line.

kroky:
> @kroky The current function effectively generates a migration file with a timestamp-based name, ensuring the correct order of migrations. While this implementation works as expected, it could be improved by adding more flexibility. Once the PR: #1196 is approved, we can further enhance this functionality by integrating Symfony Console, making migration management more flexible and accessible directly from the command line.
> 
> Once the PR is validated, I will also migrate the tests to use migrations for table creation and remove all .sql files

This was is better but I'd like to change the approach. I see you copied large chunks of class contents from other frameworks. This is not the way to go because it makes maintenance harder and probably violates licenses. If you really wish to be fancy with php migrations, you can take a look here: https://packagist.org/?query=migrate for a suitable library and use one via composer. No need to maintain all that code.
Otherwise, my idea was much more simpler - do sql migrations - files stored in database/*.sql - run them one by one with timestamps and just execute whatever is there. No rollback. This should allow for the migration runner to be as simple as 1-2 methods with several lines of code - get all the migrations, check what has not been run against the current db and execute it via PDO.

Shadow243:
> This was is better but I'd like to change the approach. I see you copied large chunks of class contents from other frameworks. This is not the way to go because it makes maintenance harder and probably violates licenses. If you really wish to be fancy with php migrations, you can take a look here: https://packagist.org/?query=migrate for a suitable library and use one via composer. No need to maintain all that code.
> Otherwise, my idea was much more simpler - do sql migrations - files stored in database/*.sql - run them one by one with timestamps and just execute whatever is there. No rollback. This should allow for the migration runner to be as simple as 1-2 methods with several lines of code - get all the migrations, check what has not been run against the current db and execute it via PDO.

You raised valid points, and I agree that maintaining a custom migration system could become complex over time.

If we consider using .sql files for database migrations, we must address compatibility across different database engines. SQL syntax can vary between systems like MySQL, PostgreSQL, and SQLite. To manage this, we could organize migrations into subdirectories by engine, such as:
```
/database/migrations/mysql
/database/migrations/postgresql
/database/migrations/sqlite
```
The migration runner would detect the current database engine and execute only the files from the relevant directory. This keeps the migration process straightforward while supporting multiple database systems.

Alternatively, leveraging a well-maintained package from Packagist could be a more robust and scalable solution. Libraries such as [bizley/migration](https://packagist.org/packages/bizley/migration) offer extensive support for multiple database systems and advanced features like schema management. Another option is [doctrine/migrations](https://packagist.org/packages/doctrine/migrations), which provides powerful migration capabilities but may introduce additional complexity.

Each approach has its strengths. Using .sql files keeps things simple and transparent but requires custom implementation and careful handling of database-specific syntax(ex: 20241209040300_add_lock_to_hm_user_session_table.sql). Adopting a mature package simplifies maintenance and enhances functionality but adds external dependencies and a learning curve.

@kroky What can you advise me ?

kroky:
Yes, both options are valid. I think it is more in the spirit of cypt to go with the simpler approach and use per-database migration file. Cypht database is relatively simple, schema changes are not done often, so it doesn't make sense to me to include big dependencies as doctrine now.

Shadow243:
> > Yes, both options are valid. I think it is more in the spirit of cypt to go with the simpler approach and use per-database migration file. Cypht database is relatively simple, schema changes are not done often, so it doesn't make sense to me to include big dependencies as doctrine now.
> 
> Understood, I will create the SQL files for each DBMS and test then after ping you.

Shadow243:
@kroky I  updated schema for all databases(sqlite,pgsql and mysql). you can review.

database/migrations/pgsql/20241209010100_create_hm_user_table.sql
```diff
@@ -0,0 +1,12 @@
+DROP TABLE IF EXISTS hm_user;
+
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM information_schema.tables
```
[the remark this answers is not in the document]
Shadow243: I'm implementing it now and will push the update tonight.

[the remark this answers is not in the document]
Shadow243: @kroky I have updated and tested the changes with MySQL, and everything is working as expected. Please review the implementation, and once confirmed, I will proceed with testing on SQLite and PostgreSQL to ensure compatibility across all databases.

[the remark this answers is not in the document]
kroky: MySQL one looks great, thanks!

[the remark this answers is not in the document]
Shadow243: @kroky everything is working as expected with other sqlite and Postgres too.

database/migrations/sqlite/20241209010200_add_lock_columns.sql
```diff
@@ -0,0 +1,21 @@
+PRAGMA foreign_keys=off;
+
+CREATE TABLE hm_user_session_new (
```
kroky, line 3: Isn't there an alter table statement in sqlite?

database/migrations/sqlite/20241209010200_add_lock_columns.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE hm_user_session 
+ADD COLUMN hm_version INT DEFAULT 1;
+
+ALTER TABLE hm_user_session 
+ADD COLUMN lock INT DEFAULT 0;
```
kroky, line 5: Why lock is only in sqlite migration? I think the migrations need to be the same for all drivers/engines.

database/mysql_schema.sql
```diff
@@ -0,0 +1,18 @@
...
+    username VARCHAR(255), 
+    hash VARCHAR(255), 
+    PRIMARY KEY (username)
+);
+
+CREATE TABLE IF NOT EXISTS hm_user_session (
+    hm_id VARCHAR(255), 
+    data LONGBLOB, 
+    date TIMESTAMP, 
+    PRIMARY KEY (hm_id)
```
kroky, line 11: I don't see the hm_version here. This file needs to have the complete up-to-date db schema structure.

database/migrations/sqlite/20241209010200_add_lock_columns.sql
kroky: Ok, I see, sqlite uses artificial locking via a column. Ignore this msg.

database/mysql_schema.sql
Shadow243: I wanted to share the approach I took for handling the recent updates. I created the initial migration to add the hm_version and lock columns specifically for SQLite. The original database schema has been kept unchanged in DB_DRIVEZ_schema.sql. For new updates, I’ve been creating individual migrations, such as 20241209010200_add_hm_version_columns.sql.

This way, users with an existing database will only need to apply the migration to add the missing hm_version and lock columns for SQLite. I deliberately didn’t include these columns in DB_DRIVEZ_schema.sql, as I considered this more of an update for users cloning the latest version.

However, if we decide to keep all columns in DB_DRIVEZ_schema.sql, should we still maintain the new migration as well? My thought was that everything introduced, including in this PR, should be treated as updates. Please let me know if this aligns with your expectations or if adjustments are needed.

Shadow243: @kroky  I updated, can your review please !

PR #1381, opened 2024-11-23
style(cypht): scope menu toggling to not interfere with Tiki elements

josaphatim:
Click on menu button is having impact on Tiki menu as same html tags involved are shared in both Cypht and Tiki.

### DIFF
--- a/modules/core/site.css
+++ b/modules/core/site.css
@@ -1461,7 +1461,7 @@ nav.collapsed .folders {
   margin-top: 50px;
 }
 
-nav.collapsed li {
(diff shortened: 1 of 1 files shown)


jacob-js:
@josaphatim, to achieve your intended behavior, you need to add the `cypht-layout` class here: https://github.com/cypht-org/cypht/blob/master/modules/core/site.css#L1464

josaphatim:
@jacob-js I like the proposed solution. Other changes were committed by error. It's a work in progress

Issue #1383, opened 2024-11-25
🐛 [Bug] SQLite support broken


indridieinarsson:
## 🐛 Bug
When running cypht with sqlite backend (in my opinion a very nice db backend for self-hosting for personal use), cypht simply refuses to start up:

```
✔ Container cypht-cypht-1  Created                                                                                                                                                                                                                                        0.1s
Attaching to cypht-1
cypht-1  |
cypht-1  | Unsupported DB driver: sqlite
cypht-1 exited with code 1
```

PR #1384, opened 2024-11-25
Fix database driver check where sqlite was forgotten

indridieinarsson:
## 🍰 Pullrequest
cypht would not work with sqlite database.

### Issues
#1383

### DIFF
--- a/scripts/setup_database.php
+++ b/scripts/setup_database.php
@@ -56,6 +56,8 @@ if ($db_driver == 'mysql') {
         error_log('The following required PostgreSQL extensions are missing: ' . implode(', ', $missing_extensions) . ". Please install them.\n");
         exit(1);
     }
+} elseif ($db_driver == 'sqlite') {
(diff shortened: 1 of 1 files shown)


scripts/setup_database.php
```diff
@@ -56,6 +56,8 @@
         error_log('The following required PostgreSQL extensions are missing: ' . implode(', ', $missing_extensions) . ". Please install them.\n");
         exit(1);
     }
+} elseif ($db_driver == 'sqlite') {
+    error_log("Nothing to be done for db driver: {$db_driver}");
```
josaphatim, line 60: Hello @indridieinarsson 

Thank you for this. But I think instead of writing to error log while there is no error we can just can the else to else if and test if $db_driver != 'sqlite'.

Thank you!

marclaporte:
Thank you @indridieinarsson and @josaphatim

PR #1379, opened 2024-11-22
feat(other): update issue templates

ulfgebhardt:
## Pullrequest

Update issue templates

image

### Motivation
Since I am invested in testing cypht lately, i figured the issue templates are now 6 years old and have evolved on my end a bit.
So here is an update.

### What has changed?

- Templates are more simple so developers are actually tempted not to just replace the whole contents but fill out the data (see pullrequest template especially)
- The title of a new issue will be preset. This results in well organized issue lists with visual cues what type of issue it is with no effort for the issue creator
- There is new types of issues: epic, devops, refactor, release
- The suggestion type is gone
- There is a new workflow which checks if a PR title is well formed. This needs some getting used to, but results in a nice changelog based on PR titles.

You can explore how things would look like here: https://github.com/ulfgebhardt/issue-templates

Example of a well formed changelog generated by https://github.com/CookPete/auto-changelog
image

### Issues

- [X] None

### Todo

- [x] Decided if all types should be part of the repo - is epic a thing here? Is devops needed?
- [ ] Do you want the PR title workflow?
- [x] Create all labels for the types decided for - see https://github.com/ulfgebhardt/issue-templates?tab=readme-ov-file#labels

PR #1389, opened 2024-11-26
refactor(workflow): pr title validation update

ulfgebhardt:
## 🍰 Pullrequest

pr title validation update requested by a comment on the implementing PR https://github.com/cypht-org/cypht/pull/1379#issuecomment-2500269681

Adds 4 new types for pr scopes observed in cypht:
- frontend
- backend
- frontend/backend
- backend/frontend

Furhtermore a comprehensive guide was added to the pull request template to make it easier for people to fullfil this requirement.

### Example

image

### Issues

- requirements https://github.com/cypht-org/cypht/pull/1379#issuecomment-2500269681
- relates https://github.com/cypht-org/cypht/pull/1379
- relates https://github.com/ulfgebhardt/issue-templates/commit/56dc4aa7847cae10de1bf1b3d0f301a519882dab

### Todo

- [ ] Verify that no scope, observed lately, is missing

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,18 @@
 <!-- You can find the latest issue templates here https://github.com/ulfgebhardt/issue-templates -->
 
+<!--
+Validate your PR title - a quick Guide
+
+A valid PR title contains a type (https://github.com/commitizen/conventional-commit-types/blob/master/index.json), a scope (https://github.com/cypht-org/cypht/blob/master/.github/workflows/test.lint.pr.yml#L31) and a title starting with a lower case letter.
+
+Here are some valid examples:
+- fix(frontend): my fix
+- feat(backend): my feature
+- docs(docker): my docs
+- refactor(frontend/backend) : my refactor
+- test(unit): my test
+-->
+
 ## 🍰 Pullrequest
 <!-- Describe the Pullrequest. Use Screenshots if possible. -->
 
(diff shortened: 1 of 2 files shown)


mercihabam:
@ulfgebhardt, thank you for your response to my comment!

Here is my feedback:

- I suggest excluding the scopes "cypht" and "other" as they do not indicate that any specific part of the project has been affected.
- `docs` sounds more appropriate as a scope compared to `docu`.
- In the proposed update, we have both `frontend/backend` and `backend/frontend`. I recommend keeping only one variation for consistency.
- These additional scopes could be useful: `modules/output`, `modules/handler`, `tests/unit`, and `tests/e2e`.

ulfgebhardt:
> @ulfgebhardt, thank you for your response to my comment!
> 
> Here is my feedback:
> 
> * I suggest excluding the scopes "cypht" and "other" as they do not indicate that any specific part of the project has been affected.
> * `docs` sounds more appropriate as a scope compared to `docu`.
> * In the proposed update, we have both `frontend/backend` and `backend/frontend`. I recommend keeping only one variation for consistency.
> * These additional scopes could be useful: `modules/output`, `modules/handler`, `tests/unit`, and `tests/e2e`.

I implemented your suggestions, except the removal of `other`. I can remove it if you want, but its always useful to give people an option to escape things in case you did not think about everything.

Furthermore I suggest to simplify your suggestions further:
- instead of `modules/output` and `modules/handler` i suggest to just use `modules` - or better  sigular `module`
- instead of `tests/unit` `tests/e2e` I suggest to use just `unit` and `e2e` since your title would be something like `test(unit): my test`.

Let me know what you think

mercihabam:
That works for me, please go ahead!

mercihabam:
Looks good!

PR #1407, opened 2024-12-19
fix(backend): ensure attachments are sent along when forwarding an email

mercihabam:


marclaporte:
Thanks!

It it's already OK in 2.x?

Issue #1402, opened 2024-12-09
🐛 [Bug] Gmail messages not shown


indridieinarsson:
## 🐛 Bug
I have an install using sqlite database, redis and memcached enabled, running as a docker image.

I have 3 different accounts, one of them gmail. The other two accounts (on two separate and different servers) work as expected, both in single-mailbox mode (under the Email tab) and in the combined views.

Gmail, on the other hand has problems. After restarting cypht, it works, showing me mails in single-mode and in the combined views.
After just a few navigations,  cypht stops showing any messages from my gmail account. 
That goes for both viewing the gmail INBOX directly (which shows no messages, although there are thousands) and also the combined views, which show no message from the gmail account, only from the other two.

It may be of importance here, that my gmail inbox is pretty huge, containing my entire email history for years. 

### Version & Environment
Rev: 412b5adbd9ada
OS:  Linux x86 / docker

indridieinarsson:
I analysed this a bit. Cypth (modules/imap/hm-imap-base.php) is successfully communicating with gmail. Getting settings, and the list of messages. All of it, the first timestamp is 2006. My guess is that it then goes on to fetch a whole lot of messages and something times out, but I haven't confirmed that yet. My feeling is that gmail is not answering imap requests very fast (throttling them even).

Now, perhaps, I should tidy up my inbox, but then again, I'm hardly the only one using the search function instead of organized folders...
It's also worth noting that (at least a couple of weeks ago) everything worked ok on a mailbox which is a mirror of my gmail box. But that one is faster.

indridieinarsson:
I looked a bit into  this. I see the following imap command being sent to gmail imap server :

`A8 UID FETCH 53418,53419,53420,53421,53422,53423,53424,53425,53429,53432,53436,53437,53438,53439,53440,53441,53442,5
3443,53444,53445 (FLAGS INTERNALDATE RFC822.SIZE X-GM-MSGID X-GM-THRID X-GM-LABELS BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES ME
SSAGE-ID X-SNOOZED)] BODY.PEEK[0.1])`

which then yields the answer : 
`A8 BAD Could not parse command`

indridieinarsson:
[...]

PR #1409, opened 2024-12-20
fix(other): gmail messages not shown

christer77:
see [issue](https://github.com/cypht-org/cypht/issues/1402)

ref: [RFC](https://www.rfc-editor.org/rfc/rfc3501#page-57)

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -901,7 +901,7 @@ if (!class_exists('Hm_IMAP')) {
             }
             $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID X-SNOOZED)]";
             if ($include_content_body) {
-                $command .= " BODY.PEEK[0.1]";
+                $command .= " BODY.PEEK[1]";
             }
             $command .= ")\r\n";
             $cache_command = $command.(string)$raw;
@@ -958,7 +958,7 @@ if (!class_exists('Hm_IMAP')) {
                                 }
                             }
                         }
-                        elseif ($vals[$i] == 'BODY[0.1') {
+                        elseif ($vals[$i] == 'BODY[1') {
                             $content = '';
                             $i++;
                             $i++;


indridieinarsson:
May I add a comment? 
The issue ( #1402 ) turns out to be really two different things. One is the inability of gmail to process `BODY.PEEK[0.1]` (but I think `BODY[0.1]` is fine, so might consider changing that one back). And I think this PR fixes that (but I'm not quite sure what info we might be missing by changing from [0.1] to [1]? )

The other thing is that in the combined view, the automatic check and renewal of the xoauth2 bearer token (which is valid for 1 hour after renewal) does not take place, and this PR doesn't fix that.

I made some code that seems to fix this part of it : https://github.com/indridieinarsson/cypht/commit/3c74c2a62a3d8808518a4cf6dd4e48ded6c034ab 
For the dedicated mailbox pages, the old code checks which mail server it is serving, and checks that server for expiry. But  the request for the combined view yields and empty list of servers so no check is performed. In this case, I just check expiry for all available servers. A bit course, but maybe harmless.

christer77:
> May I add a comment? The issue ( #1402 ) turns out to be really two different things. One is the inability of gmail to process `BODY.PEEK[0.1]` (but I think `BODY[0.1]` is fine, so might consider changing that one back). And I think this PR fixes that (but I'm not quite sure what info we might be missing by changing from [0.1] to [1]? )
> 
> The other thing is that in the combined view, the automatic check and renewal of the xoauth2 bearer token (which is valid for 1 hour after renewal) does not take place, and this PR doesn't fix that.
> 
> I made some code that seems to fix this part of it : [indridieinarsson@3c74c2a](https://github.com/indridieinarsson/cypht/commit/3c74c2a62a3d8808518a4cf6dd4e48ded6c034ab) For the dedicated mailbox pages, the old code checks which mail server it is serving, and checks that server for expiry. But the request for the combined view yields and empty list of servers so no check is performed. In this case, I just check expiry for all available servers. A bit course, but maybe harmless.

hello @indridieinarsson,

Can you push your suggestion in current PR please?

indridieinarsson:
> > May I add a comment? The issue ( #1402 ) turns out to be really two different things. One is the inability of gmail to process `BODY.PEEK[0.1]` (but I think `BODY[0.1]` is fine, so might consider changing that one back). And I think this PR fixes that (but I'm not quite sure what info we might be missing by changing from [0.1] to [1]? )
> > The other thing is that in the combined view, the automatic check and renewal of the xoauth2 bearer token (which is valid for 1 hour after renewal) does not take place, and this PR doesn't fix that.
> > I made some code that seems to fix this part of it : [indridieinarsson@3c74c2a](https://github.com/indridieinarsson/cypht/commit/3c74c2a62a3d8808518a4cf6dd4e48ded6c034ab) For the dedicated mailbox pages, the old code checks which mail server it is serving, and checks that server for expiry. But the request for the combined view yields and empty list of servers so no check is performed. In this case, I just check expiry for all available servers. A bit course, but maybe harmless.
> 
> hello @indridieinarsson,
> 
> Can you push your suggestion in current PR please?

Sorry for the late answer - I'm quite busy these days.

I honestly don't know how I can push my branch into an existing PR. My code is available in it's own branch here : https://github.com/indridieinarsson/cypht/tree/issue_1402. Feel free to merge or cherry pick what you want.

My code is not ready. It works, in the way that one we navigate to the combined mailbox, the oauth token is renewed if needed. However, if the site is left open for a while, all gmail messages will disappear after about an hour (unless the page is reloaded). The reason is that after an hour, the token has expired, and the code that checks and renews it is only run on page load. 
I was thinking this should rather be done closer to the imap module. I was thinking to catch the reply from the imap server (where it tells us that the token has expired), and trigger the renewal from there. But that's a bit more work, and I haven't found the time for it yet.

christer77:
> > > May I add a comment? The issue ( #1402 ) turns out to be really two different things. One is the inability of gmail to process  (but I think  is fine, so might consider changing that one back). And I think this PR fixes that (but I'm not quite sure what info we might be missing by changing from [0.1] to [1]? )
> > > The other thing is that in the combined view, the automatic check and renewal of the xoauth2 bearer token (which is valid for 1 hour after renewal) does not take place, and this PR doesn't fix that.
> > > I made some code that seems to fix this part of it : [indridieinarsson@3c74c2a](https://github.com/indridieinarsson/cypht/commit/3c74c2a62a3d8808518a4cf6dd4e48ded6c034ab) For the dedicated mailbox pages, the old code checks which mail server it is serving, and checks that server for expiry. But the request for the combined view yields and empty list of servers so no check is performed. In this case, I just check expiry for all available servers. A bit course, but maybe harmless.`BODY.PEEK[0.1]``BODY[0.1]`
> > 
> > 
> > hello @indridieinarsson,
> > Can you push your suggestion in current PR please?
> 
> Sorry for the late answer - I'm quite busy these days.
> 
> I honestly don't know how I can push my branch into an existing PR. My code is available in it's own branch here : https://github.com/indridieinarsson/cypht/tree/issue_1402. Feel free to merge or cherry pick what you want.
> 
> My code is not ready. It works, in the way that one we navigate to the combined mailbox, the oauth token is renewed if needed. However, if the site is left open for a while, all gmail messages will disappear after about an hour (unless the page is reloaded). The reason is that after an hour, the token has expired, and the code that checks and renews it is only run on page load. I was thinking this should rather be done closer to the imap module. I was thinking to catch the reply from the imap server (where it tells us that the token has expired), and trigger the renewal from there. But that's a bit more work, and I haven't found the time for it yet.
> 
> edit: forgot to mention @christer77 so he actually sees my message.
Normally, we have a request that allows us to do this verification.Are you sure that it wasn't your session that had expressed and that you had to reconnect?
`add_handler('ajax_hm_folders', 'imap_oauth2_token_check', true, 'imap', 'load_imap_servers_from_config', 'after');
`

christer77:
> > > > May I add a comment? The issue ( #1402 ) turns out to be really two different things. One is the inability of gmail to process  (but I think  is fine, so might consider changing that one back). And I think this PR fixes that (but I'm not quite sure what info we might be missing by changing from [0.1] to [1]? )
> > > > The other thing is that in the combined view, the automatic check and renewal of the xoauth2 bearer token (which is valid for 1 hour after renewal) does not take place, and this PR doesn't fix that.
> > > > I made some code that seems to fix this part of it : [indridieinarsson@3c74c2a](https://github.com/indridieinarsson/cypht/commit/3c74c2a62a3d8808518a4cf6dd4e48ded6c034ab) For the dedicated mailbox pages, the old code checks which mail server it is serving, and checks that server for expiry. But the request for the combined view yields and empty list of servers so no check is performed. In this case, I just check expiry for all available servers. A bit course, but maybe harmless.`` BODY.PEEK[0.1]``BODY[0.1] ``
> > > 
> > > 
> > > hello @indridieinarsson,
> > > Can you push your suggestion in current PR please?
> > 
> > 
> > Sorry for the late answer - I'm quite busy these days.
> > I honestly don't know how I can push my branch into an existing PR. My code is available in it's own branch here : https://github.com/indridieinarsson/cypht/tree/issue_1402. Feel free to merge or cherry pick what you want.
> > My code is not ready. It works, in the way that one we navigate to the combined mailbox, the oauth token is renewed if needed. However, if the site is left open for a while, all gmail messages will disappear after about an hour (unless the page is reloaded). The reason is that after an hour, the token has expired, and the code that checks and renews it is only run on page load. I was thinking this should rather be done closer to the imap module. I was thinking to catch the reply from the imap server (where it tells us that the token has expired), and trigger the renewal from there. But that's a bit more work, and I haven't found the time for it yet.
> > edit: forgot to mention @christer77 so he actually sees my message.
> > Normally, we have a request that allows us to do this verification.Are you sure that it wasn't your session that had expressed and that you had to reconnect?
> > `add_handler('ajax_hm_folders', 'imap_oauth2_token_check', true, 'imap', 'load_imap_servers_from_config', 'after'); `

@indridieinarsson

indridieinarsson:
Hello - I'm trying to get this to work. 
This issue arises (e.g.) in the combined view. The combined view pages don't have a particular mail server attached to them, which leads to them not renewing the oauth bearer token. I made a partial fix, which re-authenticates brute-force for all imap servers within the `Hm_Handler_imap_oauth2_token_check` handler

But it's only a partial fix, since the combined pages refresh the message list regularly (via ajax, so without reloading the page). After 1hr, the gmail oauth bearer token becomes invalid, so we get a statuscode 400 when retreiving the gmail message list.

At this point, I'm unsure what to do, as the handler modules seem to only run on page load, which isn't happening here. 

Could someone explain how the "handler modules" work and how I could trigger the `process` function on it from within the `authenticate` function on the `HM_Imap` class, without reloading the page?

indridieinarsson:
@christer77 @kroky 
Ok, much ado about nothing. Seems the issue is fixed with the current PR. I was testing on a version since december, where apparently snoozing messages wasn't being called when logged in through an IMAP account. I nuked my config in the meantime, so can't bisect it properly.

(the snooze worker goes on in the background and runs the oauth refresh. Snooze is part of the imap module, just as the oauth stuff, so should not be accidentally disabled)

So, with the current master, this PR fixes it, for me at least.

modules/imap/hm-imap.php
```diff
@@ -901,7 +901,7 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
             }
             $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID X-SNOOZED)]";
             if ($include_content_body) {
-                $command .= " BODY.PEEK[0.1]";
+                $command .= " BODY.PEEK[1]";
```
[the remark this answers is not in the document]
christer77: I get the idea, let me see what it can be like

modules/imap/hm-imap.php
```diff
@@ -938,9 +938,11 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
                     $x_auto_bcc = '';
                     $x_snoozed = '';
                     $count = count($vals);
+                    $head_finded = false;
```
kroky, line 941: Sorry to be nit-picky but proper past tense of the verb find is found...

modules/imap/hm-imap.php
```diff
@@ -1035,6 +1040,27 @@ public function get_message_structure($uid) {
             return $struct;
         }
 
+        private function parseEmailBody($raw_data) {
```
kroky, line 1043: This function will deal with fraction of the cases as there are numerous transfer encodings, mime details and other stuff that might break here. Can you use zbateson/mail-mime-parser that is already included in Cypht - it has methods to get the first plaintext part of html part or whatever you need. If it is html part, then, maybe you need to strip tags first before trimming?

christer77: Hello @kroky,
Here is a model of what I got as the content of the mail body:
_<0> --b1_a89fcc5955910d0b8d4ee215939294c3 Content-Type: text/plain; charset="UTF-8" Content-Transfer-Encoding: quoted-printable Hello Rakel, We are happy to have you among us..._

here the message that must be returned by the method is: _Hello Rakel, We are happy to receive you among us..._

Using zbateson/mail-mime-parser this way:
```
$parser = new MailMimeParser();
$parser->parse($content, false);
```

At the exit I have the same value that was passed as a parameter, in short, the method does not parse the content of my message. that's why I created the _parseEmailBody()_ method
Can you have more ideas on this please?

christer77: Okay, 
I'm doing some investigation on this

christer77: Compared to the last investigation, we found that the PREVIEW command
is not supported by the Gmail server or Migadu, we tested via the method available in Cypht including:
$this->is_supported( 'PREVIEW' ).

That's how we worked with BODY. PEEK[TEXT]<0.500> while subtracting
The first 3 characters that this command returns: <0>.

kroky: Ok, it is much better now, thanks. Only one last thing - can you conditionally strip the first 3 characters? I suspect not all mail servers will return <0> - it looks like some control character sequence.

Issue #1408, opened 2024-12-20
🐛 [Bug] date of last udpate of Cypht shows the date of the merge request instead of the date of the commit


marclaporte:
## 🐛 Bug

In the admin panel, we can check the version/revision. This is very handy to check how long ago Cypht was updated. There is also a link to the commit. The problem is that Cypnt reports the date on https://github.com/cypht-org/cypht/pull/1390 which is a few weeks earlier than https://github.com/cypht-org/cypht/commit/590000803a8431755607de30ca150c3e78122013 so it gives the impression that Cypht is a few weeks older than it is.

We had the same problem for Tiki Wiki CMS Groupware. The code base is the same license (LGPL 2.1) so code can be copied over if it helps.

### Version & Environment

Rev: Provided above

OS:  Not platform dependant

marclaporte:
Fixed via https://github.com/cypht-org/cypht/pull/1417

PR #1417, opened 2025-01-02
fix(other): display commited date instead of the pull request creation date

Baraka24:
### Issues

- [1408](https://github.com/cypht-org/cypht/issues/1408)

### Todo

- [X] make an additional call to the GitHub API to check if the commit is part of a pull request and if that pull request has been merged. If the PR was merged, we use the merge date (merged_at)

### DIFF
--- a/modules/developer/modules.php
+++ b/modules/developer/modules.php
@@ -51,7 +51,7 @@ class Hm_Handler_process_server_info extends Hm_Handler_Module {
         }
 
         if ($commit_hash != '-') {
-            // Get right commit date (not merge date) if not a local commit
+            // Get the correct commit date (merged PR date or commit date)
             $ch = Hm_Functions::c_init();
             if ($ch) {
                 Hm_Functions::c_setopt($ch, CURLOPT_URL, 'https://api.github.com/repos/cypht-org/cypht/commits/'.$commit_hash);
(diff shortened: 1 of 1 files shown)


modules/developer/modules.php
```diff
@@ -182,6 +186,27 @@ protected function output() {
...
         return '';
     }
+
+    /**
+     * Fetches the commit date based on the commit hash
+     * @param string $commit_hash The commit hash
+     * @return string The commit date
+     */
+    protected function getCommitDate($commit_hash) {
+        $command = 'git log -1 --format=%cd ' . escapeshellarg($commit_hash);
```
kroky, line 196: Cypht might not have a local git repo - there are other distribution channels than `git pull`... see Hm_Handler_process_server_info for more info on how it gets commit hash and tries to get the date from github API endpoint. I think you should change that class rather than add more code here.

Baraka24: Okay noted.

Baraka24: All good now. @kroky please review.

modules/developer/modules.php
```diff
@@ -71,9 +71,10 @@ public function process() {
         $res['branch_name'] = $branch_name;
         $res['commit_hash'] = $commit_hash;
         $res['commit_url'] = $commit_url;
+        $commit_date_output = shell_exec("git show -s --format=%ci {$commit_hash}");
```
kroky, line 74: There might be no git here - cypht can be deployed as a package and not through a git checkout. You cannot relay on this command to work. Look above - when package, you get the commit hash and then a call to github API to get more details on the commit. Can you get what you need from the github API?

Baraka24: Thank you for your feedback. Based on your suggestion, I have modified the logic to exclusively rely on the GitHub API to retrieve the commit date. If the commit is part of a pull request, I now fetch the merge date (merged_at) from the pull request details, ensuring that we capture the date when the pull request was merged rather than the commit's creation date. If the commit is not associated with a pull request, it falls back to the original commit date. This approach ensures accuracy regardless of whether Cypht is deployed as a package or through a Git repository.

Issue #1370, opened 2024-11-19
Junk/Spam button


ulfgebhardt:
## 🚀 Feature

Junk/Spam button

The short menu on top does no offer a Junk/Spam button, even tho the imap folders configure an Junk folder.

![image](https://github.com/user-attachments/assets/20a91803-1756-4757-bc2a-c9a965eef9a2)

### Design, Layout, Architecture

The menu should provide a Spam/Junk button whenever an Junk folder is specified in the corresponding imap folders.
The requested feature is almost the same as the Archive functionality and should be located next to it.

Baraka24:
@ulfgebhardt can you please confirm this (#1420 ) fits your requirements so we close this issue.

PR #1420, opened 2025-01-05
feat(other): add Junk button when Junk folder is configured for an IMAP Server

Baraka24:
### Issues

- [Related issue](https://github.com/cypht-org/cypht/issues/1370)


modules/imap/handler_modules.php
```diff
@@ -1205,6 +1151,53 @@ public function process() {
             }
         }
     }
+
+    private function perform_action($mailbox, $action_type, $uids, $folder, $specials, $server_details) {
```
kroky, line 1155: I think this method changes the behavior of the old code in imap_message_action... e.g. original_folder_setting was only used for archive operations, now you use it for all special folder types of operations (e.g. delete). If you refactor this code please make sure you don't introduce regressions or unintentional changes.

Baraka24: Thanks for the feedback! I've thoroughly tested the refactor, and there are no unintentional changes or regressions. If any issues arise, we'll address and fix them promptly. The refactor was done to reduce code repetition while maintaining the current functionality.

modules/imap/handler_modules.php
```diff
@@ -1205,6 +1151,53 @@ public function process() {
...
+            }
+            if (!$mailbox->message_action($folder_name, 'MOVE', $uids, $special_folder)['status']) {
+                $error = true;
+            } else {
+                foreach ($uids as $uid) {
+                    $moved[] = sprintf("imap_%s_%s_%s", $server_details['id'], $uid, $folder);
+                }
+            }
+        } else {
+            if (!$mailbox->message_action($folder_name, mb_strtoupper($action_type), $uids)['status']) {
```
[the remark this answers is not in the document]
Baraka24: Thanks for pointing that out! I'll look into implementing the JUNK action for the EWS side. If I run into any issues, I'll reach out for your help.

PR #1425, opened 2025-01-17
fix(backend): jmap config with smtp server

Shadow243:
Refactored server connection logic to handle cases where JMAP and SMTP coexist.

Currently, JMAP is added alone when `$onlyJmap` is true. The update replaces IMAP with JMAP in this case while allowing the possibility to check and add SMTP if `$isSender` is true. 

Screenshot 2025-01-17 at 11 32 10

### DIFF
--- a/modules/core/handler_modules.php
+++ b/modules/core/handler_modules.php
@@ -1071,6 +1071,21 @@ class Hm_Handler_quick_servers_setup extends Hm_Handler_Module {
             ] = $form;
 
             /*
+            *  Connect to SMTP server if user wants to send emails
+            */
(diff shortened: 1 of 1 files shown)


modules/core/handler_modules.php
```diff
@@ -1079,6 +1079,18 @@ public function process() {
                     return;
                 }
 
+                if($isSender){
```
josaphatim, line 1082: I think the best way is to put this block of code out of its parent if so we avoid duplication.

Shadow243: Done

Issue #1046, opened 2024-05-24
Subfolders do not appear in IMAP accounts with the XEAMS mail server and Charter.net (Charter Communications) aka Spectrum.net


truthsword:
## 💬 Question
I manually set up IMAP/SMTP/Profile for a Charter Spectrum email account. Subfolders under the Archive folder do not appear wither in the left hand columns, or if I attempt to select a subfolder `Archive/2024` as the default archive folder. All I see is Archives.

What have I missed?
Cypht (in docker):
![](https://imgur.com/HdwTrps.png)

Thunderbird:
![](https://imgur.com/3TbTuHQ.png)

marclaporte:
> Cypht (in docker):

FYI: https://github.com/cypht-org/cypht/issues/1047

IrAlfred:
Hello @truthsword, I've tried to reproduce this issue and on my side everything is working well as you can see on the screenshot below: 

![display_subfolders2](https://github.com/cypht-org/cypht/assets/44422833/5b30fcaa-19a0-47a4-8fa9-0a42c5305aa7)

Can you specify the exact version you are using so that we can focus on the details of that version ?

truthsword:
I've used these versions...
```
image: jonocodes/cypht:2.0.1-docker-wip
image: jonocodes/cypht:latest
```
with same result. This is specific to Cypht docker container and IMAP server `mobile.charter.net`. Two other IMAP servers I tested show subfolders.

IrAlfred:
I think this error you are facing may be directly related to the Charter Communications mail server you are using. I've tested with 2 other mail servers and everything is working. Unfortunately it seems that it is not freely accessible for all to allow us to create an account and test on our side to confirm this.

truthsword:
> Is it possible to supply @IrAlfred with a test account without any confidential data?

My allocation of available email accounts is in use. Sorry.

> I think this error you are facing may be directly related to the Charter Communications mail server you are using.

That may be, however it works fine in Thunderbird, RainLoop (docker container), and SnappyMail (docker container which autoconfigured this domain)., and multiple Android apps.

marclaporte:
> My allocation of available email accounts is in use. Sorry.

Understood. @IrAlfred can you think of some debug code that @truthsword could try? Maybe you two could do a screenshare using some folders with non-confidential data?

> That may be, however it works fine in Thunderbird, RainLoop (docker container), and SnappyMail (docker container which autoconfigured this domain), and multiple Android apps.

[...]

Issue #556, opened 2022-02-27
Imap folders and sub-folders don't appear in some cases


smitgd:
## 🐛 Bugreport

In my main imap account I only see the top-level folders. Several top-level folders definitely have sub folders, including INBOX, but a ```+``` does not appear beside them. On other accounts the sub-folders seem to be working OK. Not sure why there is  difference between accounts. All folders and sub-folders on this account have us-ASCII names.

I also noticed that when I was first setting up my accounts, all folders on accounts with folder names containing non-US-Ascii characters would not appear until I renamed the folders (in thunderbird) with US-Ascii chars. Then all the folders appeared. However, after the folders appeared and I changed a single folder name back to a non-US-Ascii name, the folders still appeared and showed the UTF-8 character. 

I don't know if these issues are related.

marclaporte:
@smitgd 

Sorry you are running into issues.

Can you create an account with no confidential information to demonstrate the issue, and share that with one of our developers?

Please join us at https://gitter.im/cypht-org/community so we can coordinate.

Thanks!

marclaporte:
@smitgd

Please retest, as a lot has changed since you reported this issue. Notably, we now have 3 active branches and recently released Cypht 2.0.0
- https://github.com/cypht-org/cypht/releases/tag/v2.0.0
- https://github.com/cypht-org/cypht/wiki/Lifecycle

truthsword:
Just saw this thread... similar report [here](https://github.com/cypht-org/cypht/issues/1046)

marclaporte:
Good catch @truthsword !

marclaporte:
@smitgd who is your ISP?

PR #1427, opened 2025-01-18
fix(other): subfolders not appearing in IMAP accounts with XEAMS mail server

IrAlfred:
Related issues https://github.com/cypht-org/cypht/issues/1046 https://github.com/cypht-org/cypht/issues/556

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -526,14 +526,15 @@ if (!class_exists('Hm_IMAP')) {
          * @param bool $lsub flag to limit results to subscribed folders only
          * @return array associative array of folder details
          */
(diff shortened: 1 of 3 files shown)


modules/imap/hm-imap.php
```diff
@@ -526,14 +526,15 @@ public function get_special_use_mailboxes($type=false) {
          * @param bool $lsub flag to limit results to subscribed folders only
          * @return array associative array of folder details
          */
-        public function get_mailbox_list($lsub=false, $mailbox='', $keyword='*') {
+        public function get_mailbox_list($lsub=false, $mailbox='', $keyword='*', $children_capability=true) {
```
kroky, line 529: This method is called in several different places including tests. Can you make sure all of them work with the updated functionality?

PR #1435, opened 2025-01-28
fix(frontend): get_list_block_sieve function not callable if sievefilters module is not enabled

christer77:
[issue](https://github.com/cypht-org/cypht/pull/1265#discussion_r1918019235)

### DIFF
--- a/modules/sievefilters/site.js
+++ b/modules/sievefilters/site.js
@@ -1003,6 +1003,25 @@ function sieveFiltersPageHandler() {
     load_sieve_filters('ajax_account_sieve_filters');
 }
 
+function get_list_block_sieve() {
+    sessionStorage.removeItem('list_blocked');
+    let detail = Hm_Utils.parse_folder_path(hm_list_path());
+    let list_blocked_senders = [];
+    if (getPageNameParam() == 'message_list') {
+        Hm_Ajax.request(
+            [
+                { name: 'hm_ajax_hook', value: 'ajax_list_block_sieve' },
+                { name: 'imap_server_id', 'value': detail.server_id},
+            ],
+            function (res) {
+                if (res.ajax_list_block_sieve) {
+                    sessionStorage.setItem('list_blocked', res.ajax_list_block_sieve);
+                }
+            }
+        );
+    }
+};
+
 $(function () {
     $(document).on('change', '#block_action', function(e) {
         if ($(this).val() == 'reject_with_message') {
--- a/modules/core/js_modules/route_handlers.js
+++ b/modules/core/js_modules/route_handlers.js
@@ -82,7 +82,7 @@ function applyMessaleListPageHandlers(routeParams) {
         Hm_Message_List.toggle_rows();
     });
 
-    get_list_block_sieve();
+    if (window.get_list_block_sieve) get_list_block_sieve();
 
     if (routeParams.list_path === 'github_all') {
         return applyGithubMessageListPageHandler(routeParams);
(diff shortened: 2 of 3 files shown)


modules/sievefilters/site.js
```diff
@@ -977,6 +977,26 @@ function sieveFiltersPageHandler() {
     });
 }
 
+function get_list_block_sieve() {
+    sessionStorage.removeItem('list_blocked');
+    var detail = Hm_Utils.parse_folder_path(hm_list_path());
```
mercihabam, line 982: 1. You should avoid using `var` at all costs because it can lead to unintended global variable pollution.
2. `hm_list_path()` and `hm_page_name()` are deprecated and do not always guarantee accurate results. Utility functions in [modules/core/navigation/utils.js](https://github.com/cypht-org/cypht/blob/master/modules/core/navigation/utils.js) should be used instead.
3. What happens to the code calling `get_list_block_sieve()` when the module is disabled?
4. Did the pull request title miss something, and why does it use the tag `other`? Shouldn't `frontend` be appropriate as the tag?

christer77: The objective is to getthe list of blocked addresses, when the mobule is deactivated, nothing happens, i.e. the function no longer get this list

mercihabam: When the module is deactivated, `get_list_block_sieve` will be undefined, thus, `get_list_block_sieve()` will throw an error in the handler function.

mercihabam:
It looks like some words are being truncated in your pull request title. Make sure to expand the title fully so that all words are visible. If necessary, you can edit the title to be more concise while still conveying the full meaning.

mercihabam:
Also, please test every change you make. Does the current code still produce the intended behavior?

marclaporte:
@christer77 was unavailable but he is now back.

modules/sievefilters/modules.php
```diff
@@ -1377,26 +1377,29 @@ public function process() {
...
             return;
         }
-
-        Hm_IMAP_List::init($this->user_config, $this->session);
-        $imap_account = Hm_IMAP_List::get($form['imap_server_id'], true);
         
-        $factory = get_sieve_client_factory($this->config);
-        try {
-            $client = $factory->init($this->user_config, $imap_account);
+        if ($this->module_is_supported('sievefilters') && $this->user_config->get('enable_sieve_filter_setting', DEFAULT_ENABLE_SIEVE_FILTER)) {
```
mercihabam, line 1381: These checks do not belong here. If that code is executed, the sievefilters module will already be supported and enabled. My emphasis was on the browser code, particularly the `get_list_block_sieve()` function, which is called in the messageList page handler; see line 85 in [modules/core/js_modules/route_handlers.js](https://github.com/cypht-org/cypht/blob/master/modules/core/js_modules/route_handlers.js#L85). This function executed successfully because you had it previously defined in the `imap/site.js`, it kept being executed but not with successful result when the sievefilters module was disabled. You moved it to the right location, but now, if the sievefilters module is disabled (preventing the browser from reading any code from `modules/sievefilters/**/*`), the messageList page handler will throw an error because `get_list_block_sieve()` is not defined.

We do not have an actual way to detect which module is disabled via the JavaScript code, so what we do is to ensure the function is globally defined before it gets executed. Here is an [example](https://github.com/cypht-org/cypht/blob/master/modules/core/js_modules/route_handlers.js#L46) with the savedSearch page handler.

christer77: Oh yes, I understand

Issue #1434, opened 2025-01-28
Warnings with PHP 8.3


s22-tech:
After successfully installing Cypht under 8.3, I'm seeing these warnings.

This warning gets produced every minute:

`PHP Warning:  Undefined array key "list_page" in /path/to/cypht/modules/imap/handler_modules.php on line 1331`

I changed this line:

`$list_page = (int) $this->request->get['list_page'];`

to:

`$list_page = (int) $this->request->get['list_page'] ?? null;`

but it still produced the same warning.  So I changed it to this:

`$list_page = intval($this->request->get['list_page'] ?? null);`

and the warnings stopped.

Same with this one:

PHP Warning:  Undefined array key "list_page" in /path/to/cypht/modules/imap/handler_modules.php on line 2089

`$list_page = intval($this->request->get['list_page'] ?? 1);`

I didn't experience a problem with other similar lines in this file, but you may want to do the same thing to lines 734 and 1270.

---

When navigating to the Snoozed screen, "Server Error" shows on the screen twice, and also produces this error:

```
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /path/to/cypht/modules/imap/functions.php:1612
Stack trace:
#0 /path/to/cypht/modules/imap/handler_modules.php(2119): getCombinedMessagesLists()
#1 /path/to/cypht/lib/modules_exec.php(177): Hm_Handler_imap_folder_data->process()
#2 /path/to/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()
#3 /path/to/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules()
#4 /path/to/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
#5 /path/to/cypht/site/index.php(53): Hm_Dispatch->__construct()
#6 {main}
  thrown in /path/to/cypht/modules/imap/functions.php on line 1612
```

---

When navigating to any screen, except Calendar or Compose, "Server Error" will flash briefly on screen but it doesn't seem to produce an error.

---

And here are 2 more that I couldn't resolve:

PHP Warning:  Undefined variable $imap in /path/to/cypht/modules/imap/handler_modules.php on line 1149

PHP Warning:  Attempt to read property "folder_state" on null in /path/to/cypht/modules/imap/handler_modules.php on line 1149

marclaporte:
[...]

PR #1436, opened 2025-01-28
fix(other): php warning: undefined array key 'list_page'

christer77:
[issue](https://github.com/cypht-org/cypht/issues/1434)

### DIFF
--- a/modules/imap/setup.php
+++ b/modules/imap/setup.php
@@ -462,5 +462,6 @@ return array(
         'ceo_suspicious_terms' => FILTER_DEFAULT,
         'ceo_rate_limit' => FILTER_VALIDATE_INT,
         'filter_type' => FILTER_DEFAULT,
+        'list_page' => FILTER_DEFAULT,
     )
 );


modules/imap/setup.php
```diff
@@ -462,5 +462,6 @@
         'ceo_suspicious_terms' => FILTER_DEFAULT,
         'ceo_rate_limit' => FILTER_VALIDATE_INT,
         'filter_type' => FILTER_DEFAULT,
+        'list_page' => FILTER_DEFAULT,
```
Shadow243, line 465: According to the issue(https://github.com/cypht-org/cypht/issues/1434), the retrieval is done via GET, but here you have added the `list_page` key to `allowed_post`. Does this really solve the problem?

This is how we get data from the request in POST: `$this->request->post['key']` but in this case this is what is used `$this->request->get['list_page']`

christer77: If we added this key in _allowed_post_, it's just because the request is sent via POST. In reference the screenshots below:

PR-1436-01

PR-1436-02

PR #1443, opened 2025-02-10
fix(frontend/backend): address all blocking issues

mercihabam:
In tandem with: https://gitlab.com/tikiwiki/tiki/-/merge_requests/6934

- [x] The filter component on mobile overlaps the content title
- [x] The Tiki bottom menu is mostly hidden from the left side
- [x] Errors “could not perform action…” when navigating to the info page
- [x] Git revision raises failed to fetch
- [x] Removing an imap/smtp server does not show it as deleted
- [x] Add an email, the label “Select an E-mail provider” is below the select element. This is also the case for the folders page
- [x] Adding a server does not refresh the state
- [x] Import accounts using yaml/csv file - Downloading a sample fails with the message “File wasn’t available on site”
- [x] Search page. Select filters do not occupy enough space for their content to be readable.
- [x] Selecting an email row from the search page, the actions component pops over the content title, making the whole thing unreadable.
- [x] When hitting the action “Read/Unread/Flag/Archive/Delete” the offline error is shown and the action is undone visually. This also happens when deleting a folder.
- [x] The refresh button on the search page does not do anything.
- [x] Deleting saved searches does not work
- [x] Saving a draft message shows the offline error
- [x] Listing message in the draft folder shows the offline error
- [x] Everything/Junk/Flagged/Sent/Snoozed/Trash/Unread view as well.
- [x] After performing a search in a folder, navigating between the paginated results, and then clearing the search, the pages keep the lists from the search result.
- [x] When filtering between read/unread/flagged/… the paginated lists do not get updated.
- [x] When opening an unread message, the counter does not get updated.
- [x] Message view - create tracker item/move to tracker does not do anything unless the window is reloaded
- [x] Snooze - Pick a date does not show a date picker
- [x] Deleting a message reloads the browser to the next message, but the prev button shows the title of the previously deleted message
- [x] The search page on mobile does not fully show the filter inputs
- [x] Server status and capabilities not shown
- [x] Failure when navigating to the filters page
- [x] ~~A bunch of errors “Sieve: Error while trying to connect to ManageSieve” when navigating to the blocklist page~~ (They are legitimate because they warn the user that Sieve is not authenticated).


marclaporte:
Thank you for collecting such a nice thorough list of issues in Cypht. This is getting us closer to Cypht 2.5.0

When it comes time to merge, I suppose we will not squash them, right?
https://en.m.wikipedia.org/wiki/Atomic_commit

mercihabam:
We can squash the commits while keeping the commit history intact, as each issue is addressed in its own commit.

marclaporte:
This is big :-)

@kroky please review when it gets out of draft.

kroky (approved):

modules/imap/site.js
```diff
@@ -133,6 +133,10 @@ var imap_delete_message = function(state, supplied_uid, supplied_detail) {
             {'name': 'folder', 'value': detail.folder}],
             function(res) {
                 if (!res.imap_delete_error) {
+                    const listPath = getParam('list_parent') || getListPathParam();
+                    const store = new Hm_MessagesStore(listPath, Hm_Utils.get_url_page_number(), `${getParam('keyword')}_${getParam('filter')}`);
+                    store.load();
+                    store.removeRow(uid);
                     if (Hm_Utils.get_from_global('msg_uid', false)) {
                         return;
                     }
```
josaphatim, line 142: @mercihabam This line and the other below should also be removed. Also remove remove_from_cached_imap_pages and check the condition above, if it is necessary then it should go after the changes. Thank you!

PR #1444, opened 2025-02-11
fix(frontend): remove deleted message from store

josaphatim:
Fixed showing a deleted message in next read message. And also deletes it in combined views.

### DIFF
--- a/modules/imap/site.js
+++ b/modules/imap/site.js
@@ -136,13 +136,15 @@ var imap_delete_message = function(state, supplied_uid, supplied_detail) {
             {'name': 'imap_msg_uid', 'value': uid},
             {'name': 'imap_server_id', 'value': detail.server_id},
             {'name': 'folder', 'value': detail.folder}],
-            function(res) {
+            async function(res) {
                 if (!res.imap_delete_error) {
                     if (Hm_Utils.get_from_global('msg_uid', false)) {
                         return;
                     }
-                    var msg_cache_key = 'imap_'+detail.server_id+'_'+getMessageUidParam()+'_'+detail.folder;
-                    remove_from_cached_imap_pages(msg_cache_key);
+                    const store = new Hm_MessagesStore(getListPathParam(), Hm_Utils.get_url_page_number());
+                    await store.load(false, true);
+                    store.removeRow(getMessageUidParam());
+
                     var nlink = $('.nlink');
                     if (nlink.length && Hm_Utils.get_from_global('auto_advance_email_enabled')) {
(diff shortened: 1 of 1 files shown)


mercihabam (commented):
@josaphatim you only needed to make a call to `store.removeRow('uid')`. I can't figure out why you needed all of the changes made.

josaphatim:
@mercihabam store.removeRow was the easiest solution but its does not remove the message wherever it can be found.

The message must also be removed in combined views. For you to understand you can go to everything page then memorize one message. Go to it’s respective inbox, open the message you memorized, Delete it then go back to everything page. Without Removing from combined views you Will see that the message is listed while deleted.

mercihabam:
Each page uses the same message storage class, so you can call `removeRow()` to each appropriate instance, and the message will be deleted. However, it's only necessary to remove it from the actual page's store because the others are updated whenever a user navigates to them.

josaphatim:
https://github.com/user-attachments/assets/fd9001c7-8ab9-442a-97e7-8d18c782f7a7

Do you see why it is necessary to remove in all combined views ?

mercihabam:
Sorry @josaphatim, I don't see anything that contradicts what I stated earlier.

josaphatim:
Did you see how many seconds it took to disappear? This is not user friendly that why I deleted it because before. User can interact (click on the email) while it is already deleted and this should never happen

mercihabam:
I'm trying my best to keep you in the context.

> Did you see how many seconds it took to disappear? This is not user friendly that why I deleted it because before. User can interact (click on the email) while it is already deleted and this should never happen

That does not justify all the changes you have made. `store.removeRow('uid')` would be sufficient.

Furthermore, I advocate for that action (removing the row from each possible page, as some pages reload noticeably slowly) not being necessary, since we have already identified that slowness as a performance issue—which I will personally address.

kroky:
If @mercihabam is going to address reloading of other page's stores to fix the performance issue then let's keep this PR simple and use only teh store.removeRow('uid') method...

mercihabam:
Sorry for the oversight, @josaphatim! This has already been addressed as part of the recent fixes I made. You can check it [here](https://github.com/cypht-org/cypht/pull/1443/commits/1d73a94dd18c1c5d4841b5a959c6783a8be1e729).

josaphatim:
Great! feel free to close

josaphatim:
@mercihabam great! I I left a comment to the other https://github.com/cypht-org/cypht/pull/1443

PR #1447, opened 2025-02-12
fix(other): save settings displaying wrong message

IrAlfred:
This PR addresses the issue where the application incorrectly displays a "Settings saved" message even when the save operation failed due to permission or folder-related issues. The changes ensure that the application only displays success messages when the save operation is confirmed to be successful


modules/core/functions.php
```diff
@@ -357,27 +357,31 @@ function max_source_setting_callback($val) {
...
+            if ($logout) {
+                $handler->session->destroy($handler->request);
+                Hm_Msgs::add('Saved user data on logout', 'info');
+                Hm_Msgs::add('Session destroyed on logout', 'info');
+            }
+            else {
+                Hm_Msgs::add('Settings saved', 'info');
+            }
         }
+    } catch (Exception $e) {
```
kroky, line 383: Catching here is OK but there are other places in Cypht that call save method on user config - you should handle the exception there as well.

lib/config.php
```diff
@@ -265,13 +265,20 @@ public function reload($data, $username = false) {
...
         $removed = $this->filter_servers();
         if (!$this->crypt) {
             $data = json_encode($this->config);
         } else {
             $data = Hm_Crypt::ciphertext(json_encode($this->config), $key);
         }
-        file_put_contents($destination, $data);
+        $result = file_put_contents($destination, $data);
+        if ($result === false) {
+            throw new Exception("Permission denied for the \"Users\" folder.\nPlease contact your site administrator.");
```
kroky, line 280: Might not be related to permission - might be out of disk space or a bunch of other reasons. Please update the error message accordingly - something like "Unable to write user config data - please check Cypht setup."

Issue #1450, opened 2025-02-15
🐛 [Bug] Hm_Handler_send_scheduled_messages causes "Server Error" message


indridieinarsson:
## 🐛 Bug
On current master, I get regular "Server Error" popups in the browser, originating from the scheduled message ajax call.

Enabling the debug php.ini, I get the following in the php log:
```
cypht-1  | 127.0.0.1 -  15/Feb/2025:22:18:03 +0000 "POST /index.php" 200
cypht-1  | 172.22.0.12 - - [15/Feb/2025:22:18:03 +0000] "POST /?page=message_list&list_path=combined_inbox HTTP/1.1" 200 56 "-" "Mozilla/5.0 (X11; Linux x86_64; rv:135.0) Gecko/20100101 Firefox/135.0"
cypht-1  | NOTICE: PHP message: PHP Fatal error:  Uncaught Error: Call to a member function connect() on null in /usr/local/share/cypht/modules/core/hm-mailbox.php:50
cypht-1  | Stack trace:
cypht-1  | #0 /usr/local/share/cypht/modules/smtp/modules.php(1580): Hm_Mailbox->connect()
cypht-1  | #1 /usr/local/share/cypht/lib/modules_exec.php(177): Hm_Handler_send_scheduled_messages->process(Array)
cypht-1  | #2 /usr/local/share/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module(Array, Array, 'Hm_Handler_send...', Array, Object(Hm_Redis_Session))
cypht-1  | #3 /usr/local/share/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules(Object(Hm_Request), Object(Hm_Redis_Session), 'ajax_send_sched...')
cypht-1  | #4 /usr/local/share/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
cypht-1  | #5 /usr/local/share/cypht/site/index.php(53): Hm_Dispatch->__construct(Object(Hm_Site_Config_File))
cypht-1  | #6 {main}
cypht-1  |   thrown in /usr/local/share/cypht/modules/core/hm-mailbox.php on line 50
cypht-1  | 127.0.0.1 -  15/Feb/2025:22:18:03 +0000 "POST /index.php" 500
[...]

PR #1451, opened 2025-02-15
fix(other): fix null reference error when smtp account has no mailbox attached.

indridieinarsson:
## 🍰 Pullrequest
Issue #1450 says it all. For some reason, I have an smtp account with no mailbox attached (I promise I didn't do anything funky with the config, all standard server configs through the web gui). This gives rise to a null reference error in the php code and an endless series of red "server error" warning dialogs in the web page.

### Issues
#1450

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -1576,18 +1576,20 @@ class Hm_Handler_send_scheduled_messages extends Hm_Handler_Module {
 
         foreach ($servers as $server_id => $config) {
             $mailbox = new Hm_Mailbox($server_id, $this->user_config, $this->session, $config);
-            if ($mailbox->connect()) {
-                $folder = 'Scheduled';
-                $ret = $mailbox->get_messages($folder, 'DATE', false, 'ALL');
-                foreach ($ret[1] as $msg) {
-                    $msg_headers = $mailbox->get_message_headers($folder, $msg['uid']);
-                    if (! empty($msg_headers['X-Schedule'])) {
-                        $scheduled_msg_count++;
-                    } else {
-                        continue;
-                    }
-                    if (send_scheduled_message($this, $mailbox, $folder, $msg['uid'])) {
-                        $scheduled_msg_count++;
+            if (!is_null($mailbox->get_connection())) {
+                if ($mailbox->connect()) {
+                    $folder = 'Scheduled';
+                    $ret = $mailbox->get_messages($folder, 'DATE', false, 'ALL');
+                    foreach ($ret[1] as $msg) {
+                        $msg_headers = $mailbox->get_message_headers($folder, $msg['uid']);
+                        if (! empty($msg_headers['X-Schedule'])) {
+                            $scheduled_msg_count++;
+                        } else {
+                            continue;
+                        }
+                        if (send_scheduled_message($this, $mailbox, $folder, $msg['uid'])) {
+                            $scheduled_msg_count++;
+                        }
                     }
                 }
             }
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -1576,20 +1576,18 @@ class Hm_Handler_send_scheduled_messages extends Hm_Handler_Module {
 
         foreach ($servers as $server_id => $config) {
             $mailbox = new Hm_Mailbox($server_id, $this->user_config, $this->session, $config);
-            if (!is_null($mailbox->get_connection())) {
-                if ($mailbox->connect()) {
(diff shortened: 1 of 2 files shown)


modules/smtp/modules.php
```diff
@@ -1576,18 +1576,20 @@ public function process() {
 
         foreach ($servers as $server_id => $config) {
             $mailbox = new Hm_Mailbox($server_id, $this->user_config, $this->session, $config);
-            if ($mailbox->connect()) {
```
mercihabam, line 1579: The connection object is not instantiated because the `$config` object lacks the type property, which the `Hm_Mailbox` class needs to determine the appropriate connection object to instantiate. Therefore, the fix should simply involve adding a type property to the `$config` object.

kroky: Right, but why is the config object missing a type - it should generally always be present...

indridieinarsson: @mercihabam  and @kroky  : yes, right. There should be a type property there. But there isn't one. I have four accounts in my setup. 3 are "normal" imap (login with username and password) - they are ok. The fourth is a gmail account. That's the one where there is no 'type' property attached to. Can't really figure out why that is.

josaphatim: @indridieinarsson 

You can add type key in these arrays: https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L197, https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L178, https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L116 and https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L128, https://github.com/cypht-org/cypht/blob/master/modules/nux/modules.php#L389

This should fix.

indridieinarsson: thanks @josaphatim . I'll take a look at this
I found out that the 'type' property is there at some point - but once we enter this code path (sending scheduled messages) it has disappeared. Only for the gmail account. 
I strongly suspect it to have something to do with the oauth2 mechanism, since that's the only difference between the gmail account and the other 3.

indridieinarsson:
Second try.
Thanks @josaphatim . It seems the 'type' property wasn't being saved with the config to the database. For some reason, gmail accounts are affected while "normal" imap accounts are not - I think it has to do with the different control flow for xoauth authentication.
Anyways, I added 'type' to the nux/modules.php as suggested, removed the gmail account and added it again. Now it works.
I think it's worth checking with office365 and outlook accounts as well

mercihabam:
@indridieinarsson, the error doesn't seem to be related only to Gmail accounts, as I'm also experiencing it with other traditional IMAP accounts.

indridieinarsson:
> @indridieinarsson, the error doesn't seem to be related only to Gmail accounts, as I'm also experiencing it with other traditional IMAP accounts.

Ok. the oauth-thing was just a hunch I had. But I did a lot of snooping around, and I could see that when reading the config for the gmail account, there was no 'type':
```
NOTICE: PHP message:  loadfromconfig  -start
cypht-1  | NOTICE: PHP message: PHP Warning:  Undefined array key "type" in /usr/local/share/cypht/lib/servers.php on line 196
cypht-1  | NOTICE: PHP message:  serverlist  type
cypht-1  | NOTICE: PHP message:  serverlist  backtrace /usr/local/share/cypht/lib/repository.php (24): Hm_IMAP_List->add()
cypht-1  | /usr/local/share/cypht/modules/imap/hm-imap.php (33): Hm_IMAP_List->initRepo()
cypht-1  | /usr/local/share/cypht/modules/imap/handler_modules.php (1702): Hm_IMAP_List->init()
cypht-1  | /usr/local/share/cypht/lib/modules_exec.php (177): Hm_Handler_load_imap_servers_from_config->process()
cypht-1  | /usr/local/share/cypht/lib/modules_exec.php (155): Hm_Module_Exec->run_handler_module()
cypht-1  | /usr/local/share/cypht/lib/dispatch.php (218): Hm_Module_Exec->run_handler_modules()
cypht-1  | /usr/local/share/cypht/lib/dispatch.php (188): Hm_Dispatch->process_request()
cypht-1  | /usr/local/share/cypht/site/index.php (53): Hm_Dispatch->__construct()
```
the `undefined key` comes from me trying to print out the type, in normal code the `type` property wouldn't be accessed at this point.

mercihabam:
@josaphatim @indridieinarsson, with this merged in, the error is still present!

josaphatim:
@mercihabam yes! You need to remove existing servers.

PR #1452, opened 2025-02-18
perf(backend): concurrently load messages for combined lists using child processes

mercihabam:


kroky:
Yes, I agree, we can continue this way and see how it goes. I assume we will hit some system issues but we can fix. What about the preference suggestion? I do think that is a good idea to keep certain environments happy - e.g. windows host, various docker setups?

mercihabam:
I don't think a new preference is necessary. This implementation has no issues with any OS, nor do we need the previous method of handling sources sequentially.

indridieinarsson:
The code from this PR does not seem to work when run standalone (i.e. not part of tiki) in a docker environment. I tried pulling it and building an image and running it. It runs, and single inboxes display messages as usual, but the combined pages are empty.  The workers/messages_list.php exits with error code 255 in all cases.

mercihabam:
@indridieinarsson, this should work in any environment! I assume it's something very basic causing it to malfunction in your container. Could you debug it further?

indridieinarsson:
@mercihabam :
With the development version of php.ini (in the docker image), I get this:
```
Warning: require(lib/framework.php): Failed to open stream: No such file or directory in /usr/local/share/cypht/modules/imap/workers/messages_list.php on line 9
cypht-1  | NOTICE: PHP message:  worker exited normally
cypht-1  | Fatal error: Uncaught Error: Failed opening required 'lib/framework.php' (include_path='.:/usr/local/lib/php') in /usr/local/share/cypht/modules/imap/workers/messages_list.php:9
cypht-1  | Stack trace:
cypht-1  | #0 {main}
cypht-1  |   thrown in /usr/local/share/cypht/modules/imap/workers/messages_list.php on line 9
cypht-1  | 127.0.0.1 -  21/Feb/2025:14:16:35 +0000 "POST /index.php" 200
cypht-1  | 172.22.0.12 - - [21/Feb/2025:14:16:35 +0000] "POST /?page=message_list&list_path=combined_inbox HTTP/1.1" 200 56 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36
(KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36"
```

mercihabam:
Seems like, it complains because it doesn't recognize the absolute path given. Could you try to prepend the required file with `/usr/local/share/cypht` on the mentioned line?

indridieinarsson:
> Seems like, it complains because it doesn't recognize the absolute path given. Could you try to prepend the required file with `/usr/local/share/cypht` on the mentioned line?

@mercihabam  : I need to prepend /usr/local/share/cypht to all the `require` statements and also have
`const APP_PATH = '/usr/local/share/cypht/';` 
then it works, sort of. Still some errors in the logs (haven't checked yet), but at least I see some messages in the combined view.

mercihabam:
Yes like that.

indridieinarsson:
@mercihabam  The worker process is still finishing with errors in some cases. Getting this now: 
OTICE: PHP message: Worker error output: PHP Fatal error:  Uncaught Error: Undefined constant "SITE_ID" in /usr/local/share/cypht/lib/cache.php:590
cypht-1  | Stack trace:
cypht-1  | #0 /usr/local/share/cypht/lib/cache.php(516): Hm_Cache->key_hash('imap67b6f947e6b...')
cypht-1  | #1 /usr/local/share/cypht/lib/cache.php(495): Hm_Cache->redis_get('imap67b6f947e6b...', false)
cypht-1  | #2 /usr/local/share/cypht/modules/imap/hm-imap.php(61): Hm_Cache->get('imap67b6f947e6b...')
cypht-1  | #3 /usr/local/share/cypht/modules/imap/hm-imap.php(67): Hm_IMAP_List::get_cache(Object(Hm_Cache), '67b6f947e6b5c')
cypht-1  | #4 /usr/local/share/cypht/modules/imap/workers/messages_list.php(46): Hm_IMAP_List::get_connected_mailbox('67b6f947e6b5c', Object(Hm_Cache))
cypht-1  | #5 {main}

mercihabam:
Those missing constants have to be defined in the worker file the same way we have declared the `APP_PATH`.

mercihabam:
> The worker process is still finishing with errors in some cases.

Are there any cases where it succeeds?

indridieinarsson:
> > The worker process is still finishing with errors in some cases.
> 
> Are there any cases where it succeeds?

well, if you don't use redis cache, then it works, at least sometimtes. I also had some cases without redis, where one mailbox failed, then nothing was displayed.

mercihabam:
@indridieinarsson, you are the one capable of diagnosing and fixing what's wrong since you are the one reporting problems.

indridieinarsson:
> @indridieinarsson, you are the one capable of diagnosing and fixing what's wrong since you are the one reporting problems.

I can look into some things. But there are still quite a few issues with this, and the proposed fixes are quite hacky. Like declaring `const APP_PATH = '/usr/local/share/cypht/';` is obviously not a proper fix, it needs to be declared in the same way as in the rest of the program. And all (or at least some) other global constants need to be propagated into the child process as well. 
@mercihabam @kroky  Might it perhaps be a good idea to backpedal with this PR, undo it and do proper fixes to the outstanding issues?

composer.json
```diff
@@ -57,7 +57,10 @@
         "twbs/bootstrap": "^5.3",
         "twbs/bootstrap-icons": "^1.11",
         "webklex/composer-info": "^0.0.1",
-        "zbateson/mail-mime-parser": "^2.4"
+        "zbateson/mail-mime-parser": "^2.4",
+        "react/promise": "^3.2.0",
```
marclaporte, line 61: Please keep this list alphabetically sorted

mercihabam:
> I can look into some things. But there are still quite a few issues with this, and the proposed fixes are quite hacky. Like declaring `const APP_PATH = '/usr/local/share/cypht/';` is obviously not a proper fix, it needs to be declared in the same way as in the rest of the program. And all (or at least some) other global constants need to be propagated into the child process as well. @mercihabam @kroky Might it perhaps be a good idea to backpedal with this PR, undo it and do proper fixes to the outstanding issues?

@indridieinarsson, the instructions I provided were not definitive fixes but rather a way to explore possible solutions for your environment. I understand that you might not be able to make it work definitively, which is fine, but please avoid making such suggestions. In cases where you’re unable to move forward, it’s better to request help instead.

marclaporte:
@mercihabam @indridieinarsson Can you please find each other https://gitter.im/cypht-org/community and organize a screenshare?

Thanks!

kroky:
@mercihabam please add a preference or I will have to revert this PR. I expect issues on some platforms as process execution is not something PHP is supporting well enough and also it is a too disturbing change to not be met with a preference. You cannot just keep saying it will work on all platforms and we take your word for granted. We need a preference for this.

mercihabam:
@kroky, I am against introducing a preference for this because React child processes do not have platform-based limitations. It would be unrealistic for me to impose such restrictions. I don’t see an issue with encountering some errors on certain platforms, but disabling it for some would only hinder us from making it adaptable for all.

I appreciate receiving issue reports from users, as this should ideally work in any environment.

mercihabam:
> @mercihabam @indridieinarsson Can you please find each other https://gitter.im/cypht-org/community and organize a screenshare?
> 
> Thanks!

@indridieinarsson messaged me that he won’t be available for a screen share anytime soon, but thanks to him, we already have enough information about the blocker. I’ll take over and implement a proper fix. Thanks again, @indridieinarsson!

mercihabam:
@kroky I’m not trying to challenge your opinion, but rather take a more rational approach. There are no known system limitations to React child processes, so disabling this implementation due to errors would only hinder its adaptability in environments where it could function efficiently.

I understand that some pages in master are mostly unusable at the moment, but this should be resolved soon. However, if you’re currently blocked from using master, please revert the relevant commit.

kroky:
We are trying to stabilize Cypht master and release stable 2.x version. I am waiting for that to happen for some time now as I need to use cypht behind corporate networks that are highly restrictive and their setups are relatively uncommon (freebsd jails for example). Subprocesses are not even allowed in some of them. Thus, it is not a matter of adapting this but making it optional. It is a performance enhancement but needs a preference, so we keep cypht capable of running in different environments and ultimately making it robust. It seems you don't trust me that we are going to have problems deploying and using your approach in some environments but I speak out of experience here. Please add a preference and make this optional. It could default to true, so user needs to specifically turn it off if something's restricted in their env.

mercihabam:
Great, I will add a preference.

mercihabam:
The reported errors should be fixed by [this pull request](https://github.com/cypht-org/cypht/pull/1457).
@indridieinarsson, please let me know when you can if this has resolved the issue for you as well.

marclaporte:
> Great, I will add a preference.

https://github.com/cypht-org/cypht/pull/1459

PR #1454, opened 2025-02-21
feat(backend): call function authed on false and server error

amaninyumu1:
[Issues](https://github.com/cypht-org/cypht/issues/1450)


modules/smtp/modules.php
```diff
@@ -1576,7 +1576,7 @@ public function process() {
 
         foreach ($servers as $server_id => $config) {
             $mailbox = new Hm_Mailbox($server_id, $this->user_config, $this->session, $config);
-            if ($mailbox->connect()) {
+            if ($mailbox || $mailbox->connect()) {
```
josaphatim, line 1579: Can you leave this line as it was because the $mailbox cannot empty in this case ?

PR #1456, opened 2025-02-24
feat(backend): sieve manager packages to 1.0.9

Shadow243:
Update php sieve manager to latest just released.


Shadow243:
> I will merge after I saw an error that I must first correct before on the sieve_filters page

The first error has been corrected; I still have two other errors,  Edit and the checkbox are not working: 
Screenshot 2025-02-25 at 01 40 06

PR #1464, opened 2025-03-06
fix(backend): add database folder to docker build script

indridieinarsson:
## 🍰 Pullrequest
database scripts were recently moved to their own folder in the source tree. This breaks the docker build script, since the database scripts are no longer available inside the docker container.

This PR adds this folder to the Dockerfile.

### DIFF
--- a/tests/selenium/settings.py
+++ b/tests/selenium/settings.py
@@ -3,11 +3,12 @@
 from base import WebTest, USER, PASS
 from selenium.webdriver.common.by import By
 from runner import test_runner
-from selenium.webdriver.support.ui import Select
-
+from selenium.webdriver.support.ui import Select, WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.common.exceptions import NoSuchElementException
+from selenium.common.exceptions import StaleElementReferenceException
 
 class SettingsHelpers(WebTest):
-
     def is_unchecked(self, name):
         selected = self.by_name(name).is_selected()
         print(f"The selected status of checkbox '{name}' is:", selected)
@@ -24,10 +25,17 @@ class SettingsHelpers(WebTest):
 
     def save_settings(self):
         self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
+        alert_message = self.by_class('sys_messages')
         self.by_name('save_settings').click()
         self.wait_with_folder_list()
         self.safari_workaround()
-        assert self.by_class('sys_messages').text == 'Settings updated'
+        WebDriverWait(self.driver, 20).until(EC.staleness_of(alert_message))
+        ignored_exceptions=(NoSuchElementException,StaleElementReferenceException,)
+        _ = WebDriverWait(self.driver, 10, ignored_exceptions=ignored_exceptions).until(
+            EC.presence_of_element_located((By.CLASS_NAME, 'sys_messages'))
+        )
+        alert_message = self.by_class('sys_messages')
+        assert alert_message.text.strip() == 'Settings updated'
 
     def settings_section(self, section):
         if not self.by_class('settings').is_displayed():
@@ -79,7 +87,6 @@ class SettingsHelpers(WebTest):
 
 
 class SettingsTests(SettingsHelpers):
-
     def __init__(self):
         WebTest.__init__(self)
         self.login(USER, PASS)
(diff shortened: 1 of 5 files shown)


indridieinarsson:
@Shadow243   : there is a test failing, but it's unlikely caused by this PR. This PR only touches the docker container build, and I don't think the unit tests are running inside the container.

Shadow243:
> @Shadow243 : there is a test failing, but it's unlikely caused by this PR. This PR only touches the docker container build, and I don't think the unit tests are running inside the container.

I'm checking, I'll give you feedback in a moment

indridieinarsson:
@Shadow243 : I had some similar issues with stale elements a few months back. Refreshing the message list caused the site to reload, so the old element got stale, while a new element in the DOM got created. I think this is a race condition, and depends on how fast the test code gets to the line where the element is checked.

I tried solving this here : https://github.com/indridieinarsson/cypht/blob/1407bbe3b47b4a24331eb154976611cfd2b3a9bc/tests/selenium/folder_list.py#L18

Maybe we can use something similar here?

tests/selenium/settings.py
```diff
@@ -24,10 +25,20 @@ def close_section(self, section):
...
         self.safari_workaround()
-        assert self.by_class('sys_messages').text == 'Settings updated'
+        # .click() triggers site reload, so we explicitly wait for element to become stale
+        WebDriverWait(self.driver, 20).until(EC.staleness_of(alert_message))
+        ignored_exceptions=(NoSuchElementException,StaleElementReferenceException,)
+        # And once it is stale, we can now wait for it to become available again.
+        _ = WebDriverWait(self.driver, 10, ignored_exceptions=ignored_exceptions).until(
+            EC.presence_of_element_located((By.CLASS_NAME, 'sys_messages'))
+        )
+        alert_message = self.by_class('sys_messages')
```
Shadow243, line 39: This is already defined at https://github.com/cypht-org/cypht/pull/1464/files#diff-015513616f5a3e387c6d1e06e82a29d9be33debc8de739fcf998719ab2d62280R28 we can reuse it

indridieinarsson: Hmm, I think we need to renew the element. The reference from line 28 becomes stale once the page reloads (as a consequence of the .click() command).
The logic here is to 
- get the element from the page before reload. 
- Click the button, Page reloads and we wait for it to become stale as the old page ceases to exist.
- Wait for an element with the same name to be created in the DOM after reload

I'm actually not an expert on these things, but my understanding is that we need to get a reference to the `sys_messages` class again after reload.

Shadow243: Oh you are right.

Shadow243:
@indridieinarsson I ran the tests again to reassure myself but they failed again.

indridieinarsson:
> @indridieinarsson I ran the tests again to reassure myself but they failed again.

I'll take a look. I think that a few of these test failures (ones involving stale elements) are race conditions, similar to the one we were discussing above, so occurring randomly, based on who is faster, the page/server updating the page, or the test code fetching the element.

indridieinarsson:
@Shadow243  : the test that failed on the last run was a different one than the first failure (I think this is a race condition, and some of the tests just fail randomly due to that).
Anyways, I added some explicit wait for the `sys_messages` class in the test that failed most recently, so I hope that it's fixed now.

Shadow243:
@indridieinarsson Can you squash these three latest commits into one to keep the commit history cleaner ?

indridieinarsson:
> @indridieinarsson Can you squash these three latest commits into one to keep the commit history cleaner ?

done :)

tests/selenium/settings.py
```diff
@@ -24,10 +25,20 @@ def close_section(self, section):
 
     def save_settings(self):
         self.driver.execute_script("window.scrollTo(0, document.body.scrollHeight);")
+        alert_message = self.by_class('sys_messages')
         self.by_name('save_settings').click()
         self.wait_with_folder_list()
         self.safari_workaround()
-        assert self.by_class('sys_messages').text == 'Settings updated'
+        # .click() triggers site reload, so we explicitly wait for element to become stale
```
Shadow243, line 32: Everything looks good, I’m going to merge this. Just one last thing: can you remove these comments and https://github.com/cypht-org/cypht/pull/1464/files#diff-015513616f5a3e387c6d1e06e82a29d9be33debc8de739fcf998719ab2d62280R35 ? Thanks!

PR #1465, opened 2025-03-07
fix(backend): manually reconnect cache services after deserialize in message worker

indridieinarsson:
## 🍰 Pullrequest
PR #1452 serializes/unserializes the cache class when handing it to the worker process.
The Redis cache class (and presumably also the memcached class) don't survive this, as the connection to the service is not preserved when serializing. Thus the fetching of messages fails, with a `Redis server went away` exception message. In some cases, the scripts seems to sort-of work and fetch messages, but it takes longer, since the cache is probably not used.

This PR tries to fix that by adding a public manual `reconnect' function to the hm_cache class, which reconnects the redis or memcached backends.

### DIFF
--- a/lib/cache.php
+++ b/lib/cache.php
@@ -266,6 +266,13 @@ class Hm_Redis {
     }
 
     /**
+     * @return boolean
+     */
+    public function reconnect() {
+        return $this->connect();
+    }
+
+    /**
      * @return void
      */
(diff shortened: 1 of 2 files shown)


mercihabam:
This works fine, thank you @indridieinarsson for the fix.

PR #1467, opened 2025-03-10
fix(backend): fix error when adding new mail server

indridieinarsson:
## 🍰 Pullrequest
When adding a new mailserver, sometimes there is an error when adding the smtp server (adding only imap server sidesteps the error message). This is probably a regression from pr #1425. 

### To Review
- Review that adding smtp server still works for ews or jmap servers - the code from #1425 that added a call to `connect_to_smtp_server` addressed an issue with those kinds of servers. The change in this PR  shouldn't affect that code, but I can't really test it reliably.


indridieinarsson:
Fixed in the meantime by commit a152ab4

PR #1470, opened 2025-03-11
fix(backend): improve error handling is message-list sub-process

indridieinarsson:
## 🍰 Pullrequest
PR #1452 introduced a separate process for concurrently loading messages. In my use case, where cypht runs standalone in a docker container (compiled using the official cypht Dockerfile), most error messages and exceptions leave no traces in the logs, even if using the development versions of php.ini (also supplied by the base docker image). This makes it very hard to debug if things go wrong in message_list.php

This PR tries to remedy this by wrapping the entire contents of message_list.php in a try/catch clause, and then piping the resulting error messages to the parent process via stderr. This way, errors in the subprocess leave understandable log entries.

According to my (admittedly quite limited) research, there are negligible performance implications of wrapping the code in a try block, as long as no exceptions are thrown (performance hit of try/catch is supposed to be only due to the cost of handling the exception).


modules/imap/workers/messages_list.php
```diff
@@ -3,75 +3,81 @@
...
+            'total' => $result['total'],
+            'dataSource' => $result['dataSource'],
+            'folder' => $result['folder']
+        ]);
+    } else {
+        echo json_encode(['error' => 'Failed to establish connection to the mailbox.']);
+        exit;
+    }
+} catch (\Exception $e) {
+    Hm_Functions::error_log(sprintf(' error processing mailbox' .$e->getMessage() ));
```
kroky, line 80: Isn't it better to write to stderr any error messages and let the calling scripts decide how to log them? Otherwise, we have logging patterns on too many locations and it is harder to maintain.

modules/imap/functions.php
```diff
@@ -1673,9 +1673,15 @@ function getCombinedMessagesLists($sources, $context, $search) {
...
 
-                    $process->on('exit', function ($exitCode, $signal) use ($reject) {
-                        $reject(new \Exception($signal . ' Error code: ' . $exitCode));
+                    $process->stderr->on('data', function ($errorOutput) {
+                        error_log('Worker error output: ' . $errorOutput);
                     });
+
+                    $process->on('exit', function ($exitCode, $signal) use ($reject) {
+                        if ($exitCode != 0) {
+                            error_log(sprintf(' worker exited with abornaml exit code' . $exitCode ));
```
kroky, line 1682: Hm_Functions::error_log maybe?

modules/imap/workers/messages_list.php
kroky: Btw, even if you don't add a big try/catch block, exceptions thrown in the related code should end up in the stderr stream of the process and thus can be catched in the calling code.

modules/imap/functions.php
indridieinarsson: yes.

PR #1471, opened 2025-03-13
fix(frontend): only check sieve block lists if sieve module is enabled

indridieinarsson:
## 🍰 Pullrequest
Javascript code is calling `ajax_list_block_sieve` on page load (on combined_inbox at least), even if I don't have sievefilters module enabled. This causes an error in the backend ('not callable'), since the sieve stuff isn't available. This in turn gives rise to an error message on the web gui, each time the site is reloaded.

This PR makes sure this isn't called unless the module is enabled.


indridieinarsson:
PR #1435 resolved the same issue.

PR #1483, opened 2025-03-24
fix(backend): fix release asset files to include dot files

josaphatim:


.github/workflows/release.yml
```diff
@@ -26,14 +26,17 @@ jobs:
       - name: Install Composer dependencies
         run: composer install --no-dev --optimize-autoloader
 
+      - name: Generate configuration files
+        run: php scripts/config_gen.php
+
       - name: Create tarball
-        run: tar -czvf cypht.tar.gz ./*
+        run: cd .. && tar --exclude .github --exclude .git --exclude .gitignore --exclude .travis --exclude .travis.yml --exclude .coveralls.yml --exclude .env.example -pczf cypht.tar.gz cypht
```
kroky, line 33: I'd expect to find the project files when I untar/open the archive and not a directory cypht containing the project files. Can you tar with a star and probably exclusively add the dot files you want?

josaphatim: Thanks for this. Now fixed.

PR #1484, opened 2025-03-25
fix(other): php warning: undefined array key 'x-profile-id'

christer77:
PHP Warning:  Undefined array key "X-Profile-ID" in `...\cypht\modules\smtp\functions.php on line 81`

### DIFF
--- a/modules/smtp/functions.php
+++ b/modules/smtp/functions.php
@@ -70,7 +70,7 @@ function send_scheduled_message($handler, $imapMailbox, $folder, $msg_id, $send_
     $msg_headers = $imapMailbox->get_message_headers($folder, $msg_id);    
     $mailbox_details = $imapMailbox->get_config();  
     try {
-        if (empty($msg_headers['X-Schedule'])) {
+        if (empty($msg_headers['X-Schedule']) || empty($msg_headers['X-Profile-ID'])) {
             return false;
         }
 


modules/smtp/functions.php
```diff
@@ -75,6 +75,9 @@ function send_scheduled_message($handler, $imapMailbox, $folder, $msg_id, $send_
         }
 
         if (new DateTime($msg_headers['X-Schedule']) <= new DateTime() || $send_now) {
+            if (! isset($msg_headers['X-Profile-ID'])) {
```
Shadow243, line 78: Put the condition at the same as X-Schedule at line 73 and use empty instead of isset

Shadow243 (changes requested):
A small modification is necessary.

modules/smtp/functions.php
```diff
@@ -70,11 +70,18 @@ function send_scheduled_message($handler, $imapMailbox, $folder, $msg_id, $send_
     $msg_headers = $imapMailbox->get_message_headers($folder, $msg_id);    
     $mailbox_details = $imapMailbox->get_config();  
     try {
-        if (empty($msg_headers['X-Schedule'])) {
+        if (! isset($msg_headers['X-Schedule'])) {
```
Shadow243, line 73: @christer77 `empty` has isset and other condition. keep empty not isset

PR #1486, opened 2025-04-03
fix(backend): update sieve manager to latest release

Shadow243:
Update sieve manager to the latest released version


marclaporte:
Also done for branch 1.4.x: https://github.com/cypht-org/cypht/pull/1487

PR #1490, opened 2025-04-08
fix(frontend/backend): prevent Sieve connection error when disabled in settings

Shadow243:


modules/imap/hm-ews.php
```diff
@@ -151,7 +151,7 @@ public function get_folder_status($folder) {
             } else {
                 $result = $this->api->getFolderByDisplayName($folder, Enumeration\DistinguishedFolderIdNameType::MESSAGE_ROOT);
                 if (! $result) {
-                    throw new Exception('Folder not found.');
+                    throw new Exception('Folder '. $folder .' not found.');
```
Shadow243, line 154: I'm going to leave a message for @kroky  to confirm my theory about this.

language/az.php
```diff
@@ -411,7 +411,7 @@
     'Added SMTP server!' => 'SMTP server əlavə edilmişdir!',
     'Added repository' => 'Anbar əlavə edilmişdir',
     'Added server!' => 'Server əlavə edilmişdir',
-    'Are you sure?' => 'Siz əminsiniz?',
+    'Are you sure you wanto to delete this server? ' => 'Siz əminsiniz?',
```
marclaporte, line 414: typo

modules/imap/hm-ews.php
kroky: Looks good

modules/sievefilters/modules.php
```diff
@@ -1373,31 +1374,32 @@ public function process() {
...
-        try {
-            $client = $factory->init($this->user_config, $imap_account);
-
-            $blocked_senders = [];
-            $current_script = $client->getScript('blocked_senders');
-            if ($current_script != '') {
-                $blocked_list = prepare_sieve_script ($current_script);
-                foreach ($blocked_list as $blocked_sender) {
-                    $blocked_senders[] = $blocked_sender;
+        if($this->module_is_supported('sievefilters')  && $this->user_config->get('enable_sieve_filter_setting', DEFAULT_ENABLE_SIEVE_FILTER)){
```
kroky, line 1377: How would the sievefilters module be not supported when you are in that module? If other modules call this one in their setup files, we should probably move those setup lines to the sievefilters/setup file which is only loaded when the module is loaded.

Shadow243: @kroky We added the ability to disable a module via the interface(Related PR: https://github.com/cypht-org/cypht/pull/675), even if the module is defined in the `.env` file. Since this update, it's now possible to disable Sieve even if it remains declared in `.env`.

Screenshot 2025-05-01 at 19 55 42

I checked: all calls are coming from this module.

Shadow243:
@kroky I've applied the changes requested at [https://github.com/cypht-org/cypht/commit/18aea3e82694b9a81fc4e02008f0885d1ff77329#r157342289](https://github.com/cypht-org/cypht/commit/18aea3e82694b9a81fc4e02008f0885d1ff77329#r157342289)

modules/sievefilters/modules.php
kroky: What does it mean that sieve filter is not enabled via UI? Isn't it the same as the .env - user selects it and the module is not loaded at all or does it mean something else?

Shadow243:
> Yes, that's quite better, thanks! I am doing some more considerable refactoring in master due to the performance degradation we face and once I merge this, we can check the conflicts here and then merge this one as well.


@kroky I’ve just resolved the conflict. Everything should be in order now. feel free to review.

marclaporte:
@Shadow243 A test is failing

Shadow243:
> @Shadow243 A test is failing

@marclaporte Yes, indeed. The Selenium test is buggy but is being fixed at https://github.com/cypht-org/cypht/pull/1534.

modules/sievefilters/modules.php
Shadow243: The idea was to provide an extra level of control: even if Sieve is enabled in the .env, we should still be able to disable it through the UI if needed. This way, a user can deactivate Sieve at the settings level without touching the .env file. It acts as a second layer of deactivation. so even if it’s technically active via .env, it becomes inaccessible if turned off in the UI.

kroky: Yes, I understand the UI part. I don't understand the technical implementation. If sieve is not enabled via .env or via settings, then the relevant sieve modules should not be loaded. We shouldn't have many places where we check 'if enabled do this, otherwise do that". Module system is flexible enough to provide checks which modules to load and if modules are loaded, they are executed.

modules/api_login/modules.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+require_once APP_PATH.'modules/core/handler_modules.php';
```
kroky: Please get rid of these includes. Module loading system shouldn't be part of the individual modules' code. Furthermore, core module should always be loaded.

modules/imap_folders/modules.php
```diff
@@ -319,6 +319,17 @@ public function process() {
         if (array_key_exists('imap_server_id', $this->request->get)) {
             $this->out('folder_server', $this->request->get['imap_server_id']);
             $this->out('page', $this->request->get['page']);
+            $this->out('trigger_default_submit', false);
+        }else {
+            // as requested, if there no selected server we should select on by default based on profiles
+            require_once APP_PATH.'modules/profiles/hm-profiles.php';
```
kroky: Again, we don't require modules and functions like this. If profiles module is enabled, use it directly. It already has methods to init the profiles, etc. - which should be included in this module's setup.php, so the profiles are already available (and only available) when the profiles module is turned on.

modules/imap_folders/site.js
```diff
@@ -1,51 +1,79 @@
...
-    $('a', folder_location).not('.'+link_class).not('.close').on("click", function() { set_folders_page_value($(this).data('id'), container, target, id_dest); return false; });
-    $('.close', folder_location).on("click", function() {
+
+    // 1. Get the <a> element
+    const link = document.querySelector('.'+link_class);
+    // 2. Save the <i> element
+    const original_icon = link.querySelector('i');
+    const original_icon_clone = original_icon.cloneNode(true);
+    // 3. Create the spinner
+    const spinner = document.createElement('div');
```
kroky: Why do you have all these unrelated updates when the PR is for sieve connection? It is better to have PRs that are self-contained - easier to review, easier to trace in the history.

modules/keyboard_shortcuts/modules.php
```diff
@@ -61,6 +61,8 @@ public function process() {
             $this->session->set('user_data', $user_data);
             $this->session->record_unsaved('Shortcut updated');
             Hm_Msgs::add('Shortcut updated');
+            $this->save_hm_msgs();
+            Hm_Dispatch::page_redirect('?page=shortcuts');
```
kroky: This also seems totally unrelated. Please commit bug fixes to separate PRs.

modules/sievefilters/modules.php
```diff
@@ -1373,39 +1373,64 @@ public function process() {
...
-        } catch (Exception $e) {
-            Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
-            return;
         }
     }
 }
 
+class Hm_Handler_check_sieve_configuration extends Hm_Handler_Module {
+    public function process() {
+        if($this->module_is_supported('sievefilters')  && $this->user_config->get('enable_sieve_filter_setting', DEFAULT_ENABLE_SIEVE_FILTER)){
```
kroky: You can safely skip 'module_is_supported' here. You are inside sievefilters modules, if this is loaded, then module is supported, yes. Keep the check for config.

Regarding the check from config - I believe it is better to move this elsewhere. If the setting is disabled by the user config, then many of the modules defined here in this file need to be disabled (i.e. not executed). Unfortunately, we cannot do this for all the modules, as the setting itself is a module and triggering that needs to work even if it is unchecked (disabled by the UI). Thus, I thought of doing this in setup.php but we don't have the user config loaded there as there are modules to load the config. Hmm, then we might have no other choice but have all these if user enabled sieve, then execute the module. Maybe you can at least move this to a function and call it immediately at the start of each module process method which checks if the user has enabled it. If not, it returns immediately. This way, we don't need to wrap the whole module code in an if, thus indenting one way further.

Shadow243:
Hi Victor,

I’ve split the changes as you suggested — the updated PR now focuses solely on the Sieve connection fix, keeping it clean and self-contained.

Original PR: #1490
Updated PR: 

1. https://github.com/cypht-org/cypht/pull/1536
2. https://github.com/cypht-org/cypht/pull/1537
3. https://github.com/cypht-org/cypht/pull/1538
4. https://github.com/cypht-org/cypht/pull/1539

Let me know if you have any further suggestions!

modules/sievefilters/modules.php
```diff
@@ -1373,39 +1408,69 @@ public function process() {
...
-        list($success, $form) = $this->process_form(array('imap_server_id'));
-        if (!$success) {
-            return;
-        }
+        if ($this->should_skip_execution('enable_sieve_filter_setting', DEFAULT_ENABLE_SIEVE_FILTER)) return;
 
-        Hm_IMAP_List::init($this->user_config, $this->session);
-        $imap_account = Hm_IMAP_List::get($form['imap_server_id'], true);
+        // Sieve may be enabled via .env but disabled in the settings, so we need to verify
+        if($this->user_config->get('enable_sieve_filter_setting', DEFAULT_ENABLE_SIEVE_FILTER)) {
```
kroky: No need for this if statement - you already check with should_skip_execution.

modules/sievefilters/modules.php
kroky: Looks better, thanks. Still not ideal but we can move forward with this approach for now and refactor module lib later to allow more elegant exclusion of modules based off user config.

modules/sievefilters/modules.php
Shadow243: Updated

PR #1491, opened 2025-04-13
feat(other): add HTML email sanitizer to remove external background images

Baraka24:
### Issues

Add sanitize_email_html function that:
- Parses HTML content with DOMDocument
- Removes potentially dangerous background-image URLs (http/https)
- Preserves all other HTML structure and styling
- Handles malformed HTML gracefully with libxml error suppression

### Todo

- [X] Test Cypht With [Email Privacy Tester](https://www.emailprivacytester.com/)

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -46,6 +46,19 @@ function format_msg_html($str, $images=false) {
 }}
 
 /**
+ * Sanitize HTML for email
+ * @subpackage core/functions
+ * @param string $html content to sanitize
+ * @return string
+ */
+if (!hm_exists('sanitize_email_html')) {
+function sanitize_email_html($html) {
+    $html = preg_replace('/<([^>]+)style\s*=\s*["\'][^"\']*background-image\s*:\s*url\((["\']?)https?:\/\/.*?\2\)[^"\']*["\']/i', '<$1', $html);
+
+    return $html;
+}}
+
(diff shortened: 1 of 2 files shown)


marclaporte:
Super!

Does this need a backport to 1.4.x?

modules/core/message_functions.php
```diff
@@ -45,6 +45,37 @@ function format_msg_html($str, $images=false) {
...
+ * @return string
+ */
+if (!hm_exists('sanitize_email_html')) {
+function sanitize_email_html($html) {
+    libxml_use_internal_errors(true);
+
+    $doc = new DOMDocument();
+    $doc->loadHTML('<?xml encoding="utf-8" ?>' . $html);
+
+    $xpath = new DOMXPath($doc);
```
kroky, line 61: The classes and functions used here need libxml but we don't have it in composer requirements. Do we really need to make libxml cypht requirement just to sanitize the background images? As you can see, external image sanitization is done via regular expressions. I think it is better to continue that approach with background images as well.

Baraka24: Okay well, I've applied changes.

Baraka24:
> Super!
> 
> Does this need a backport to 1.4.x?

No need to backport but, we'll also test Cypht 1.4.x accordingly. Thanks @marclaporte

PR #1494, opened 2025-04-18
fix(other): reviewing delete action on folders

christer77:
Deleting a folder seems to be successful,
But when it comes to a folder that contains subfolders, nothing was deleted.

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -135,7 +135,7 @@ function format_imap_folder_section($folders, $id, $output_mod, $with_input = fa
 
     foreach ($folders as $folder_name => $folder) {
         $folder_name = bin2hex($folder_name);
-        $results .= '<li class="imap_'.$id.'_'.$output_mod->html_safe($folder_name).'">';
+        $results .= '<li class="imap_'.$id.'_'.$output_mod->html_safe($folder_name).'" data-number-children="'.$output_mod->html_safe($folder['number_of_children']).'">';
         if ($folder['children']) {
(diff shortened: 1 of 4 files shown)


marclaporte:
Good catch!

I suggest 
1. A confirmation message: Ex.: this folder contains 4 sub-folder. Are you sure you want to delete them all?
2. Add a line in https://github.com/cypht-org/cypht/wiki/Test-Plan-Summary-Cypht-Webmail for this case.

christer77:
> Good catch!
> 
> I suggest:
> 
> 1. A confirmation message: Ex.: this folder contains 4 sub-folder. Are you sure you want to delete them all?
> 2. Add a line in https://github.com/cypht-org/cypht/wiki/Test-Plan-Summary-Cypht-Webmail for this case.

Good suggestion. I'm getting started

modules/core/hm-mailbox.php
```diff
@@ -164,14 +165,36 @@ public function delete_folder($folder) {
...
+            } else {
+                return false;
+            }
         } else {
             $del_folder = decode_folder_str($folder);
             return $this->connection->delete_folder($del_folder);
         }
     }
 
+    public function get_recursive_subfolders($parentFolder) {
```
kroky, line 189: hm-mailbox is a bridge towards different connections (protocols) for accessing mail. What you do with the changes here is imap-specific, so please move it to hm-imap file where it belongs.

christer77:
> Good catch!
> 
> I suggest:
> 
> 1. A confirmation message: Ex.: this folder contains 4 sub-folder. Are you sure you want to delete them all?
> 2. Add a line in https://github.com/cypht-org/cypht/wiki/Test-Plan-Summary-Cypht-Webmail for this case.

Case now managed

PR #1504, opened 2025-05-02
fix(frontend/backend): block sender and Screen email (dislike email) not working

christer77:
1. For email server that supports Sieve, opening a message, in the headers section try to click "Block Sender".
A php error is returned in the server logs.

2. When you click on dislike email, there are 2 behaviors that are random:
- sometimes the page reloads again
- If the page does not reload, there is a PHP error is returned.

PHP error "Fatal error: Uncaught TypeError: array_key_exists(): Argument #2 ($array) must be of type array, stdClass given in"

### DIFF
--- a/modules/sievefilters/functions.php
+++ b/modules/sievefilters/functions.php
@@ -285,13 +285,13 @@ if (!hm_exists('get_sieve_client_factory')) {
 }
 
 if (!hm_exists('prepare_sieve_script ')) {
-    function prepare_sieve_script ($script, $index = 1, $action = "decode")
+    function prepare_sieve_script ($script, $index = 1, $action = "decode", $object_to_array = true)
     {
         $blocked_list = [];
         if ($script != '') {
             $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[$index]);
             if ($action == "decode") {
-                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)));
+                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)), $object_to_array);
             } else {
                 $blocked_list = json_encode(base64_decode($base64_obj));
             }
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -77,7 +77,7 @@ class Hm_Output_imap_custom_controls extends Hm_Output_Module {
             }
             if ($this->get('folder')) {
                 if ($this->get('screen_emails')) {
-                    $custom .= '<a title="Disike" href="#" class="screen-email-unlike"><i class="bi bi-hand-thumbs-down-fill"></i></a><a title="Like" href="#" class="screen-email-like"></i><i class="bi bi-hand-thumbs-up-fill"></i></a>';
+                    $custom .= '<span title="Disike" class="screen-email-unlike cursor-pointer"><i class="bi bi-hand-thumbs-down-fill"></i></span><span title="Like" class="screen-email-like cursor-pointer"></i><i class="bi bi-hand-thumbs-up-fill"></i></span>';
                     if ($this->get('move_messages_in_screen_email')) {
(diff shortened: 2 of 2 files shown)


modules/sievefilters/functions.php
```diff
@@ -285,13 +285,13 @@ function get_sieve_client_factory($site_config)
 }
 
 if (!hm_exists('prepare_sieve_script ')) {
-    function prepare_sieve_script ($script, $index = 1, $action = "decode")
+    function prepare_sieve_script ($script, $index = 1, $action = "decode", $object_to_array = true)
```
kroky, line 288: It is better to standardize this - either always return objects or arrays when decoding. Also checking calling code that it works fine with whatever type you choose to use here.

marclaporte:
@christer77 Please resolve conflicts.

christer77:
Hello @kroky , 

there is something that happened here that I didn't understand. 
The report indicates that I closed it while it wasn't me. 

Do you understand?

marclaporte:
Hmmm, weird. It now says "Files changed" is 0.

modules/imap/handler_modules.php
```diff
@@ -260,6 +261,23 @@ public function process() {
...
                     $form['imap_move_to'] = $parts[0] ."_". $parts[1] ."_".bin2hex($screen_folder);
+                    $imap_move_ids = explode(",", $form['imap_move_ids']);
+
+                    foreach ($imap_move_ids as $imap_msg_id) {
+                        $array_imap_msg_id = explode("_", $imap_msg_id);
+                        if (isset($array_imap_msg_id[2])) {
+                            
+                            $msg_header = $mailbox->get_message_headers(hex2bin($array_imap_msg_id[3]), $array_imap_msg_id[2]);
+                            
+                            $test_pattern = "/(?:[a-z0-9!#$%&'*+=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+=?^_`{|}~-]+)*|\"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*\")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/";
```
kroky: Where do you get this from? Can we move it to a constant or make it a global config that we can easily update in the future? Isn't it better to use mail-mime-parser or another lib to extract the address part of the from header and use that instead of relying on a complex regular expressions that might miss something?

christer77: Noted

kroky:
@christer77 I am not sure about closing - shouldn't happen automatically. Overall, this looks good - just one comment added.

modules/core/functions.php
```diff
@@ -777,3 +777,16 @@ function getSettingsSectionOutput($section, $sectionLabel, $sectionIcon, $settin
...
     return $res;
 }
+
+/**
+ * @subpackage imap/functions
+ */
+if (!hm_exists('extract_email_adress')) {
+    function extract_email_adress($str)
+    {
+        if (preg_match("/^(.*?) <(.*?)>$/", $str, $matches)) {
```
kroky: I believe a simple email address without the name and enclosing tags is also a valid address, so we should consider using that too. Didn't you find a suitable function already present in the libs we are using?

christer77: If you are right, I could find process_address_fld

kroky: Yes, that or addr_parse

PR #1516, opened 2025-05-13
feat(frontend/backend): adding calendar icon to emails with a calendar invitation

christer77:
In message list, emails with a calendar invitation should have an icon.

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -979,10 +979,11 @@ if (!class_exists('Hm_IMAP')) {
             if ($this->is_supported( 'X-GM-EXT-1' )) {
                 $command .= 'X-GM-MSGID X-GM-THRID X-GM-LABELS ';
             }
-            $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID IN-REPLY-TO X-SNOOZED X-SCHEDULE X-PROFILE-ID X-DELIVERY)]";
+            $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID IN-REPLY-TO X-SNOOZED X-SCHEDULE X-PROFILE-ID X-DELIVERY X-MS-EXCHANGE-CALENDAR-SERIES-INSTANCE-ID SENDER)]";
             if ($include_content_body) {
                 $command .= " BODY.PEEK[TEXT]<0.500>";
             }
+            $command .= " BODY.PEEK[1.MIME] BODY.PEEK[2.MIME]";
             $command .= ")\r\n";
             $cache_command = $command.(string)$raw;
             $cache = $this->check_cache($cache_command);
@@ -993,8 +994,10 @@ if (!class_exists('Hm_IMAP')) {
             $res = $this->get_response(false, true);
             $status = $this->check_response($res, true);
             $tags = array('X-GM-MSGID' => 'google_msg_id', 'X-GM-THRID' => 'google_thread_id', 'X-GM-LABELS' => 'google_labels', 'UID' => 'uid', 'FLAGS' => 'flags', 'RFC822.SIZE' => 'size', 'INTERNALDATE' => 'internal_date');
-            $junk = array('X-AUTO-BCC', 'MESSAGE-ID', 'IN-REPLY-TO', 'REFERENCES', 'X-SNOOZED', 'X-SCHEDULE', 'X-PROFILE-ID', 'X-DELIVERY', 'LIST-ARCHIVE', 'SUBJECT', 'FROM', 'CONTENT-TYPE', 'TO', '(', ')', ']', 'X-PRIORITY', 'DATE');
-            $flds = array('x-auto-bcc' => 'x_auto_bcc', 'message-id' => 'message_id', 'in-reply-to' => 'in_reply_to', 'references' => 'references', 'x-snoozed' => 'x_snoozed', 'x-schedule' => 'x_schedule', 'x-profile-id' => 'x_profile_id', 'x-delivery' => 'x_delivery', 'list-archive' => 'list_archive', 'date' => 'date', 'from' => 'from', 'to' => 'to', 'subject' => 'subject', 'content-type' => 'content_type', 'x-priority' => 'x_priority', 'body' => 'content_body');
+            $junk = array('X-AUTO-BCC', 'MESSAGE-ID', 'IN-REPLY-TO', 'REFERENCES', 'X-SNOOZED', 'X-SCHEDULE', 'X-PROFILE-ID', 'X-DELIVERY', 'LIST-ARCHIVE', 'SUBJECT', 'FROM', 'CONTENT-TYPE', 'TO', '(', ')', ']', 'X-PRIORITY', 'DATE', 'X-MS-EXCHANGE-CALENDAR-SERIES-INSTANCE-ID', 'SENDER');
+            $flds = array('x-auto-bcc' => 'x_auto_bcc', 'message-id' => 'message_id', 'in-reply-to' => 'in_reply_to', 'references' => 'references', 'x-snoozed' => 'x_snoozed', 'x-schedule' => 'x_schedule', 'x-profile-id' => 'x_profile_id', 'x-delivery' => 'x_delivery', 'list-archive' => 'list_archive', 'date' => 'date', 'from' => 'from', 'to' => 'to', 'subject' => 'subject', 'content-type' => 'content_type', 'x-priority' => 'x_priority', 'body' => 'content_body', 'type_msg' => 'type_msg');
+            $flds['x-ms-exchange-calendar-series-instance-id'] = "x_ms_exchange_calendar_series_instance_id";
+            $flds['sender'] = "sender";
             $headers = array();
 
             foreach ($res as $n => $vals) {
@@ -1021,9 +1024,12 @@ if (!class_exists('Hm_IMAP')) {
                     $x_schedule = '';
                     $x_profile_id = '';
                     $x_delivery = '';
+                    $x_ms_exchange_calendar_series_instance_id = '';
+                    $sender = '';
                     $count = count($vals);
                     $header_founded = false;
                     $body_founded = false;
+                    $flds['type_msg'] = '';
                     for ($i=0;$i<$count;$i++) {
                         if ($vals[$i] == 'BODY[HEADER.FIELDS' && !$header_founded) {
                             $header_founded = true;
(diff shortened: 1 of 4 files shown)


modules/imap/hm-imap.php
```diff
@@ -974,7 +974,7 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
             if ($this->is_supported( 'X-GM-EXT-1' )) {
                 $command .= 'X-GM-MSGID X-GM-THRID X-GM-LABELS ';
             }
-            $command .= "BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID X-SNOOZED X-SCHEDULE X-PROFILE-ID X-DELIVERY)]";
+            $command .= "BODYSTRUCTURE BODY.PEEK[HEADER.FIELDS (SUBJECT X-AUTO-BCC FROM DATE CONTENT-TYPE X-PRIORITY TO LIST-ARCHIVE REFERENCES MESSAGE-ID X-SNOOZED X-SCHEDULE X-PROFILE-ID X-DELIVERY)]";
```
kroky, line 977: This will have a serious performance impact. We shouldn't retrieve the whole body structure of imap messages in list context as this is done to many items as ones. That's why the command uses peek - it only gets specific header fields for performance reasons.

modules/imap/hm-imap.php
```diff
@@ -1038,6 +1039,9 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
                                 }
                             }
                         }
+                        elseif ($vals[$i] == 'invite.ics' ) {
```
kroky, line 1042: Seems like a random hard-coded name to check. If you want to check for invitation, maybe use content type or other more-reliable method? What if the file is called invitation.ics? Maybe check for .ics extension?

modules/imap/hm-imap.php
christer77: Great, we need to look for another more optimal solution.

christer77:
Hello @kroky 

In this solution, I handled the cases where the event invitation was sent from Outlook or Gmail.

modules/imap/hm-imap.php
```diff
@@ -2635,5 +2644,11 @@ public function server_support_children_capability() {
 
             return false;
         }
+        private function is_invitaion_message($str) {
```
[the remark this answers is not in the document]
christer77: Noted

modules/imap/hm-imap.php
```diff
@@ -978,6 +978,7 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
             if ($include_content_body) {
                 $command .= " BODY.PEEK[TEXT]<0.500>";
             }
+            $command .= " BODY.PEEK[1] BODY.PEEK[2]";
```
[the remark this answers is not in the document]
kroky: Sender is a valid envelope header for all messages. We can't consider it a calendar if it has a sender. What's the problem of fetching each mime part header content type?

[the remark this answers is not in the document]
christer77: It must be said that I tested with MINE, 
the problem lies in the fact that, for an invitation sent via Gmail, we find the information regarding the event at 2.MINE, 
while with Outlook it is at 3.MINE.Now, 
it seems that Google does not have a 3.MINE, which means that any message without this part will not be displayed. 
That is why I preferred SENDER and X-MS-EXCHANGE-CALENDAR-SERIES-INSTANCE-ID.
SENDER for Gmail returns content like "calendar ... calendar@gmail.com" and X-MS-EXCHANGE-CALENDAR-SERIES-INSTANCE-ID contains information only if it is indeed an event.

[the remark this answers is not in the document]
christer77: Alright, let me see if there is a possibility. I will get back to you shortly.

modules/imap/hm-imap.php
```diff
@@ -1038,6 +1044,14 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
                                 }
                             }
                         }
+                        elseif ($vals[$i] == 'BODY[1.MIME' || $vals[$i] == 'BODY[2.MIME') {
+                            $content = '';
+                            $i++;
+                            $i++;
+                            if (strpos($vals[$i], 'text/calendar') !== false || strpos($vals[$i], '.ics') !== false) {
```
kroky: We've switched to mb_ versions of all these string functions for safety.

modules/imap/hm-imap.php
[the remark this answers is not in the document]
kroky: Yes, I think that would be enough for now, thanks!

modules/imap/hm-imap.php
christer77: Noted

modules/imap/hm-imap.php
```diff
@@ -1091,7 +1105,16 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
...
-                                         'x_snoozed'  => $x_snoozed, 'x_schedule' => $x_schedule, 'x_profile_id' => $x_profile_id, 'x_delivery' => $x_delivery);
+                                         'x_snoozed'  => $x_snoozed, 'x_schedule' => $x_schedule, 'x_profile_id' => $x_profile_id, 'x_delivery' => $x_delivery, 'x_ms_exchange_calendar_series_instance_id' => $x_ms_exchange_calendar_series_instance_id, 'sender' => $sender);
+                        $headers[$uid]['type_msg'] = $flds['type_msg'] != "type_msg" ? $flds['type_msg'] :  "";
+                        if ($headers[$uid]['x_ms_exchange_calendar_series_instance_id'] || $headers[$uid]['sender']) {
+                            if (strpos($headers[$uid]['sender'], 'Google Calendar <calendar-')) {
+                                $headers[$uid]['type_msg'] = "calendar";
+                            }
+                            if (! empty(trim($headers[$uid]['x_ms_exchange_calendar_series_instance_id']))) {
+                                $headers[$uid]['type_msg'] = "calendar";
```
kroky: One last thing - can we add this type_msg to EWS as well?

christer77: Okay

marclaporte:
@christer77 A test is failing

.github/workflows/release.yml
```diff
@@ -26,14 +26,17 @@ jobs:
       - name: Install Composer dependencies
         run: composer install --no-dev --optimize-autoloader
 
+      - name: Generate configuration files
```
marclaporte, line 29: @christer77 
Seems unrelated https://en.wikipedia.org/wiki/Atomic_commit#Atomic_commit_convention

marclaporte:
This enhancement is not a blocker to release 2.5.0 but it is also low-risk enough to sneak in 2.5.0

modules/imap/hm-imap.php
kroky: This is still not resolved.

christer77: Currently I am stuck due to issues with the ews authentication.

marclaporte: @christer77 Are you still blocked? If so, what is the task URL?

christer77: Actually, no, 

We're waiting for the ews connection issue to be resolved.

marclaporte: @christer77 You answered one of my question, but I asked two.

christer77: https://github.com/Garethp/php-ews/pull/275

marclaporte: @christer77 

On https://github.com/Garethp/php-ews/pull/275, we see a PR to unblock: https://github.com/cypht-org/cypht/pull/1635

christer77: Noted.

Thanks

christer77: > One last thing - can we add this type_msg to EWS as well?

@kroky,  this is a good catch really, we were trying to work on this with @Baraka24  and fund that **EWS implementation in Cypht does not contain a part to display messages with calendar items** and we tried several times to fix it even after testing on a simple script expecting the issue is from EWS itself and got it worked well. We need help from you (@kroky) to handle this so we apply appropriate changes.

kroky: Sorry, I didn't understand - what kind of help do you need? Do you need me to review another PR or help with a specific question you might have?

christer77: Hello @kroky  , sorry,

I opened here for a good explanation:
https://github.com/cypht-org/cypht/issues/1672

marclaporte:
FYI: https://github.com/cypht-org/cypht/pull/1635 has been merged.

christer77:
> FYI: #1635 has been merged.

Now we are waiting the merge of  https://github.com/cypht-org/cypht/pull/1699

marclaporte:
https://github.com/cypht-org/cypht/pull/1699 has now been merged

modules/imap/hm-imap.php
kroky: @christer77, I did some more research and it seems your original idea of BODYSTRUCTURE is better in terms of performance. Fetching mime headers with peek 1.mime, etc. actually makes the server parse the message while bodystructure is a lightweight variant where we can see attachments and content types of parts. Can you please convert it back to use bodystructure and also add a setting to make this optional - might be enabled by default but if it turns out to be slow, user can disable the feature. Sorry for the back and forth here...

christer77: Noted, thank you

modules/imap/hm-imap.php
```diff
@@ -1021,10 +1026,16 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
...
                     $x_profile_id = '';
                     $x_delivery = '';
+                    $x_ms_exchange_calendar_series_instance_id = '';
+                    $sender = '';
                     $count = count($vals);
                     $header_founded = false;
                     $body_founded = false;
+                    $flds['type_msg'] = '';
                     for ($i=0;$i<$count;$i++) {
+                        if ($i == 0 && (preg_grep('/^(calendar|text\/calendar)$/i', $vals) || preg_grep('/\.ics$/i', $vals))) {
```
kroky: I don't get it how the line beginning with * will return the bodystructure as well - where do you parse the bodystructure to see if a calendar attachment is available? Simple grep through the whole thing might have false-positives.
You should check specific headers as content-type and then the bodystructure response specifically parsing and searching for calendar type.

kroky: Also, can you extend the test/phpunit/imap_commands.php with a sample BODY.PEEK and BODYSTRUCTURE in the same command?

christer77: Noted

modules/imap/hm-imap.php
```diff
@@ -2621,5 +2651,75 @@ public function server_support_children_capability() {
...
+                }
+            }
+
+            return false;
+        }
+
+        /**
+         * Parse the IMAP response BODYSTRUCTURE
+         */
+        private function parseBodystructure($response, $startIndex) {
```
kroky: Can't you use the existing hm-imap-bodystructure class and functionality?

modules/imap/hm-imap.php
```diff
@@ -2621,5 +2651,75 @@ public function server_support_children_capability() {
             return false;
         }
 
+        /**
+         * Parse the IMAP BODYSTRUCTURE to detect calendar parts
+         */
+        private function containsCalendarPart($bodystructure) {
```
kroky: I think you can use the Hm_IMAP_Struct::recursive_search to achieve the same result.

modules/imap/hm-imap.php
christer77: Noted

PR #1524, opened 2025-05-25
fix(backend): error Sieve: Invalid config host

Shadow243:
Screenshot 2025-05-26 at 01 31 20

### DIFF
--- a/modules/sievefilters/site.js
+++ b/modules/sievefilters/site.js
@@ -1007,7 +1007,7 @@ function get_list_block_sieve() {
     sessionStorage.removeItem('list_blocked');
     let detail = Hm_Utils.parse_folder_path(hm_list_path());
     let list_blocked_senders = [];
-    if (getPageNameParam() == 'message_list') {
+    if (getPageNameParam() == 'message_list' && detail !== false) {
         Hm_Ajax.request(
             [
                 { name: 'hm_ajax_hook', value: 'ajax_list_block_sieve' },


modules/sievefilters/site.js
```diff
@@ -1007,7 +1007,7 @@ function get_list_block_sieve() {
     sessionStorage.removeItem('list_blocked');
     let detail = Hm_Utils.parse_folder_path(hm_list_path());
     let list_blocked_senders = [];
-    if (getPageNameParam() == 'message_list') {
+    if (getPageNameParam() == 'message_list' && detail !== false) {
```
kroky, line 1010: Why would detail be false? Isn't this a problem of calling get_list_block_sieve on an incorrect place where imap folder has not been chosen? In that case, maybe a user error occurred and a special error msg should be displayed instead of silently ignoring?

Shadow243: The issue is with this line: let detail = Hm_Utils.parse_folder_path(hm_list_path());.
It only works if we pass a parameter like: `list_path=imap_xxxxx_xxxx`, for example:
`http://cypht.test/?page=message_list&list_path=imap_xxxxx_xxxx`.
Otherwise, it returns false.

In this context, if I’m just on:
`http://cypht.test/?page=message_list` (like when I click on unread in the main menu), it returns false because no parameter is passed.

But if I go through the menu 
Screenshot 2025-05-26 at 14 57 56
, I get a valid result because the IMAP server is known.

PR #1527, opened 2025-05-28
fix(backend): crash when messages key is null in IMAP result

Shadow243:
Screenshot 2025-05-28 at 04 57 12


kroky:
I am doing a major refactoring of this code in another PR and this will be gone, so no need to merge now.

PR #1528, opened 2025-05-28
feat(frontend/backend): add snooze functionality toggle

DKing2222:
Related [task ](https://avan.tech/item128981#threadId89840)
## 🍰 Pullrequest
This PR adds a toggle to enable or disable the Snooze functionality in the settings page. By default, the Snooze functionality is disabled.

### Changes
- Added `enable_snooze` setting in `modules/imap/output_modules.php`.
- Updated `modules/imap/setup.php` to register the handler and output module.
- Modified `modules/core/output_modules.php` to conditionally display the Snoozed folder.
- Default value for `enable_snooze` set to `false` in `config/app.php`.


kroky:
Besides the UI we also need to disable checking for snoozed messages if the setting is off - this is what is really making the current implementation unscalable and very slow - it checks for snoozed messages on almost every request for every defined imap server.

kroky:
@DKing2222 are you actively working here? I want to polish this one and merge soon as an effort to stabilize master branch and make it run smoothly, so we can release 2.5 and tag Tiki to use it. I can wait a couple more days but if you are busy elsewhere, I can polish this and merge.

DKing2222:
@kroky . Yes, I’m actively working on it. I’ll push my updates very soon.

modules/imap/output_modules.php
```diff
@@ -866,8 +866,13 @@ class Hm_Output_filter_imap_folders extends Hm_Output_Module {
      */
     protected function output() {
         $res = '';
+        $settings = $this->get('user_settings', array());
+        $enable_snooze = array_key_exists('enable_snooze', $settings) && $settings['enable_snooze'];
         if ($this->get('imap_folders')) {
             foreach ($this->get('imap_folders', array()) as $id => $folder) {
+                if (!$enable_snooze && mb_strtolower($folder) === 'snoozed') {
```
kroky: I don't think we should filter out snoozed folder by name here. It might be a normal folder name for some IMAP servers and is not directly connected with enabling or disabling the snoozing functionality. It is enough to remove the buttons to snooze/unsnooze and checking for snoozing messages. We should display any folder defined on the remote IMAP server.

kroky:
This is still far from ready. Snooze and unsnooze actions are still callable. Furthermore, there is still the JS code that calls unsnooze action initially on navigation and then every 60 seconds. All these should be disabled unless the setting is enabled.

modules/imap/output_modules.php
DKing2222: Thanks @kroky . 
You're right, I’ve removed the filter

DKing2222:
@kroky Everything should be resolved now.

modules/imap/site.js
```diff
@@ -1202,10 +1202,12 @@ var imap_setup_snooze = function() {
 }
 
 var imap_unsnooze_messages = function() {
-    Hm_Ajax.request(
-        [{'name': 'hm_ajax_hook', 'value': 'ajax_imap_unsnooze'}],
-        function() {},
-    );
+    if(window.hm_default_setting_enable_snooze) {
```
kroky: Disable calling the function. Calling a function that does nothing makes no sense to me.

kroky: Also, the modules that snooze and unsnooze need to be disabled as well. They are potentially dangerous. I think I say this for 3rd time already. Can you please make sure you check where those modules are and disable them depending on the setting?

Shadow243: Noted

Shadow243:
@DKing2222 check `Hm_Handler_imap_snooze_message` & `Hm_Handler_imap_unsnooze_message`.

We have added `should_skip_execution` that you can use.

modules/imap/site.js
Shadow243: > Also, the modules that snooze and unsnooze need to be disabled as well. They are potentially dangerous. I think I say this for 3rd time already. Can you please make sure you check where those modules are and disable them depending on the setting?

@kroky we've added https://github.com/cypht-org/cypht/pull/1528/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR999 & https://github.com/cypht-org/cypht/pull/1528/files#diff-cee0660c8870a8d10bc27d1f6179b5f67fc8fff6237ae2951b08edf2c8c14d5cR1048 to escape module execution for snooze and unsnooze

PR #1530, opened 2025-05-29
refactor(frontend/backend): performance optimization

kroky:
Refactor the way Cypht deals with mailbox connections in combined pages, single message pages and the way it gathers data sources: make single ajax call for each mailbox sorting on the frontend and aggregating like version 1.4 used to work; optimize folder name checks and unnecessary connects to the mail servers; various frontend routing updates and cache usage to speed things up.

Still a WIP and probably has quite some bugs... will be pushing updates in the coming days.


kroky:
master should be stabilized by this PR. Merging it now. Will pull to Tiki master and fix anything remaining there.

indridieinarsson:
@kroky : This works great! Thanks!!
 I just tested this real quick, and found one behaviour, I think it's a bug: 
I am using the "Feeds" functionality to subscribe to RSS feeds (yes, those still exist :) ). The entries of the subscribed RSS feeds seem to be not sorted along with the emails.  Instead, they appear at the top of the list, even if they are much older than all the emails in the list.

kroky:
Thanks for the report. I've tested feeds in combined inbox before merging - they were sorting but the second time they were fetched. First time they were prepended in the beginning. I will fix that but I think we have some other cache invalidation issues here and there to fix.

kroky:
Fixed feed sorting here: https://github.com/cypht-org/cypht/commit/b9434cfd31d0ff21652cdac6c9bdd7ca5f9caaa4

PR #1536, opened 2025-06-06
Fix shortcut refresh issue by adding redirect after update; improve folder page by using default profile to auto-select server

Shadow243:
Fixed two issues:

1. Shortcut update did not refresh the page — added a redirect after update to ensure changes are reflected immediately.
2. On the folder page, added support for selecting a default server using the default profile, instead of waiting for the user to manually choose one. This implements a feature requested by John.


Shadow243:
@kroky As requested, this PR is based on [#1490](https://github.com/cypht-org/cypht/pull/1490).

> **Feedback from previous PR:**  
> *"Why do you have all these unrelated updates when the PR is for Sieve connection? It is better to have PRs that are self-contained — easier to review, easier to trace in the history."*

I've now separated the changes so that this PR focuses solely on the shortcut & imap_folders

modules/imap_folders/modules.php
```diff
@@ -319,6 +319,18 @@ public function process() {
         if (array_key_exists('imap_server_id', $this->request->get)) {
             $this->out('folder_server', $this->request->get['imap_server_id']);
             $this->out('page', $this->request->get['page']);
+            $this->out('trigger_default_submit', false);
+        }else {
+            // as requested, if there no selected server we should select on by default based on profiles
+            if($this->module_is_supported('profiles')) {
```
kroky: I'd like to continue decoupling modules which means avoiding module_is_supported when possible (should be possible most of the times).
Please check Hm_Handler_compose_profile_data as an example - it is a profile module that is called on a compose page that is setup in smtp module and mainly used there. You can check how profile module unobtrusively adds functionality to the smtp module by adding profile data.
Please do something similar here - call a suitable method from profiles module (or create a new one for getting the default profile if none exists) but do it from the profiles module. You can use its setup.php to make the handler module  called in the exact place where folders_server_id is (or beforeit). Then, you will have the data needed for subsequent modules or you can use its data in folders_server_id module (here) to output default server id or whatever is needed.

Shadow243: @kroky I've submitted the update in a separate PR: https://github.com/cypht-org/cypht/pull/1549

kroky: Thanks!

PR #1538, opened 2025-06-06
fix logo in dark mode theme

Shadow243:
use logo_dark.svg when the used theme is darkly


Shadow243:
@kroky As requested, this PR is based on [#1490](https://github.com/cypht-org/cypht/pull/1490).

> **Feedback from previous PR:**  
> *"Why do you have all these unrelated updates when the PR is for Sieve connection? It is better to have PRs that are self-contained — easier to review, easier to trace in the history."*

I've now separated the changes so that this PR focuses solely on logo with dark theme(darkly)

Issue #1541, opened 2025-06-11
🐛 [Bug]Attempting to connect to database ..


ljcbaby:
## 🐛 Bug

I tries to use postgres but it runs not well.

![Image](https://github.com/user-attachments/assets/922a7539-b1da-4067-9887-8922fa98c111)

### Version & Environment

Rev: [2.4.0]

OS:  [cypht/cypht]

### Compose

```docker compose
services:
  cypht:
    image: cypht/cypht:2.4.0
    container_name: cypht
    restart: unless-stopped
    environment:
      - TZ=Asia/Shanghai
      - UMASK=0022
      - USER_CONFIG_TYPE=DB
      - ATTACHMENT_DIR=/var/lib/attachments
      - DEFAULT_LANGUAGE=zh-Hans
      - ALLOW_LONG_SESSION=true
      - DISABLE_IP_CHECK=true
      - DEFAULT_SETTING_TIMEZONE=Asia/Shanghai
      - DEFAULT_SETTING_LANGUAGE=zh-Hans
      - DB_CONNECTION_TYPE=host
      - DB_HOST=postgres
      - DB_PORT=5432
      - DB_NAME=cypht
      - DB_USER=cypht
      - DB_PASS=bf4610e60d38df1d956dbf567952895b
      - DB_DRIVER=pgsql
      - DEFAULT_SETTING_NO_PASSWORD_SAVE=false
      - FANCY_LOGIN=true
    networks:
      - apps
    ports:
      - 3008:80
    volumes:
      - /mnt/ssd/AppConfig/cypht:/var/lib/hm3
      - /mnt/Data/apps/cypht:/var/lib/attachments

networks:
  apps:
    external: true

x-portals:
- host: 192.168.3.4
  name: Web UI
  path: /
  port: 3008
  scheme: http
x-notes: 邮件聚合

```

ljcbaby:
I add these code and switch to daily image, it runs.

```compose
    entrypoint:
      - sh
      - -c
      - |
        docker-php-ext-install pdo_pgsql pgsql
        docker/docker-entrypoint.sh
```

PR #1542, opened 2025-06-11
fix(docker): can't run with pgsql

ljcbaby:
### Issues

- fixes: #1541 

### Todo

- [X] None


marclaporte:
Thank you @ljcbaby and welcome to the community!

FYI, here is where we hang out: https://gitter.im/cypht-org/community

PR #1543, opened 2025-06-11
feat(docker): auto release image by tag

ljcbaby:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/Release-Image.yml
+++ b/.github/workflows/Release-Image.yml
@@ -1,0 +1,44 @@
+name: Build Daily Image
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build-and-push:
+    if: github.repository_owner == 'cypht-org'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Get tags
+        id: tags
+        run: |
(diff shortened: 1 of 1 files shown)


marclaporte:
Thank you @ljcbaby

jonocodes:
This is going to build every tag, not just releases? Are you concerned about running out of space in docker hub? Maybe you want to constrain it to only building releases.

ljcbaby:
> This is going to build every tag, not just releases? Are you concerned about running out of space in docker hub? Maybe you want to constrain it to only building releases.

```yaml
on:
  release:
    types: [created]
```

This defines that only release is executed when it is created

ljcbaby:
> This is going to build every tag, not just releases? Are you concerned about running out of space in docker hub? Maybe you want to constrain it to only building releases.

if DockerHub org storage space is limited, consider GHCR ?

jonocodes (approved):

marclaporte:
We have the plan for sponsored Open Source projects so we probably have high limits (to be verified)
https://docs.docker.com/docker-hub/repos/manage/trusted-content/dsos-program/

We already have daily tags as per https://hub.docker.com/r/cypht/cypht/tags
Do we keep these indefinitely?
The Cypht project doesn't have that many commits per day so this permits flexibility.

A tag per commit goes even further and can facilitate the testing process (as would an easy to test build of each PR which is deleted once PR is merged in). Would we keep these indefinitely?

ljcbaby:
> We have the plan for sponsored Open Source projects so we probably have high limits (to be verified) https://docs.docker.com/docker-hub/repos/manage/trusted-content/dsos-program/
> 
> We already have daily tags as per https://hub.docker.com/r/cypht/cypht/tags Do we keep these indefinitely? The Cypht project doesn't have that many commits per day so this permits flexibility.
> 
> A tag per commit goes even further and can facilitate the testing process (as would an easy to test build of each PR which is deleted once PR is merged in). Would we keep these indefinitely?

This PR doesn't involve a development or testing process, just a release with a semantic version number that builds the corresponding version image and updates the latest tag.

PR #1549, opened 2025-06-15
refactor(frontend/backend): improve folder page by using default profile to auto-select server

Shadow243:
On the folder page, added support for selecting a default server using the default profile, instead of waiting for the user to manually choose one. This implements a feature requested by John.
 
 I separated from https://github.com/cypht-org/cypht/pull/1536

### DIFF
--- a/lib/servers.php
+++ b/lib/servers.php
@@ -216,6 +216,27 @@ trait Hm_Server_List {
     }
 
     /**
+     * Return server details matching $match in column $column.
+     *
+     * @param mixed  $match        Value to match.
+     * @param string $column       Column name to match against.
+     * @param bool   $returnFirst  If true, return only the first matching server.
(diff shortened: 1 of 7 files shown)


Shadow243:
> I'd like to continue decoupling modules which means avoiding module_is_supported when possible (should be possible most of the times). Please check Hm_Handler_compose_profile_data as an example - it is a profile module that is called on a compose page that is setup in smtp module and mainly used there. You can check how profile module unobtrusively adds functionality to the smtp module by adding profile data. Please do something similar here - call a suitable method from profiles module (or create a new one for getting the default profile if none exists) but do it from the profiles module. You can use its setup.php to make the handler module called in the exact place where folders_server_id is (or beforeit). Then, you will have the data needed for subsequent modules or you can use its data in folders_server_id module (here) to output default server id or whatever is needed.

I've updated

PR #1553, opened 2025-06-18
fix(other): profile tests and make server_stmp_and_imap_add work correclty with a real smtp & imap server

Shadow243:


.github/workflows/Test-Build.yml
```diff
@@ -16,78 +16,78 @@ on:
...
+  #     - name: "Script: setup.sh"
+  #       run: bash .github/tests/setup.sh
+
+  #     - name: "Composer Install Dependencies"
+  #       run: |
+  #           composer install
+  #           composer require --dev php-coveralls/php-coveralls
+
+  #     - name: "Script: test.sh"
+  #       run: bash tests/phpunit/run.sh
```
kroky: Why do you comment out running the unit tests?

tests/selenium/runall.sh
```diff
@@ -3,7 +3,8 @@
 PYTHON=$(command -v python3)
 rm -rf __pycache__/
 
-for suite in login.py folder_list.py pages.py profiles.py settings.py servers.py send.py search.py inline_msg.py keyboard_shortcuts.py
+for suite in login.py servers.py profiles.py send.py search.py inline_msg.py keyboard_shortcuts.py
+# for suite in login.py folder_list.py pages.py settings.py servers.py profiles.py send.py search.py inline_msg.py keyboard_shortcuts.py
```
kroky: Why are we skipping some tests?

Shadow243: I’m running some tests. It’s taking too long to wait. Once it works, I’ll uncomment others.

kroky: Ok, then put this in draft and only open it when it is ready to be reviewed.

PR #1565, opened 2025-07-02
fix(frontend): show meaningful message when email is already moved

josaphatim:
## 🍰 Pullrequest
This PR fixes a use case when a message is deleted or moved but still in user history (it show a warning and disable actions.

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -430,14 +430,16 @@ class Hm_Alert {
         
         this.container.appendChild(alert);
 
-        setTimeout(() => {
-            alert.style.opacity = '0';
-            alert.style.transform = 'translateY(-20px)';
+        if (dismissible) {
             setTimeout(() => {
-                const bsAlert = new bootstrap.Alert(alert);
-                bsAlert.close();
-            }, 500);
-        }, dismissTime * 1000);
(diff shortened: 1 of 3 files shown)


kroky:
I don't like the approach to store deleted flag for such messages in local cache. I'd rather redirect to parent (message list or combined page) based off the url params and display a warning there that the message they wanted to view is no longer available on the server.

josaphatim:
@kroky I understand. But which is the best approach ? Should we delete the message in cache when the message is deleted ? Because here the message is first loaded from cache and showed to the user then fetched from server. user will see message for some seconds before it disappear

josaphatim:
Another reason is that I'm thinking of showing a link of the new message location (like gmail does) in case the message as been moved (archive/move or other actions)

kroky:
I think we should show a blocking modal msg that this msg has been removed on the server and then user clicks 'Ok' to redirect to the parent page (or moved message folder).

josaphatim:
Hello @kroky
I did necessary changes

kroky:
We should resolve the pipeline before merging.

josaphatim:
Everything okay for now

PR #1567, opened 2025-07-03
feat(backend): servers count on servers page

Shadow243:
Fixed server count using filter to remove EWS when displaying.

### DIFF
--- a/modules/core/output_modules.php
+++ b/modules/core/output_modules.php
@@ -2285,7 +2285,8 @@ class Hm_Output_server_config_stepper extends Hm_Output_Module {
         $hasJmapActivated = in_array('jmap', $this->get('router_module_list'), true);
 
         if($hasImapActivated){
-            $imap_servers_count = count(array_filter($this->get('imap_servers', array()), function($v) { return !array_key_exists('type', $v) || $v['type'] == 'imap'; }));
+            $servers = array_filter($this->get('imap_servers', array()), fn($s) => ($s['type'] ?? null) !== 'ews');
+            $imap_servers_count = count($servers);
             $accordionTitle .= 'IMAP';
             $configuredText .=  '<span class="imap_server_count"> ' . $imap_servers_count .'</span> IMAP';
             $hasEssentialModuleActivated = true;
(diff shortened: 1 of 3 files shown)


modules/smtp/modules.php
```diff
@@ -1283,7 +1283,7 @@ protected function output() {
         if ($this->get('single_server_mode')) {
             return '';
         }
-        $count = count($this->get('smtp_servers', array()));
+        $count = count(array_filter($this->get('smtp_servers', array()), fn($s) => ($s['type'] ?? null) !== 'ews'));
```
kroky, line 1286: SMTP servers cannot have ews type. EWS servers are added to the imap list and distinguished there by the type.

modules/imap/output_modules.php
```diff
@@ -447,7 +447,7 @@ protected function output() {
         if ($this->get('single_server_mode')) {
             return '';
         }
-        $list = $this->get('imap_servers', array());
+        $list = array_filter($this->get('imap_servers', array()), fn($s) => ($s['type'] ?? null) !== 'ews');
```
kroky, line 450: Why wouldn't we want ews servers listed here? They should be managed via the imap servers page but just their type is different.

modules/core/output_modules.php
```diff
@@ -2304,7 +2305,8 @@ protected function output() {
         }
 
         if($hasSmtpActivated){
-            $smtp_servers_count = count($this->get('smtp_servers', array()));
+            $servers = array_filter($this->get('smtp_servers', array()), fn($s) => ($s['type'] ?? null) !== 'ews');
```
kroky, line 2308: SMTP servers don't need a filter here - the are only smtp ones.

modules/core/output_modules.php
```diff
@@ -2285,7 +2285,8 @@ protected function output() {
         $hasJmapActivated = in_array('jmap', $this->get('router_module_list'), true);
 
         if($hasImapActivated){
-            $imap_servers_count = count(array_filter($this->get('imap_servers', array()), function($v) { return !array_key_exists('type', $v) || $v['type'] == 'imap'; }));
+            $servers = array_filter($this->get('imap_servers', array()), fn($s) => ($s['type'] ?? null) !== 'ews');
```
kroky, line 2288: You will include jmap ones here but they have a separate entry below. Why did you change this code at all?

modules/core/output_modules.php
Shadow243: Even when I use exit(var_dump(Hm_SMTP_List::dump())) inside Hm_Handler_add_smtp_servers_to_page_data, the EWS server is still included.

Shadow243:
Here is the picture of what is displayed before fix: 
Screenshot 2025-07-04 at 14 50 02

I only have one EWS server configured

modules/core/output_modules.php
kroky: Oh, maybe I forgot that they are included in smtp list as well. The, it is fine but be sure to not remove them from somewhere, they are actually used as EWS is used for both - incoming (read - e.g. imap) and outgoing (i.e. smtp).

PR #1570, opened 2025-07-06
fix(frontend): navigation not working on mobile

Shadow243:
fix navigation that is not working on mobile, some videos available at [item136429](https://avan.tech/item136429)

### DIFF
--- a/modules/core/navigation/navigation.js
+++ b/modules/core/navigation/navigation.js
@@ -53,6 +53,7 @@ async function navigate(url, loaderMessage) {
     try {
         const response = await fetch(url, {
             method: 'GET',
+            credentials: 'include',
         });
 
         if (!response.ok) {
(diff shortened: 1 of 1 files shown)


modules/core/navigation/navigation.js
```diff
@@ -53,25 +53,50 @@ async function navigate(url, loaderMessage) {
...
+            throw new Error("Empty or invalid response body.");
+        }
+        // const mainMatch = html.match(/<main[^>]*>((.|[\n\r])*)<\/main>/i);
+        // const titleMatch = html.match(/<title[^>]*>((.|[\n\r])*)<\/title>/i);
+
+        // if (!mainMatch || !titleMatch) {
+        //     throw new Error("Could not find <main> or <title> in HTML.");
+        // }
+        // const main = mainMatch[0];
+        // const title = titleMatch[0];
```
mercihabam: Could you get rid of the commented code?

modules/core/navigation/navigation.js
```diff
@@ -95,7 +120,9 @@ async function navigate(url, loaderMessage) {
         trackLocationSearchChanges();
     } catch (error) {
         Hm_Notices.show(error.message, 'danger');
+        console.log(error.message);
         console.log(error);
+        // window.location.href = url;
```
mercihabam: Could you also remove these debugging messages?

modules/core/navigation/navigation.js
```diff
@@ -53,25 +53,42 @@ async function navigate(url, loaderMessage) {
...
         }
 
         const html = await response.text();
-        const main = html.match(/<main[^>]*>((.|[\n\r])*)<\/main>/i)[0];
-        const title = html.match(/<title[^>]*>((.|[\n\r])*)<\/title>/i)[0];
+        if (!html || typeof html !== 'string') {
+            throw new Error("Empty or invalid response body.");
+        }
+
+        const parser = new DOMParser();
```
kroky: I'd prefer to keep the regex for now. We are using this navigation for each and every link in Cypht and full-fledged DOM parsing for each page load seems a bit too much to me (performance wise). We will end up parsing it once to extract just the title and main section, then converting to cash.js elements and then replacing existing DOM structures with that. We can skip at least one of these 3 resource-intensive steps.

modules/core/navigation/navigation.js
```diff
@@ -53,25 +53,42 @@ async function navigate(url, loaderMessage) {
...
+        const doc = parser.parseFromString(html, 'text/html');
+
+        const main = doc.querySelector('main');
+        const title = doc.querySelector('title');
 
+        if (!main || !title) {
+            throw new Error("main or title not found in parsed HTML");
```
kroky: This error is sent to the end user but I assume that very small number of Cypht end users will actually know what does this mean. Can you improve the error messages here connected with navigation? Maybe add a nice user-facing error msg like 'Failed navigating to page ...' and then output the actual error in the console as a warning or error, so developers can look into that?

modules/core/navigation/navigation.js
Shadow243: @kroky i tried with regex, i'm now getting the error i was trying to fix. 
CleanShot 2025-07-13 at 01 27 31

I'm looking for an other way to make it work with regex as you requested.

kroky: What is the server error? Do you have the error message somewhere?

Shadow243: The issue came from the regex used for `<main>`. On some pages, `<main>` contains more complex structures, including inline scripts, which the initial regex didn’t handle, causing the `Cannot read properties of null` error when trying to access `[0]`. Using DOMParser worked correctly but was too heavy for every navigation. By refining the regex (tested on https://regexr.com/), we can now reliably capture `<main>` and navigation works as expected, while keeping performance intact.

PR #1571, opened 2025-07-09
fix(frontend): unblock sender action, block domain action and edit behavior not working

christer77:


modules/sievefilters/functions.php
```diff
@@ -308,9 +308,9 @@ function prepare_sieve_script ($script, $index = 1, $action = "decode")
         if ($script != '') {
             $base64_obj = str_replace("# ", "", preg_split('#\r?\n#', $script, 0)[$index]);
             if ($action == "decode") {
-                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)), true);
+                $blocked_list = json_decode(str_replace("*", "", base64_decode($base64_obj)), true) ?? [];
```
kroky: Shouldn't we catch any json parsing errors and report to the user? Falling silently to default empty value will be odd for some users trying to see why it is not working.

modules/sievefilters/site.js
```diff
@@ -196,76 +196,110 @@ function blockListPageHandlers() {
...
-            const reject = parent.find('.select_default_reject_message');
-            payload.push({'name': 'reject_message', 'value': reject.val()});
-        }
 
-        submit.attr('disabled', 1);
-        Hm_Ajax.request(
-            payload,
-            function(res) {
-                submit.removeAttr('disabled');
+    $(function() {
```
kroky: Why do you wrap this in a cash.js function call? Adding event handlers to document should deal with document not being loading initially, so this shouldn't be necessary.

christer77: I understood that I cannot find the child on which we click. 
For example, `$(document).on('click', '.unblock_button', function(e) {.` 

So I will manage it with the parent as closely as possible.

`$("#sieve_accounts").on('click', '.unblock_button', function(e) {`

kroky: I think the attaching to the document element was done on purpose because of these being loaded in a modal. Can't you get the child element by e.target or e.currentTarget?

modules/sievefilters/functions.php
christer77: So the problem usually arises when we pass a $index that is not found.
For example, when we try to do as follows: `$blocked_list_actions = prepare_sieve_script($current_script, 2);`.
Here `prepare_sieve_script` returns null when trying to retrieve a list of actions that is not available in our ` $current_script.`

modules/sievefilters/site.js
christer77: I tried to think, but I don't understand how I can use them in the current file to access the element `'.unblock_button'` for example. Can you clarify further, please?

modules/sievefilters/functions.php
kroky: Previously, if you requested index 2 and it was not available, there would have been a warning of some sort that you request an index that's not available or you try to access an offset on null or something similar. Now, you just silently fail which is not good. We mask a problem that calling code must fix - why does it request non-existing index? Maybe it gets some data incorrectly, maybe cached data is wrong, may be one of a dozen more reasons. At any rate, user will deserve to know what is going on. We should either fix the calling code to stop requesting invalid indexes or tell the user that sieve script is not right and have them fix another way. I still think that it is a good idea to check for json parsing errors and report to the user, even if we return an empty array. Also, we can add a special log message if a non-existing index is requested and debug more to see why it is requested.

modules/sievefilters/site.js
kroky: Well, according to https://learn.jquery.com/events/event-delegation/ which is more or less implemented in cash.js we are using, having $(document).on('click', '.unblock_button', function() { console.log($(this)); }); should log you the element you clicked on... I don't see a reason why you should be more specific in this binding events. What is `this` pointing to if you leave the bindings on the document element?

modules/sievefilters/functions.php
christer77: If we go back to the concept of what this resolves, you will notice that when we try to block an address and it wasn't previously in the list, that is where the bug occurs, simply because in the file `module/ssievefilters/modules.php` at line `561 `:`} elseif (array_key_exists($blocked_sender, $blocked_list_actions)) {`, we expect to have an array (which may not contain data) from `$blocked_list_actions.`
But unfortunately, our function `prepare_sieve_script` returns null at line `531 `via `$blocked_list_actions = prepare_sieve_script ($current_script, 2);`

modules/sievefilters/site.js
christer77: That's exactly the problem, if we consider `$(document).on('click', '.unblock_button', function() { console.log($(this)); });,` there is no event that is triggered. It's as if this is not taken into account.

modules/sievefilters/functions.php
kroky: What is the content of $current_script? It seems either invalid or we need to handle it when it doesn't contain the syntax we expect - i.e. initialize with proper syntax. I don't see any of these in the proposed changes.

modules/sievefilters/site.js
kroky: Then: either event handler was removed by other code or the selector is not present (invalid selector). We have kind of a strange navigation at the moment where ajax calls get the whole body replaced with the content of new pages which might introduce an issue with event handlers attached to the document level. Can you debug more to see when exactly do the event handler get removed? I assume we will find a deeper issue here to fix - probably connected with the navigation.

christer77: Alright,

christer77: I found the real cause, by the way, in this code:
`$(document).off('click').on('click', '.sievefilters_accounts_title', function() {$(this).parent().find('.sievefilters_accounts').toggleClass('d-none');});`

The event`$(document).off('click')` removes all `click` handlers on the document.

That's why to remove only the handlers for `.sievefilters_accounts_title`, 

I did:`$(document).off('click', '.sievefilters_accounts_title').on('click', '.sievefilters_accounts_title', function() {$(this).parent().find('.sievefilters_accounts').toggleClass('d-none');});`

modules/sievefilters/functions.php
christer77: Hello @kroky,

Your skepticism made sense regarding what was done here; 

by the way, 

I restarted everything and found that this was not the source of the problem. However, it was in the file "modules\sievefilters\modules.php". In this file, the handling of email addresses coming from the "screen email" action was not managed properly. 
I think it is now the case.

Thank you.

modules/sievefilters/site.js
kroky: Ah, nice catch, thank you!

Issue #1499, opened 2025-05-01
🚀 [Feature] Automatically Create Default User Profiles


Revisto:
## 🚀 Feature
Automatically create user profiles in Cypht based on configured IMAP and SMTP servers, reducing manual setup and improving the usability of features that rely on profiles.

Currently, users must manually create Profiles (`Settings -> Profiles`) to link IMAP and SMTP accounts, define display names, signatures, etc.

christer77:
Hello @Revisto , 

Can you please close this if it is resolved by https://github.com/cypht-org/cypht/pull/1577.

PR #1577, opened 2025-07-18
feat(frontend/backend): add feature create default user profiles quickly

christer77:
https://github.com/cypht-org/cypht/issues/1499

### DIFF
--- a/modules/profiles/modules.php
+++ b/modules/profiles/modules.php
@@ -92,6 +92,16 @@ class Hm_Handler_process_profile_update extends Hm_Handler_Module {
         }
         list($success, $form) = $this->process_form(array('profile_id', 'profile_address',
             'profile_smtp', 'profile_replyto', 'profile_name', 'profile_imap', 'profile_rmk'));
+            
+        if (isset($this->request->post['profile_quickly_create_value']) && $this->request->post['profile_quickly_create_value'] === "yes") {
+            $array_profil_imap = explode("|", $form['profile_imap']);
+            if (count($array_profil_imap) == 2) {
+                $profile_name = explode("@", $array_profil_imap[1]);
+                $form['profile_name'] = $profile_name[0];
+                $form['profile_address'] = $array_profil_imap[1];
+                $form['profile_replyto'] = $array_profil_imap[1];
+            }
+        }
         if (!$success) {
             return;
         }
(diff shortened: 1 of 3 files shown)


modules/profiles/modules.php
```diff
@@ -345,21 +354,27 @@ function profile_form($form_vals, $id, $smtp_servers, $imap_servers, $out_mod) {
     $res .= '<div class="edit_profile row p-3" '.($form_vals['name'] ? '' : 'style="display: none;"').'><div class="col-12 col-lg-8 col-xl-5">';
 
     $res .= '<form method="post" action="?page=profiles">';
+    if (empty($form_vals['id'])) {
+        $res .= '<div class="form-check form-switch mt-3 mb-3">';
+        $res .= '<input class="form-check-input" name="profile_quickly_create" type="checkbox" role="switch" onchange="handleQuicklyCreateProfileCheckboxChange(this)" id="profile_quickly_create">';
```
kroky: Event handlers are registered in module's site.js, route_handlers or whatever other places are used in other modules. Can we be consistent here and avoid registration from the DOM?

modules/profiles/modules.php
```diff
@@ -90,6 +90,15 @@ public function process() {
         if (array_key_exists('profile_delete', $this->request->post)) {
             return;
         }
+        if (isset($this->request->post['profile_quickly_create']) && $this->request->post['profile_quickly_create'] == "on") {
```
kroky: Please use a specific value on the checkbox instead of relying on a default browser value.

modules/profiles/modules.php
```diff
@@ -90,6 +90,15 @@ public function process() {
...
+            $array_profil_imap = explode("|", $this->request->post['profile_imap']);
+            if (count($array_profil_imap) == 2) {
+                $profile_name = explode("@", $array_profil_imap[1]);
+                $this->request->post['profile_name'] = $profile_name[0];
```
kroky: Modifying request variables besides the purpose of validation or sanitizing is usually not recommended as their purpose of being user input is changed here. That's why usually frameworks make the access to request variables read-only. Cypht might go this way in the future, so can you please process the form first and instead of modifying the request variables, modify the form variable which is our own custom variable and can be safely modified?

Issue #676, opened 2023-02-22
We can now search in a folder. Let's add "include subfolders", "all folders" and special folders (sent, draft, archive, etc.)


marclaporte:
The current search in Cypht is quite impressive in that you can search throughout various mail accounts, which is not possible in Roundcube as per: https://github.com/roundcube/roundcubemail/issues/4972

https://github.com/jasonmunro/cypht/tree/master/modules/advanced_search is very powerful, but there is a an issue that you need to indicate each folder you want to search in.

In contrast, here is what Roundcube offers:
![2023-02-22_103531](https://user-images.githubusercontent.com/1004261/220673325-83eca503-b51b-41f4-a081-6acce8472733.png)

So let's add what Roundcube does. And given Cypht's multi-account feature, we can also add:
* Search in all sent folders
* Search in all draft folders
* Search in all archive folders (here subfolders would be useful)
* etc.

It will be really efficient when combined with https://github.com/jasonmunro/cypht/issues/306

marclaporte:
Also, there should be a filter for subscribed vs not subscribed, as non-subscribed folders are often more archival and thus, not so interesting to search through (of course, sometimes they are so we still want to be able to)
https://github.com/jasonmunro/cypht/pull/622

marclaporte:
"Is it too much to ask, for example, for a webmail client with global search? The only one that I found so far is Roundcube, which can do a global search (all parts, all folders) with "just" 4 additional clicks. Why is that? I had a server running Horde Groupware in 2013 that could do that."
https://www.reddit.com/r/selfhosted/comments/16d3ae0/the_sad_state_of_selfhosted_webmail/

I agree. We will work on this.

marclaporte:
Related discussion in SnappyMail: https://github.com/the-djmaze/snappymail/issues/196

TheDigitalOrchard:
I like how Apple Mail handles searches, by presenting a "scope selection" as the search term is being entered.

<img width="420" alt="apple-mail-search" src="https://github.com/cypht-org/cypht/assets/3195423/f26fff6e-f080-4cdd-8a20-0484593c3e02">

Selecting a scope is optional, but very intuitive. The tag-based interface is easy to tweak at any time.

<img width="387" alt="apple-mail-search-scope" src="https://github.com/cypht-org/cypht/assets/3195423/19273671-56b7-4d6a-b254-24d92710962b">

[...]

PR #1582, opened 2025-07-22
feat(backend): allow search flagged in all folders if enabled in settings

Shadow243:
This PR introduces a new parameter that allows Cypht to search for flagged messages across **all folders**.  
By default, the search is limited to the **INBOX**, which is lightweight and optimized for performance.

Enabling the new option allows a broader search across the entire server, but it's **more resource-intensive**.  
Therefore, users should be aware of the performance implications before activating it.

Reported at [https://avan.tech/item136195#threadId91599](https://avan.tech/item136195#threadId91599) as point 5


modules/imap/handler_modules.php
```diff
@@ -1317,33 +1323,54 @@ public function process() {
...
-                } elseif (isset($headers[bin2hex($uid)])) {
-                    $msg = $headers[bin2hex($uid)];
-                } else {
-                    continue;
+            if($allow_all_folders) {
+                $all_folders = $mailbox->get_folders();
+                foreach (array_keys($all_folders) as $folder) {
+                    $uids = $mailbox->search($folder, $filter, $terms, $sort, $reverse);
+                    $headers = $mailbox->get_message_list($folder, $uids);
+                    $uids = array_slice($uids, 0, $limit);
```
kroky, line 1519: How do you apply the limit across folders? Limit is flagged per source, so it should be decreased when a folder returns some results.

kroky:
@Shadow243 why is this still in draft? Isn't it ready for review?

marclaporte:
Brainstorming: There is often this trade off of search here (fast) vs search everywhere (slow).

It's not just for flagged emails, but also for tags and general search. I wonder if we should generalize this?

So the default is the quick search, and when relevant, add a link/button/option to extend this search to all folders (or something)

"Did you find what you were looking for? Would you like to expand this search to all folders of this account or all folders of all accounts?" (or something like this)

See also: https://github.com/cypht-org/cypht/issues/676

This is not a blocker for 2.5.0 but quite important, so I will put it for milestone 2.6

Let's discuss at our https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings

Shadow243:
> Brainstorming: There is often this trade off of search here (fast) vs search everywhere (slow).
> 
> It's not just for flagged emails, but also for tags and general search. I wonder if we should generalize this?
> 
> So the default is the quick search, and when relevant, add a link/button/option to extend this search to all folders (or something)
> 
> "Did you find what you were looking for? Would you like to expand this search to all folders of this account or all folders of all accounts?" (or something like this)
> 
> See also: #676
> 
> This is not a blocker for 2.5.0 but quite important, so I will put it for milestone 2.6
> 
> Let's discuss at our https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings

Screenshot 2026-05-09 180559

Shadow243:
@kroky this is ready for an other review.

modules/imap/handler_modules.php
```diff
@@ -1434,6 +1437,9 @@ public function process() {
                 $date = process_since_argument($this->user_config->get('all_since_setting', DEFAULT_SINCE));
                 break;
             case 'flagged':
+                $filter = 'FLAGGED';
+                $date = process_since_argument($this->user_config->get('flagged_since_setting', DEFAULT_FLAGGED_SINCE));
+                $limit = $this->user_config->get('flagged_per_source_setting', DEFAULT_FLAGGED_PER_SOURCE);
```
kroky, line 1460: You'd need a break here to keep these overridden values...

modules/imap/handler_modules.php
```diff
@@ -1476,27 +1483,54 @@ public function process() {
...
+                            $msg = $headers[bin2hex($uid)];
+                        } else {
+                            continue;
+                        }
+                        $msg['server_id'] = $id;
+                        $msg['server_name'] = $details['name'];
+                        $msg['folder'] = $folders[$key];
+                        $messages[] = $msg;
+                    }
+                    $status['imap_'.$id.'_'.bin2hex($folder)] = $mailbox->get_folder_state();
```
kroky: You repeat 20+ lines in the two blocks. Please DRY it up.

kroky:
@Shadow243 can you move it out of draft if it is ready from your side. Also, please respond to individual comments when they are ready/addressed with some small details.

modules/imap/handler_modules.php
Shadow243: Added break after the flagged case;

modules/imap/handler_modules.php
Shadow243: Unified the two branches into a single `foreach ($search_folders as $folder)` loop. The only remaining branch is the 4 lines that differ between expand and non-expand (how `$uids` is sliced).

PR #1585, opened 2025-07-25
fix(frontend): make profile_replyto optional

Shadow243:
As requested in the previous Cypht Test meeting, we are not obliged to put the profile_replyto, it's optional


kroky:
@Shadow243 can you please resolve the conflict and merge after that - this is OK to be merged?

PR #1586, opened 2025-07-25
fix(frontend): redesign edit draft button to make it more visible

Shadow243:
CleanShot 2025-07-25 at 21 31 15@2x

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -323,6 +323,10 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
                     }
                 }
             }
+            $is_draft = isset($headers['Flags']) && mb_stristr($headers['Flags'], 'draft');
+            if($is_draft) {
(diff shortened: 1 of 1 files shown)


modules/imap/site.css
```diff
@@ -138,14 +138,18 @@
...
+.header_links #dropdownMenuForward, .header_links #dropdownMenuSnooze, .header_links #dropdownMenuTag, .header_links #dropdownMenuNexterDate {
   padding: 0;
   border: unset;
   font-variant: inherit;
   text-transform: inherit;
   font-size: inherit;
   vertical-align: baseline;
 }
+.header_links .dropdown .btn:hover, .header_links .dropdown .btn:active, .header_links .dropdown .btn:focus {
+    color: inherit !important;
```
kroky: Can you please avoid using important here in the stock css themes? There would be no way to override for the user afterwards.

Shadow243: Copy that, i will also apply https://github.com/cypht-org/cypht/pull/1594#discussion_r2261840034

kroky:
@Shadow243 can you please check the tests - the failures might be related with some css/DOM changes in this MR.

kroky:
Actually, saw it while merging others as well, so it should be something else. Please still check it but I will merge this one.

PR #1588, opened 2025-07-25
feat(contact): add enbale collect addresses while sending a mail

Shadow243:
Task: [item64946](https://avan.tech/item64946)


modules/smtp/js_modules/route_handlers.js
```diff
@@ -67,6 +67,21 @@ function applySmtpComposePageHandlers() {
                 showBtnSendAnywayDontWarnFuture = false;
             }
 
+            // Below we add emails to Trusted sender
+            if (hm_enable_collect_address_on_send) {
+                const parseEmails = (selector) => {
+                    return $(selector).val().split(',').map(e => e.trim()).filter(Boolean);
```
kroky: Isn't it possible to have multiple addresses separated by a semicolon as well? Why don't we parse them in the backend using a suitable lib like the compose action does?

Shadow243: Sure , multiple addresses can be separated by semicolons. Currently, Cypht officially uses commas (,) as separators in the compose fields. Even if you enter addresses separated by semicolons (;), the frontend automatically reformats them to commas behind the scenes.

That said, I’ve extended the backend parsing to also support semicolons just in case, so it’s more flexible and can handle both separators if needed.

kroky: ok, thanks!

PR #1591, opened 2025-07-26
fix(frontend): Prevent the duplication of footer buttons when reopening the modal

christer77:
**Problem:**
During the repeated opening of the Hm_Modal modal, the custom button (Add Emails to Trust contact) was duplicated each time, creating multiple instances of the same button.

**Cause:** 
The addFooterBtn method systematically added a new button without removing the previous ones, leading to duplication with each call.

add-to-trust-contact-modal

### DIFF
--- a/modules/core/site.js
+++ b/modules/core/site.js
@@ -312,7 +312,7 @@ var Hm_Ajax_Request = function() { return {
  * Show a modal dialog with a title, content and buttons.
  */
 function Hm_Modal(options) {
-    var defaults = {
+    const defaults = {
         title: 'Cypht',
         size: '',
         btnSize: '',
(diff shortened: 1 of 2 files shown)


modules/core/site.js
```diff
@@ -369,6 +369,8 @@ function Hm_Modal(options) {
     };
 
     this.addFooterBtn = (label, classes, callback) => {
+        this.modalFooter.find('button').filter((_, btn) => btn.textContent.trim() === label).remove();
```
kroky: If it is called multiple times on the same modal, there might be other issues with duplicate event handlers or other duplicated content (even if they are not present now, they can show up in the future). I think it is better if we destroy the modal properly before showing it up again (or re-initializing it again). The code that calls this seems to initialize another instance of Hm_Modal but maybe it is reusing dom components which will lead to further issues. When we create a new modal, we need to make sure it is really new in terms of DOM components.

christer77: Noted.

PR #1580, opened 2025-07-22
Add-a-spinner-at-least-until-some-messages-are-shown

espoirbaraka:
As part of, or after Bootstrap 5: Cypht: Add a spinner at least until some messages are shown

- In lists
- For a message
- Search
- etc.

Related task : [avan.tech](https://avan.tech/item38377)

PR #1126, opened 2024-07-21
Add-a-spinner-at-least-until-some-messages-are-shown

David-Muhasa:
As part of, or after Bootstrap 5: Cypht: Add a spinner at least until some messages are shown
In lists
For a message
Search
etc.

PR #1595, opened 2025-08-01
feat(frontend): introduce NProgress and Cypht custom spinner using Bootstrap5

Shadow243:
This PR introduces NProgress and a custom Cypht spinner for improved visual feedback during asynchronous operations. Integrated using Bootstrap 5 styling.
CleanShot 2025-08-01 at 03 25 05

### DIFF
--- a/modules/core/js_modules/Hm_MessagesStore.js
+++ b/modules/core/js_modules/Hm_MessagesStore.js
@@ -220,6 +220,10 @@ class Hm_MessagesStore {
     }
 
     fetch(hideLoadingState = false) {
+        // show my custom cypht spinner(class = cypht-spinner)
+        const spinner = document.querySelector('.cypht-spinner');
+        if (spinner) spinner.style.display = 'block';
+
         let store = this;
         return this.getRequestConfigs().map((config) => {
             return new Promise((resolve, reject) => {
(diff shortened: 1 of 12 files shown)


Shadow243:
The updates from https://github.com/cypht-org/cypht/pull/1580 have been included, along with additional improvements to the spinner and the integration of NProgress using Bootstrap 5.

Requested a long time ago in https://github.com/cypht-org/cypht/pull/1126#issuecomment-2282889459

modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -237,6 +241,9 @@ class Hm_MessagesStore {
                     reject,
                     this.abortController?.signal
                 );
+            }).finally(() => {
+                // Hide the spinner
+                if (spinner) spinner.style.display = 'none';
```
kroky, line 246: These can run several in parallel. Don't we want to hide the spinner only after the last one finishes?

Shadow243: Yes, I agree it would make sense for the spinner to remain visible until all requests have completed.
In this case, I hid it earlier to let NProgress act as the main loading indicator, once results begin appearing, NProgress still shows that loading is in progress.

This gives the user immediate feedback that some data is available, while still showing that more is being loaded in the background.
That said, we could absolutely adjust it so the spinner also stays visible until the very end if we prefer that behavior.

kroky: If the progress is still showing, then we should be fine.

PR #1596, opened 2025-08-04
fix(backend): test failing with latest updates

Shadow243:


Shadow243:
I’m currently quite exhausted, so I’ll revisit and continue this later today when I’m more refreshed to ensure better focus and quality.

Shadow243:
@kroky  I finally tracked down the culprit 🕵️ 

Fixed unexpected spaces/newlines in list_controls() caused by multi-line string formatting and a small typo (classs → class). Clean output now.

PR #1601, opened 2025-08-08
fix(frontend): list the sieve filter whose number is greater than 0 i…

christer77:


christer77:
I am closing this PR because its expected behavior will no longer allow the creation of additional filters if there are 0 filters in a mailbox.

PR #1604, opened 2025-08-09
refactor(backend): improve block page performance by reusing Sieve connections

Shadow243:
As discussed with @kroky via telegram, we aim to minimize the number of Sieve server reconnections to improve the block page performance.

### DIFF
--- a/modules/sievefilters/functions.php
+++ b/modules/sievefilters/functions.php
@@ -627,7 +627,6 @@ if (!hm_exists('get_sieve_linked_mailbox')) {
                     $folders[$s] = $obj->value;
                 }
             }
-            $client->close();
             return $folders;
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
--- a/modules/sievefilters/modules.php
+++ b/modules/sievefilters/modules.php
@@ -41,7 +41,6 @@ class Hm_Handler_sieve_edit_filter extends Hm_Handler_Module {
             } else {
                 $this->out('test_type', 'ANYOF');
             }
-            $client->close();
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
             return;
@@ -109,7 +108,6 @@ class Hm_Handler_sieve_edit_script extends Hm_Handler_Module {
         try {
             $client = $factory->init($this->user_config, $imap_account, $this->module_is_supported('nux'));
             $script = $client->getScript($this->request->post['sieve_script_name']);
-            $client->close();
             $this->out('script', $script);
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -158,7 +156,6 @@ class Hm_Handler_sieve_delete_filter extends Hm_Handler_Module {
             $main_script = generate_main_script($scripts);
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
             Hm_Msgs::add('Script removed');
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -198,7 +195,6 @@ class Hm_Handler_sieve_delete_script extends Hm_Handler_Module {
 
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
             Hm_Msgs::add('Script removed');
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -277,7 +273,6 @@ class Hm_Handler_sieve_block_domain_script extends Hm_Handler_Module {
             );
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
             $this->out('reload_page', true);
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -439,7 +434,6 @@ class Hm_Handler_sieve_unblock_sender extends Hm_Handler_Module {
             );
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
 
             if ($unblock_sender) {
                 Hm_Msgs::add('Sender Unblocked');
@@ -610,7 +604,6 @@ class Hm_Handler_sieve_block_unblock_script extends Hm_Handler_Module {
             );
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
 
             if (isset($this->request->post['change_behavior'])) {
                 Hm_Msgs::add($scope_title . ' Behavior Changed');
@@ -1032,7 +1025,6 @@ class Hm_Handler_sieve_save_filter extends Hm_Handler_Module {
 
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
             Hm_Msgs::add('Filter saved');
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -1071,7 +1063,6 @@ class Hm_Handler_sieve_save_script extends Hm_Handler_Module {
                 $script_name,
                 $this->request->post['script']
             );
-            $client->close();
             Hm_Msgs::add('Script saved');
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
@@ -1535,7 +1526,6 @@ class Hm_Handler_sieve_toggle_script_state extends Hm_Handler_Module {
             $main_script = generate_main_script($scripts);
             save_main_script($client, $main_script, $scripts);
             $client->activateScript('main_script');
-            $client->close();
 
             Hm_Msgs::add("Script $state");
         } catch (Exception $e) {
@@ -1732,7 +1722,6 @@ class Hm_Handler_load_custom_actions extends Hm_Handler_Module
                     ];
                 }
             }
-            $client->close();
         } catch (Exception $e) {
             Hm_Msgs::add("Sieve: {$e->getMessage()}", "danger");
         }
@@ -1827,7 +1816,6 @@ class Hm_Handler_sieve_remame_folder extends Hm_Handler_Module
                             $script_parsed
                         );
                     }
-                    $client->close();
                     Hm_Msgs::add('Sieve filters using the folder were also updated to use the new folder name.', 'info');
                 } catch (Exception $e) {
                     Hm_Msgs::add("Failed to rename folder in sieve scripts", "warning");
--- a/modules/sievefilters/hm-sieve.php
+++ b/modules/sievefilters/hm-sieve.php
@@ -4,9 +4,16 @@
 require_once APP_PATH.'modules/sievefilters/functions.php';
 
 class Hm_Sieve_Client_Factory {
+    private static $instances = [];
+
     public function init($user_config = null, $imap_account = null, $is_nux_supported = false)
     {
         if ($imap_account && ! empty($imap_account['sieve_config_host'])) {
+            $cache_key = md5($imap_account['sieve_config_host'].'|'.$imap_account['user']);
+            if (isset(self::$instances[$cache_key])) {
+                return self::$instances[$cache_key];
+            }
+
             // Check if module nux is enabled and if it is, get the sieve host from the services
             if($is_nux_supported && $sieve_config = get_sieve_host_from_services($imap_account['server'])) {
                 $sieve_host = $sieve_config['host'];
@@ -17,6 +24,7 @@ class Hm_Sieve_Client_Factory {
             }
             $client = new PhpSieveManager\ManageSieve\Client($sieve_host, $sieve_port);
             $client->connect($imap_account['user'], $imap_account['pass'], $sieve_tls, "", "PLAIN");
+            self::$instances[$cache_key] = $client;
             return $client;
         } else {
             $errorMsg = 'Invalid config host';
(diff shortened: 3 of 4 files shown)


modules/sievefilters/functions.php
```diff
@@ -565,7 +557,45 @@ function get_blocked_senders($mailbox, $mailbox_id, $icon_svg, $icon_block_domai
...
+/* *
+ * This function is used to get the blocked senders script and its content.
+ * It returns an array with the list of scripts and the current script content.
+ */
+if (!hm_exists('get_all_scripts')) {
+    function get_all_scripts($imapServer, $load_current = true, $return_only = null) {
+        try {
+            $client  = SieveConnectionPool::get($imapServer);
+            if(!is_null($client)){
+                $scripts = $client->listScripts();
```
Shadow243: @kroky  `listScripts` doesn’t usually take too much time; should we also add it to the cache?

Shadow243:
@kroky this is now working with Postale. Could you please review it when you have a chance?

modules/sievefilters/functions.php
kroky: Yes, please. The more connection requests we spare, the better.

modules/sievefilters/hm-sieve-pool.php
```diff
@@ -0,0 +1,144 @@
+<?php
+ini_set('memory_limit', '2048M');
+
+use PhpSieveManager\ManageSieve\Client;
+class SieveConnectionPool
```
kroky: Why do you call it a connection pool when there is not a pool of connections reused between requests but rather a cached script data stored in the cache... maybe you should call it something like SieveCache?

modules/sievefilters/hm-sieve-pool.php
```diff
@@ -0,0 +1,144 @@
+<?php
+ini_set('memory_limit', '2048M');
```
kroky: Not sure if you did this for debugging or something but we can't go in production with such a setting.

modules/sievefilters/hm-sieve-pool.php
```diff
@@ -0,0 +1,144 @@
...
+               (time() - self::$lastConnectionTimes[$key]) < self::$timeout;
+    }
+
+    /**
+     * Establish the Sieve connection
+     */
+    private static function connectServer(array $serverConfig)
+    {
+        $client = new Client($serverConfig['host'], $serverConfig['port']);
+        $client->connect(
```
kroky: This caching class shouldn't have anything to do with the connection. Its logic should be encapsulating the sieve data cache - is it fresh enough to use, caching keys, the data that it chooses to cache and return false if no data is cached. Another class must deal with actual connection and fetching sieve data.

modules/sievefilters/hm-sieve-pool.php
Shadow243: It was just for debugging, but now that I’ve moved on to the Cypht cache class, I’ll remove it.

modules/sievefilters/hm-sieve-pool.php
```diff
@@ -0,0 +1,144 @@
...
+        if (!isset(self::$config[$key])) {
+            throw new Exception("No configuration found for '$key'");
+        }
+
+        if (!isset(self::$connections[$key]) || !self::isAlive($key)) {
+            self::$connections[$key] = self::connectServer(self::$config[$key]);
+            self::$lastConnectionTimes[$key] = time();
+        }
+
+        return self::$connections[$key];
```
kroky: If you want to implement a connection reuse for a single web request, you can just add a global connection singleton or facade that returns an existing connection if available, otherwise, it starts a new connection and stores it globally, so it is accessible for code executed down the road.

If you want to implement a connection pool that can be shared between requests that's a totally different story and the approach should be different. I am not sure we can cache/serialize an active connection, so it can be reused in a later ajax or normal request - you can test that but if it is possible, then you can actually implement a connection pool, spin connections when necessary and return unserialized active connections when available.

modules/sievefilters/functions.php
Shadow243: Done

modules/sievefilters/hm-sieve-pool.php
Shadow243: Updated

modules/sievefilters/hm-sieve-pool.php
Shadow243: I removed

modules/sievefilters/hm-sieve-pool.php
Shadow243: @kroky I’ve added some changes, please take a look.

I’ll also add PHPUnit tests for all those classes once you validate.

As for your comment: implementing a pool would indeed require an additional service, which is why I went with the first option (connection reuse within a single request) in my implementation.

modules/sievefilters/modules.php
```diff
@@ -66,9 +74,17 @@ public function process() {
...
             try {
-                $client = $factory->init($this->user_config, $server, $this->module_is_supported('nux'));
+                // Initialize SieveService with cache and server configurations
+                $servers = [];
+                foreach ($this->user_config->get('imap_servers') as $id => $s) {
+                    if (!empty($s['sieve_config_host'])) {
+                        $servers[$id] = $s;
+                    }
+                }
+                SieveService::init($this->cache, $servers);
```
kroky: You can init once in a module and reuse the module whenever you need it.

modules/sievefilters/README.md
```diff
@@ -37,6 +37,114 @@ if address :is "from" "${boss_email}" {
...
+
+- **"caching class shouldn't have anything to do with the connection"**  
+  → `SieveScriptCache` only handles cache logic
+- **"encapsulating the sieve data cache"**  
+  → Pure cache operations with clear responsibilities  
+- **"return false if no data is cached"**  
+  → `getCachedScript()` returns `false` when no cache
+- **"Another class must deal with actual connection"**  
+  → `SieveConnectionManager` handles all connections
+
```
kroky: I'd rather get rid of this long AI-generated readme - it should all be self-sufficient to read the code or comments in the relevant sievefilters module classes.

modules/sievefilters/hm-sieve-pool.php
kroky: OK, added just a couple of comments. Please proceed.

Shadow243:
@christer77 @Baraka24 @IrAlfred I've implemented some enhancements on the block_list page. Please take some time to test them before @kroky proceeds with the review.

Baraka24:
Thanks @shadow243 for this big work, we took the time to test differently and here are what we fund:
* when at ?page=block_list, and connecting with `IMAP SESSION_TYPE`(.env configured IMAP to log in), we were getting :
```
Sieve: Failed to get Sieve scripts for server '68b9aa55b0378': Socket creation failed: php_network_getaddresses: getaddrinfo for tls failed: Temporary failure in name resolution
```
We expected that this should be coming from tls config along side `IMAP_AUTH_TLS`, then we tried to connect using `DB SESSION_TYPE`, and here trying to login(no account configured) we got:
```
- Uncaught TypeError: array_filter(): Argument #1 ($array) must be of type array, false given in /var/www/html/cypht/modules/sievefilters/functions.php:19
Stack trace:
#0 /var/www/html/cypht/modules/sievefilters/functions.php(19): array_filter()
#1 /var/www/html/cypht/modules/sievefilters/modules.php(64): ensure_sieve_service_initialized()
#2 /var/www/html/cypht/lib/modules_exec.php(177): Hm_Handler_sieve_filters_enabled->process()
#3 /var/www/html/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()
#4 /var/www/html/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules()
#5 /var/www/html/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
#6 /var/www/html/cypht/index.php(55): Hm_Dispatch->__construct()
#7 {main}
  thrown in /var/www/html/cypht/modules/sievefilters/functions.php on line 19
```
* note also, only unblock sender action is working at ?page=block_list, other ones are not working.

Baraka24:
One more thing to note, please deal with caches, it is likely the updated block_list content are cached, to see updates we  have to clear all caches, logout and login again the find the updates as well. Please check also as this a part of performance.

Shadow243:
> Thanks @Shadow243 for this big work, we took the time to test differently and here are what we fund:
> 
> * when at ?page=block_list, and connecting with `IMAP SESSION_TYPE`(.env configured IMAP to log in), we were getting :
> 
> ```
> Sieve: Failed to get Sieve scripts for server '68b9aa55b0378': Socket creation failed: php_network_getaddresses: getaddrinfo for tls failed: Temporary failure in name resolution
> ```
> 
> We expected that this should be coming from tls config along side `IMAP_AUTH_TLS`, then we tried to connect using `DB SESSION_TYPE`, and here trying to login(no account configured) we got:
> 
> ```
> - Uncaught TypeError: array_filter(): Argument #1 ($array) must be of type array, false given in /var/www/html/cypht/modules/sievefilters/functions.php:19
> Stack trace:
> #0 /var/www/html/cypht/modules/sievefilters/functions.php(19): array_filter()
> #1 /var/www/html/cypht/modules/sievefilters/modules.php(64): ensure_sieve_service_initialized()
> #2 /var/www/html/cypht/lib/modules_exec.php(177): Hm_Handler_sieve_filters_enabled->process()
> #3 /var/www/html/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()
> #4 /var/www/html/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules()
> #5 /var/www/html/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
> #6 /var/www/html/cypht/index.php(55): Hm_Dispatch->__construct()
> #7 {main}
>   thrown in /var/www/html/cypht/modules/sievefilters/functions.php on line 19
> ```
> 
> * note also, only unblock sender action is working at ?page=block_list, other ones are not working.

Fixed https://github.com/cypht-org/cypht/pull/1604/commits/7a0bea37adab982ae5d413503cbebfde59daee0b

marclaporte:
@Shadow243 This branch has conflicts that must be resolved

Shadow243:
> Thanks @Shadow243 for this big work, we took the time to test differently and here are what we fund:
> 
> * when at ?page=block_list, and connecting with `IMAP SESSION_TYPE`(.env configured IMAP to log in), we were getting :
> 
> ```
> Sieve: Failed to get Sieve scripts for server '68b9aa55b0378': Socket creation failed: php_network_getaddresses: getaddrinfo for tls failed: Temporary failure in name resolution
> ```
> 
> We expected that this should be coming from tls config along side `IMAP_AUTH_TLS`, then we tried to connect using `DB SESSION_TYPE`, and here trying to login(no account configured) we got:
> 
> ```
> - Uncaught TypeError: array_filter(): Argument #1 ($array) must be of type array, false given in /var/www/html/cypht/modules/sievefilters/functions.php:19
> Stack trace:
> #0 /var/www/html/cypht/modules/sievefilters/functions.php(19): array_filter()
> #1 /var/www/html/cypht/modules/sievefilters/modules.php(64): ensure_sieve_service_initialized()
> #2 /var/www/html/cypht/lib/modules_exec.php(177): Hm_Handler_sieve_filters_enabled->process()
> #3 /var/www/html/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()
> #4 /var/www/html/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules()
> #5 /var/www/html/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
> #6 /var/www/html/cypht/index.php(55): Hm_Dispatch->__construct()
> #7 {main}
>   thrown in /var/www/html/cypht/modules/sievefilters/functions.php on line 19
> ```
> 
> * note also, only unblock sender action is working at ?page=block_list, other ones are not working.

@Baraka24 everything should work now.

Baraka24:
@Shadow243 Thanks. I tried and it worked.
To @kroky to rerview

modules/sievefilters/hm-sieve-script-cache.php
```diff
@@ -0,0 +1,499 @@
...
+    {
+        return isset(self::$connections[$key]) && self::isConnectionAlive($key);
+    }
+}
+
+/**
+ * Service orchestrator - combines cache and connection management
+ * This provides backward compatibility and clean API
+ */
+class SieveService
```
kroky: You combine several very different classes in the same file which is quite confusing.

Also, the way we now use phpsievemanager is quite from obvious. It was purposefully designed as an external repo to handle sieve-related connection and interaction. Now we add so much sieve-related connection functionality directly in Cypht. I think we should add all that caching and changes to the php-sieve-manager repo and then keep cypht lightweight - just reuse.

Also, I don't like the fact that we got rid of the client factory - that is reused in Tiki and overridden there, so we will have even more changes to do.

In short, something simple to add like caching resulting in too much boilerplate here. I think we should greatly simplify the current approach. Add simple caching in php-sieve-manager. Bring back the factory and client usage. Add global modules that run on every sieve request rather than calls to methods like ensure_sieve_service_initialized in random list of modules. I don't even think we should have that method - the factory and client usage should be enough - it initializes once, keep cached version, return cached version when requested. No sieve module methods need changes in this case.

Shadow243: Noted, I will push required changes.

Shadow243: sieve PR: https://github.com/cypht-org/php-sieve-manager/pull/26

modules/sievefilters/modules.php
```diff
@@ -22,26 +19,29 @@ class Hm_Handler_sieve_edit_filter extends Hm_Handler_Module {
...
+        
+        if (!$imap_account) {
+            Hm_Msgs::add("Sieve: IMAP account not found", "danger");
+            return;
+        }
 
-        $factory = get_sieve_client_factory($this->config);
         try {
-            $client = $factory->init($this->user_config, $imap_account, $this->module_is_supported('nux'));
+            $client = get_sieve_client($imap_account, $this->config, $this->cache);
```
kroky: There seem to be too many changes in this MR not related to the purpose of reusing connections. Why do we get rid of the client factory? It was reused in Tiki and is a good pattern. Sometimes the client factory returns one class, sometimes another one. You could achieve the same result by modifying the client factory implementation is Cypht and reusing the connections via the cache. Why all these changes in the modules? I'd like to see a severely trimmed down version of a diff here just introducing the caching of connections change... note that you can cache the result of a client factory init and return the same client no matter how many times it is called...

PR #1607, opened 2025-08-11
fix(backend): add .htaccess security rules, root redirect, and sample…

Shadow243:
Added .htaccess rules to block access to .env and other sensitive files, disable directory listing, protect .git, redirect root to index.php, and allow specific sample CSV/YAML files to be publicly accessible.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,43 @@
+# ----------------------------------------------------
+# Rewrite rules (only if mod_rewrite is enabled)
+# ----------------------------------------------------
+<IfModule mod_rewrite.c>
+    RewriteEngine On
+
+    RewriteCond %{REQUEST_URI} ^/$
+    RewriteRule ^$ index.php [L]
+
+    # Force HTTPS (uncomment to enable)
+    # RewriteCond %{HTTPS} !=on
+    # RewriteRule ^ https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
+</IfModule>
+
+# ----------------------------------------------------
+# Security Rules
+# ----------------------------------------------------
+
+# Block hidden files (starting with .)
+<FilesMatch "^\.">
+    Require all denied
+</FilesMatch>
+
+# Block sensitive project files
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -0,0 +1,37 @@
+RewriteEngine On
```
kroky: It would be nice to check if mod rewrite is available first.

kroky:
What do we do with nginx setups? We should add documentation to help users secure their deployments.

Also, can you check our stock docker images/files? Is this htaccess working out of the box? Apache must be used and override directives enabled.

Shadow243:
> What do we do with nginx setups? We should add documentation to help users secure their deployments.
> 
I'm adding a step to `https://www.cypht.org/install/` with this for nginx:
```
location = / {
    rewrite ^/$ /index.php last;
}

# Block hidden files starting with .
location ~ /\. {
    deny all;
}

# Block sensitive files
location ~* \.(env|ini|log|conf|json|lock|yml|yaml|md|txt|sh|bat|ps1|xml|bak|sql|dist|inc|cfg|db|csv)$ {
    deny all;
}

# Allow exceptions for specific files
location ~* ^/(server_accounts_sample\.yaml|server_accounts_sample\.csv|contact_sample\.csv)$ {
    allow all;
}

# Block RELEASE_NOTES, Makefile, Docker-related configs
location ~* /(RELEASE_NOTES|Makefile|Dockerfile|docker-compose\.yml|docker-compose\.dev\.yaml|docker-compose\.prod\.yaml)$ {
    deny all;
}

# Block .git directory
location ~ /\.git {
    deny all;
}

# Disable directory listing
autoindex off;
```

Shadow243:
@kroky We are using Nginx: `https://github.com/cypht-org/cypht/blob/master/docker/nginx.conf` as the web server inside our Docker setup, not Apache.

According to this line in the entrypoint script:
https://github.com/cypht-org/cypht/blob/master/docker/docker-entrypoint.sh#L32
the web root is set by linking /var/www to `${APP_DIR}/site`.

This means the application serves files from the site directory directly, just like in production. ensuring all files are properly secured without relying on .htaccess files.

PR #1610, opened 2025-08-13
refactor (backend) : Improve error messages to be more usefull

marcellintacite:
## 🍰 Pullrequest
This pull request improves the clarity and usefulness of error messages in the SMTP module. The new messages are more user-friendly and provide actionable suggestions, making it easier for non-technical users to understand and resolve issues.

These changes make Cypht more accessible to all users, reducing confusion and support requests related to unclear error messages.

### Issues

- relates : https://avan.tech/item63816

### Todo

- [X] None

### DIFF
--- a/modules/smtp/hm-smtp.php
+++ b/modules/smtp/hm-smtp.php
@@ -264,7 +264,7 @@ class Hm_SMTP {
     function connect() {
         $certfile = false;
         $certpass = false;
-        $result = 'An error occurred connecting to the SMTP server';
+        $result = "We couldn't connect to the email server (SMTP). Please check your internet connection or server settings, and try again.";
         $server = $this->server;
 
         if ($this->tls) {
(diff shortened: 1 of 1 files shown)


modules/smtp/hm-smtp.php
```diff
@@ -283,7 +283,7 @@ function connect() {
         else {
             $this->debug[] = 'Could not connect to the SMTP server';
             $this->debug[] = 'fsockopen errors #'.$errorno.'. '.$errorstr;
-            $result = 'Could not connect to the configured SMTP server';
+            $result = "Unable to connect to the email server. Please check your internet connection or server settings, and try again.";
```
kroky: Please try to keep the details it previously provided. It will be useful to know it is the SMTP server here and not an IMAP server, for example.

modules/smtp/hm-smtp.php
```diff
@@ -610,12 +610,12 @@ function send_message($from, $recipients, $message, $from_params = '', $recipien
                     $result = false;
                 }
                 else {
-                    $result = 'An error occurred sending the message DATA';
+                    $result = "Your message could not be sent. The server did not confirm delivery. Please try again later.";
```
kroky: We should log the technical error somewhere - even the php error logs should be enough. This will be useful when debugging the connection problems later.

marcellintacite: Thank you @kroky for the feedback! I've implemented both suggestions in the updated code. First, I've preserved technical specificity by ensuring all error messages include relevant command names (DATA, RCPT, EHLO, STARTTLS) to distinguish from other protocols like IMAP. Second, I've added comprehensive error_log() call.

PR #1613, opened 2025-08-15
fix(frontend): bring back missing 'prev' and 'next' links on the message page and restore interactivity of the move and copy message buttons

mercihabam:

PR #1614, opened 2025-08-18
fix(frontend): copy or move action that is no longer working

Shadow243:
fix copy and move click issue
CleanShot 2025-08-18 at 05 12 03@2x


Shadow243:
Will be fixed by @mercihabam in [https://github.com/cypht-org/cypht/pull/1613](https://github.com/cypht-org/cypht/pull/1613)

PR #1615, opened 2025-08-18
fix(frontend): information about the event no longer appears in tiki-webmail

christer77:
Before:
before

After:
after

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -352,6 +352,7 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
             }
 
             // Action links section
+            $txt .= '<div class="event_calendar_section"></div>';
             $txt .= '<div class="row g-0 py-3">';
             $txt .= '<div class="col-12 msg_actions">';
             $txt .= '<div class="d-flex flex-wrap gap-2 mb-3">';


modules/imap/output_modules.php
```diff
@@ -352,6 +352,7 @@ protected function output() {
             }
 
             // Action links section
+            $txt .= '<div class="tiki_event_calendar_section"></div>';
```
kroky: This should go in Tiki codebase. Nothing in Cypht relates to Tiki. We have all the custom code for the integration in the Tiki repo.

christer77: Yes, you are right, it is also done in tiki. 

But this part will be replaced by the tiki code.

[See MR.](https://gitlab.com/tikiwiki/tiki/-/merge_requests/8317)

kroky: We don't want references to Tiki in Cypht codebase. It is a standalone client, after all. Please use another approach - target existing header class, existing dom structure or as a last resort - you can use a placeholder but name it in a generic way - e.g. 'event_calendar_section'.

christer77: I think I can use the latest approach(event_calendar_section). 

Thank you very much.

marclaporte:
Corresponding change in Tiki: https://gitlab.com/tikiwiki/tiki/-/merge_requests/8317

PR #1616, opened 2025-08-18
fix(frontend): submenu display issue

Shadow243:
CleanShot 2025-08-19 at 02 22 00@2x

### DIFF
--- a/modules/imap/site.css
+++ b/modules/imap/site.css
@@ -603,8 +603,8 @@
     opacity: 1;
 }
 .inner_list li {
-  display: flex !important;
-  flex-direction: row !important;
+  display: flex;
+  flex-direction: row;
   gap: 2px;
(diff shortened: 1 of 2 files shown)


modules/imap/site.css
```diff
@@ -614,4 +614,20 @@
     left: 33.5%;
     top: 50%;
     transform: translate(-50%, -50%);
-}
\ No newline at end of file
+}
+
+.inner_list li[class^="m-has-children"] {
+    display: flex !important;
```
kroky: There should be another way to specify these styles. I prefer we keep the stock cypht css files free of !important overrides, so theme developers and integrations like Tiki-Cypht can override more easily.

Shadow243: @kroky I've updated

PR #1622, opened 2025-08-21
fix(backend): delete profile after imap or smtp server removed to avoid orphaned profiles

Shadow243:
Removal of SMTP/IMAP servers results in orphaned profiles.

### DIFF
--- a/modules/profiles/modules.php
+++ b/modules/profiles/modules.php
@@ -59,6 +59,44 @@ class Hm_Handler_compose_profile_data extends Hm_Handler_Module {
 /**
  * @subpackage profile/handler
  */
+class Hm_Handler_process_smtp_server_data_delete extends Hm_Handler_Module {
+    public function process() {
+        if (isset($this->request->post['smtp_delete'])) {
+            $deleted_server_id = $this->get('deleted_server_id', '');
+            Hm_Profiles::init($this);
+            $servers = Hm_Profiles::getBy($deleted_server_id,'smtp_id');
+            if ($servers) {
+                foreach ($servers as $server) {
+                    Hm_Profiles::del($server['id']);
+                    Hm_Msgs::add("Profile {$server['name']} has been deleted since its server was removed.");
+                }
+            }
+        }
+    }
+}
+
+/**
+ * @subpackage profile/handler
+ */
+class Hm_Handler_process_imap_server_data_delete extends Hm_Handler_Module {
+    public function process() {
+        if (isset($this->request->post['imap_delete'])) {
+            $deleted_server_id = $this->get('deleted_server_id', '');
+            Hm_Profiles::init($this);
+            $servers = Hm_Profiles::getBy($deleted_server_id,'imap_id');
+            if ($servers) {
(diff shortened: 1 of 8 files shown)


marclaporte:
Good one. This happened to me before.

modules/profiles/modules.php
```diff
@@ -56,6 +56,40 @@ public function process() {
...
 
+/**
+ * @subpackage profile/handler
+ */
+class Hm_Handler_process_smtp_server_data_delete extends Hm_Handler_Module {
+    public function process() {
+        if (isset($this->request->post['smtp_delete'])) {
+            $deleted_server_id = $this->get('deleted_server_id', '');
+            Hm_Profiles::init($this);
+            $server = Hm_Profiles::getBy($deleted_server_id,'smtp_id', true);
```
kroky: What if you have more than one profiles matching these servers? I think you should loop through the results.

modules/profiles/setup.php
```diff
@@ -24,6 +24,9 @@
 add_handler('ajax_smtp_save_draft', 'compose_profile_data', true, 'profiles', 'load_smtp_servers_from_config', 'after');
 add_handler('ajax_smtp_attach_file', 'compose_profile_data', true, 'profiles', 'load_smtp_servers_from_config', 'after');
 add_handler('servers', 'compose_profile_data', true, 'profiles', 'load_smtp_servers_from_config', 'after');
+add_handler('ajax_smtp_debug', 'process_smtp_server_data_delete', true);
```
kroky: Don't we need something more specific here - execute after the exact action that removes the smtp (or imap) server?

marclaporte:
Related: https://github.com/cypht-org/cypht/pull/1517

modules/profiles/setup.php
Shadow243: I referred to this: https://github.com/cypht-org/cypht/blob/master/modules/smtp/setup.php#L56
 and https://github.com/cypht-org/cypht/blob/master/modules/imap/setup.php#L135

modules/profiles/modules.php
Shadow243: I just updated

modules/profiles/setup.php
kroky: Yes, I still prefer you execute the profile delete exactly after imap/smtp delete, so we are sure further caching or other handler modules don't stand in the way...

Shadow243: @kroky I've updated.

modules/profiles/modules.php
```diff
@@ -56,6 +56,45 @@ public function process() {
     }
 }
 
+/**
+ * @subpackage profile/handler
+ */
+class Hm_Handler_process_smtp_server_data_delete extends Hm_Handler_Module {
+    public function process() {
+        if (isset($this->request->post['smtp_delete'])) {
+            exit(var_dump($this->request->post));
```
kroky: exit?

Shadow243: Updated

PR #1623, opened 2025-08-21
fix(frontend): prevent archiving already archived messages

christer77:
**Issue**:
The messages already present in the 'Archive' folder could be archived again, which made no functional sense and created confusion for users.

archiver-un-message-contenu-d-archive

Solution: disabling the 'Archive' button/option in this context.

### DIFF
--- a/modules/core/hm-mailbox.php
+++ b/modules/core/hm-mailbox.php
@@ -633,7 +633,7 @@ class Hm_Mailbox {
 
     public function is_archive_folder($id, $user_config, $current_folder) {
         if ($this->is_imap()) {
-            return $this->connection->is_archive_folder($id, $user_config, $current_folder);
+            return is_imap_archive_folder($id, $user_config, $current_folder);
         }
         return false;
     }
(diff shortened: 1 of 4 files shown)


marclaporte:
Good catch!

modules/imap/output_modules.php
```diff
@@ -383,7 +383,11 @@ protected function output() {
             $txt .= '<a class="delete_link hlink text-decoration-none" id="delete_message" href="#">'.$this->trans('Delete').'</a>';
             $txt .= '<div class="position-relative"><a class="hlink" id="copy_message" href="#">'.$this->trans('Copy').'</a><div class="move_to_location"></div></div>';
             $txt .= '<div class="position-relative"><a class="hlink" id="move_message" href="#">'.$this->trans('Move').'</a><div class="move_to_location"></div></div>';
-            $txt .= '<a class="archive_link hlink text-decoration-none" id="archive_message" href="#">'.$this->trans('Archive').'</a>';
+                        
+            $current_folder = end(explode("_", $this->html_safe($this->get('msg_list_path'))));
+            if (bin2hex('Archive') != $current_folder) {
```
kroky: Archive might be the one or might not - we have a folder selection mechanism where people choose which folder to act as junk, sent, archive, etc. Furthermore, EWS tends to handle these folder comparisons a bit differently. Please move the check for archive folder in the relevant implementations (hm-mailbox bridge and then passing to hm-imap, hm-ews, etc. where they know the best way to get/compare the Archive folder).
Also, please see if the current folder is not already passed as a variable from handlers. If not, pass it. Getting it here from list_path can lead to problems in the future (e.g. changing list path) and is not exactly the things output modules should do.

modules/core/hm-mailbox.php
```diff
@@ -631,6 +631,13 @@ public function select_folder($folder) {
         return true;
     }
 
+    public function is_archive_folder($id, $user_config, $current_folder) {
+        if ($this->is_imap()) {
+            return $this->connection->is_archive_folder($id, $user_config, $current_folder);
```
kroky: This should be retrieved by the special folder settings... I think you do it correctly in hm-imap, just here it doesn't make sense - is it actually defined for hm-imap?

kroky: Sorry, I meant it is OK in hm-imap but not the right place - probably worth adding it as a module in imap handler modules and then using it where you want. Or maybe in the imap_folders modules as that's where special folders are mostly used. We want to keep the code contextually close to other meaningful code.

christer77: > Sorry, I meant it is OK in hm-imap but not the right place - probably worth adding it as a module in imap handler modules and then using it where you want. Or maybe in the imap_folders modules as that's where special folders are mostly used. We want to keep the code contextually close to other meaningful code.

Resolved

PR #1626, opened 2025-08-23
fix(frontend): send correct page to Hm_MessagesStore while refreshing…

Shadow243:
fix issue: [https://avan.tech](https://avan.tech/item140411)


kroky:
Nice catch, thanks!

PR #1630, opened 2025-08-24
refactor(backend): Drop legacy versions and update dependenciesfor PHP 8.1+ support

bienvenuushindi:
We do not need to keep compatibility constraints for PHP ≤ 8.01 anymore, because the project explicitly excludes those versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "docs": "https://cypht.org/documentation.html"
     },
     "require": {
-        "bacon/bacon-qr-code": "^1.0.3 || ^2.0.0 || ^3.0.0",
+        "bacon/bacon-qr-code": "^3.0.0",
         "christian-riesen/base32": "^1.3.2",
         "composer": "^2.0.0",
         "ext-curl": "*",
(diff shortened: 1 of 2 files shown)


kroky:
@bienvenuushindi please resolve the conflict. Please also test with the new version of the packages.

bienvenuushindi:
> @bienvenuushindi Please also test with the new version of the packages.


I verified the package, and the newer version does not affect the existing implementation in our codebase

marclaporte:
I use https://dev.tiki.org/How-to-upgrade-Cypht-within-Tiki-via-Composer to update Tiki master to use Cypht master. I am now getting:

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires jason-munro/cypht dev-master -> satisfiable by jason-munro/cypht[dev-master].
    - jason-munro/cypht dev-master requires symfony/dotenv ^6.4 -> found symfony/dotenv[v6.4.0-RC1, ..., v6.4.24] but the package is fixed to v5.4.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
```

bienvenuushindi:
Updated composer.lock 

 https://gitlab.com/tikiwiki/tiki/-/merge_requests/8435

marclaporte:
I confirm my issue is solved. Thank you!

PR #1633, opened 2025-08-25
fix(frontend): ensure background message list refresh requests use the current list_page query

mercihabam:


kroky:
As much as I know, we can only have one table of messages per page view. Why do we need an id? Is there a case when multiple tables of messages are displayed in the same page/view?

mercihabam:
We indeed have only one table of messages per page view, and in no case are there multiple tables of messages displayed. The need for an id arises when refreshing the principal message list in the background while the current page differs from it. So by using a unique id per store, we prevent blind updates of the view by the background request.

kroky:
Isn't it better to cancel any background requests that shouldn't update the current view? I think select_imap_folder is doing it in the internal method that it clears the interval upon navigation. What other background jobs are updating the view? In theory, we shouldn't run background jobs for a view that is not instantiated (or currently visible) - it is a waste of resources.

mercihabam:
The background job that isn’t cancelled is the one for the current page query, which has been paginated to a different list_page number. That said, I see your point—there’s no need to fetch a non-displayed list in the background. I’ll work on finding a way to cancel such requests instead.

mercihabam:
It should be working properly now.

PR #1634, opened 2025-08-25
fix(frontend): js selectors-related regressions

mercihabam:


modules/inline_message/site.js
```diff
@@ -128,8 +128,8 @@ var capture_subject_click = function() {
         var uid = msg_details[0];
         var list_path = msg_details[1];
         var inline_msg_loaded_callback = function() {
-            if ($('.header_subject th').find('i.bi.bi-x-lg.close_inline_msg').length === 0) {
-                $('.header_subject th').append('<i class="bi bi-x-lg close_inline_msg"></i>');
+            if ($('.js-header_subject th').find('i.bi.bi-x-lg.close_inline_msg').length === 0) {
```
kroky: Is it still a th? Didn't we change to div layout the header section? While here, can you please check other modules like github and wordpress to see if they are in line with what has been done in imap and core modules regarding message header refactoring?

Also, I saw a couple of header_subject selector usages in selenium tests - can you check that as well?

mercihabam: Absolutely right, there should be no more th in the selector. I removed it, and checked a few other places; I found only in the feeds module where `.header_subject` is still necessary.

kroky: OK, I am merging this but please also check github and wordpress modules - I see some table headers there - are they in sync with the rest? Can you or @Shadow243 test these modules?

kroky: Actually, I see several MRs about that, so it should be done now.

Issue #1624, opened 2025-08-21
🌟 [EPIC] release Cypht 2.5.0


marclaporte:
## 🌟 EPIC

I know I have been talking about the release of Cypht 2.5.0 for many months

It has been 11 months since 2.4.0: https://github.com/cypht-org/cypht/releases/tag/v2.4.0

https://openhub.net/p/cypht/commits/summary reports 656 commits by 33 contributors modifying 349 files in the last 12 months. So activity level is very high (given the overall size of the project) but 11 months is way too long. It should be 2-3 months between stable releases. For at least 6 months, I thought we could release 2.5.0 within 4-5 weeks, but I was mistaken. There were bugs and yet more bugs to fix. But now, I think we really are 4-5 weeks away. I sure hope to be right this time.

Since 2.4.x, Cypht master has undergone refactoring of major parts by two devs (Merci Jacob and Victor Emanouilov). These efforts are to build a better long term future but they cause side effects, and many small subsequent adjustments.

We have done significant work on Tiki-Cypht interoperability: https://doc.tiki.org/Cypht
So how to evolve Cypht as a standalone app while making it easier to integrate within Tiki, and eventually other apps. For example, I could imagine one day integrating Cypht within Dolibarr.

Beyond Merci and Victor (mentioned above), thank you to Steven, Josaphat, Baraka and Christer for the many contributions in the last 12 months.

https://github.com/cypht-org/cypht/graphs/contributors?from=8%2F17%2F2024

https://openhub.net/p/cypht/contributors?query=&time_span=&sort=twelve_month_commits

So now, we need Cypht master to be at least as stable as 1.4.x. Then
- we spin out 2.5.0 (from master)
- use Cypht 2.5.x for Tiki 29.x, 28.x and 27.x LTS (so will need to align a lot of code in Tiki)

From then on, we get back to stable releases every few months (2.6.0, 2.7.0, etc.)

Somewhere in 2026 (Q3 or Q4), we for Cypht 3.0

marclaporte:
I created a 2.5 milestone and added some blockers and notable issues and PRs:
https://github.com/cypht-org/cypht/milestone/6

marclaporte:
I think we have solved all the blockers: https://github.com/cypht-org/cypht/milestone/6

The big blocker was: https://github.com/cypht-org/cypht/pull/1635

So now, let's polish up!

marclaporte:
[...]

Issue #280, opened 2018-08-21
Notices on mobile are off screen (over the folder list)


jasonmunro:
System notices are not aligned correctly on mobile when viewing the main content area, instead they are showing up above the hidden folder list and only visible when accessing the list.

jasonmunro:
turned out a little z-index and position style tweaking did the trick! :)

PR #1635, opened 2025-08-25
fix(backend): fix EWS issues and enhance interoperability

kambereBr:
This PR fixes EWS issues and improves interoperability with Cypht.

~Temporary patch until https://github.com/Garethp/php-ews/pull/275 is merged in.~ (merged)


composer.json
```diff
@@ -56,11 +56,15 @@
         "twbs/bootstrap": "^5.3",
         "twbs/bootstrap-icons": "^1.11",
         "webklex/composer-info": "^0.0.1",
-        "zbateson/mail-mime-parser": "^3.0"
+        "zbateson/mail-mime-parser": "^3.0",
+        "cweagans/composer-patches": "^1.7"
```
marclaporte: Please keep lists like this alphabetically sorted.

kambereBr: Sure, thank you

modules/imap/hm-ews.php
```diff
@@ -269,7 +269,29 @@ public function send_message($from, $recipients, $message, $delivery_receipt = f
             $msg = new Type\MessageType();
             $msg->setFrom($from);
             $msg->setToRecipients($recipients);
-            $msg->setMimeContent(base64_encode($message));
+            $parser = new MailMimeParser();
```
kroky: This would omit some message parts if there is more than html and plain text parts. It will also use either the html or plain text one but not both. I don't really want to parse here and rebuild the mime message. We should have already received a well formatted mime message that smtp module builds. Why can't we use it as a raw string to send the message?

modules/imap/functions.php
```diff
@@ -261,18 +261,21 @@ function format_imap_message_list($msg_list, $output_module, $parent_list=false,
...
+            $timestamp = strtotime($convertedDate);
         } else {
             if ($list_sort == 'date') {
                 $date_field = 'date';
             } else {
                 $date_field = 'internal_date';
             }
-            $date = translate_time_str(human_readable_interval($msg[$date_field]), $output_module);
-            $timestamp = strtotime($msg[$date_field]);
+            $convertedDate = $msg[$date_field] instanceof DateTime ? $msg[$date_field]->format('Y-m-d H:i:s') : $msg[$date_field];
```
kroky: Please standardize the format of the dates returned by hm-mailbox no matter what the underlying connection is. I see it used to parse strings to dates, so this code expects strings. Leave it like this to expect strings and instead modify offending code in hm-ews to return strings of dates. Otherwise, you can refactor imap module to return DateTime objects but make sure hm-mailbox always returns the same type of data. I'd go with strings as ews is an add-on to imap and is more rarely used, so we should keep focus on imap mainly.

Shadow243:
@kroky @marclaporte We just finished fixing the display of the mail list on the message_list page and the full message details (including all headers) on the message page.

Next step: test all remaining actions (copy, move, flag/unflag(already tested and working fine), and other actions).

kroky:
@kambereBr @Shadow243 please ping me when this is ready to review again.

Shadow243:
> @kambereBr @Shadow243 please ping me when this is ready to review again.

Copy that @kroky ,

We are almost done here, some small issues remaining

marclaporte:
> fix(backend): apply ews type converter patch
> 
> Temporary patch until [Garethp/php-ews#275](https://github.com/Garethp/php-ews/pull/275) is merged in.

@kambereBr Now that patch has been merged, description could use an update.

kambereBr:
> @kambereBr @Shadow243 please ping me when this is ready to review again.

@kroky could you please take a look?

We ran some tests and most passed, but we found a few issues:

- Editing a draft with an attachment: adding a new attachment and sending the mail doesn’t remove it from Drafts, and it fails when trying to access the new file.

- For the recipient, In Gmail, when selecting **Show Original**, the message content is displayed in base64 instead of plain text.

- With `AUTH_TYPE = DB`, date/time of recent messages may appear incorrectly. For example, composing a mail and seeing a message from 2 minutes ago might display as 3h ago. But With `AUTH_TYPE = IMAP`, the date/time formatting is correct

Since these issues may take some time to resolve, we’re marking this PR as ready for review and will address them in a separate task : https://avan.tech/item141962

modules/imap/handler_modules.php
```diff
@@ -550,18 +550,31 @@ public function process() {
...
+                    
                     $mailbox = Hm_IMAP_List::get_mailbox_without_connection($details);
                     $label = $mailbox->get_folder_name($folder);
-                    $title = array(strtoupper($details['type'] ?? 'IMAP'), $details['name'], $label);
+                    
+                    // For EWS, if we can't get a friendly name, try to use a connected mailbox
+                    if (!$label && isset($details['type']) && $details['type'] === 'ews') {
+                        $connected_mailbox = Hm_IMAP_List::get_connected_mailbox($parts[1], $this->cache);
+                        if ($connected_mailbox && $connected_mailbox->authed()) {
+                            $folder_status = $connected_mailbox->get_folder_status($folder, false);
```
kroky: This is rather slow to get to the server for each folder label. I think I implemented label caching and returning it immediately when requested in the mailbox class. Also, abstract these types of checks in the mailbox class. We shouldn't pollute handler modules with if statements that depend on server type.

modules/imap/hm-ews.php
```diff
@@ -840,6 +860,25 @@ public function get_mime_message_by_id($itemId) {
         return $parser->parse($content, false);
     }
 
+    protected function get_item_id_base64($itemId) {
```
kroky: I understand the desire to use this function but it is masking other problems. If we expect hex internal format, we should receive hex internal format. If we receive base64 instead, there is a problem somewhere else and we shouldn't mask it.

modules/imap/hm-ews.php
```diff
@@ -681,9 +695,11 @@ public function message_action($action, $itemIds, $folder=false, $keyword=false)
     }
 
     public function get_message_headers($itemId) {
+        $base64Id = $this->get_item_id_base64($itemId);
         $request = array(
             'ItemShape' => array(
                 'BaseShape' => 'AllProperties',
+                'IncludeMimeContent' => true,
```
kroky: This will be a huge effort on the server and the network to return the full mime content every time with the headers. Why do you add it here? It should be part of get message content only.

modules/imap/hm-ews.php
```diff
@@ -916,6 +972,61 @@ protected function search_mime_part_in_struct($struct, $conditions, $all = false
...
+            return $description;
+        }
+        
+        return null;
+    }
+    
+    protected function decode_attachment_name($name) {
+        $name = trim($name, '"\'');
+        
+        if (preg_match('/=\?([^?]+)\?([BQ])\?([^?]+)\?=/i', $name, $matches)) {
```
kroky: Where did you get this regex from? There should already be a lib in cypht parsing mime headers or whatever you are trying to do here. It seems rather low-level to do it like that.

modules/imap/hm-ews.php
```diff
@@ -972,28 +1099,40 @@ protected function is_distinguished_folder(&$folder) {
...
-                        return (new Type\ItemIdType($itemId))->toArray();
-                    }, $itemIds),
-                ]
-            ];
-            $request = Type::buildFromArray($request);
+        try {
+            $archiveFolder = null;
             try {
-                $result = $result && $this->ews->ArchiveItem($request);
+                $archiveFolder = $this->api->getFolderByDistinguishedId(Enumeration\DistinguishedFolderIdNameType::ARCHIVE_INBOX);
```
kroky: Why do you change the archiving mechanism? It seems you are doing an artificial archive folder and moving there while archiving in EWS is done with ArchiveItem operation: https://learn.microsoft.com/en-us/exchange/client-developer/web-service-reference/archiveitem-operation?redirectedfrom=MSDN

I think you are changing the semantics of this operation...

kroky:
> > @kambereBr @Shadow243 please ping me when this is ready to review again.
> 
> @kroky could you please take a look?
> 
> We ran some tests and most passed, but we found a few issues:
> 
> * Editing a draft with an attachment: adding a new attachment and sending the mail doesn’t remove it from Drafts, and it fails when trying to access the new file.
> * For the recipient, In Gmail, when selecting **Show Original**, the message content is displayed in base64 instead of plain text.
> * With `AUTH_TYPE = DB`, date/time of recent messages may appear incorrectly. For example, composing a mail and seeing a message from 2 minutes ago might display as 3h ago. But With `AUTH_TYPE = IMAP`, the date/time formatting is correct
> 
> Since these issues may take some time to resolve, we’re marking this PR as ready for review and will address them in a separate task : https://avan.tech/item141962

Since there are already things to fix here, I'd rather put all such EWS updates here and we do one complete review and merge.

modules/imap/hm-ews.php
Shadow243: @kroky we are facing this issue: 
`Uncaught Error: Call to undefined method garethp\ews\API\Message\ArrayOfResponseMessagesType::getArchiveItemResponseMessage() in /cypht/vendor/garethp/php-ews/src/API/ExchangeWebServices.php:402
Stack trace:
#0 /cypht/vendor/garethp/php-ews/src/API/ExchangeWebServices.php(381): garethp\ews\API\ExchangeWebServices::drillDownResponseLevels(Object(garethp\ews\API\Message\ArrayOfResponseMessagesType))
#1 /cypht/vendor/garethp/php-ews/src/API/ExchangeWebServices/MiddlewareFactory.php(57): garethp\ews\API\ExchangeWebServices->processResponse(Object(garethp\ews\API\Message\ArrayOfResponseMessagesType))`

we tried `exit(var_dump($this->ews));`as well but there is no ArchiveItem function.

based on https://learn.microsoft.com/en-us/exchange/client-developer/web-service-reference/archiveitem-operation?redirectedfrom=MSDN it was introduced since `VERSION_2013` but we are not getting it.

Any thoughts ?

kroky: I think the archiving feature is enabled or disabled by the administrator of the Exchange service. It might be disabled for the server you are testing on. We need to check if that's the case and either fall back to manual operation or better - hid the archive option/functionality as a whole. I prefer we keep it the way Exchange was supposed to work and if it is disabled, it must have been done for a reason.

Shadow243: @kroky  We checked with the Exchange server admin and confirmed that there’s no need to explicitly enable Exchange itself, the archive functionality is already available. We then updated the package to implement the feature, tested it, and it’s now working as expected. The fix is included in this PR: https://github.com/Garethp/php-ews/pull/280

Should we apply the patch of it ([#280](https://github.com/Garethp/php-ews/pull/280)) ?

kroky: Let's wait for 281 to be merged upstream, update the lib here and then we can test/review this PR.

kambereBr: @kroky We’ve just added a [comment](https://github.com/Garethp/php-ews/pull/281#issuecomment-3272302999) to PR 281 regarding the Archive Item issue. We’ll wait for feedback there while we continue working on other issues here.

Thanks!

modules/contacts/modules.php
```diff
@@ -104,9 +104,19 @@ public function process() {
...
+                if(is_array($value)) {
+                    foreach ($value as $val) {
+                        foreach (Hm_Address_Field::parse($val) as $v) {
+                            if (!$existing->search(array('email_address' => $v['email']))) {
+                                $addresses[] = $v;
+                            }
+                        }
+                    }
+                }else {
+                    foreach (Hm_Address_Field::parse($value) as $vals) {
```
kroky: The same block is repeated above - please dry it up.

modules/imap/handler_modules.php
```diff
@@ -674,6 +683,11 @@ public function process() {
                     unset($msgs[$folder]);
                 }
                 $this->cache->set('imap_folders_'.$path, $msgs);
+                Hm_EWS::cache_all_folders($form['imap_server_id'], $this->cache, $msgs);
+                // var_dump([
```
kroky: Please remove debugging code.

modules/imap/handler_modules.php
```diff
@@ -1990,6 +2005,7 @@ public function process() {
 
         $mailbox = Hm_IMAP_List::get_connected_mailbox($server_id, $this->cache);
         if ($mailbox && $mailbox->authed()) {
+            $prefetch = true;  // Set prefetch to true for this context
```
kroky: The comment is not very useful - can you explain why prefetch is needed here - in a code comment, please.

modules/imap/hm-ews.php
```diff
@@ -148,22 +150,100 @@ public function get_folder_name_quick($folder) {
         }
     }
 
+    /**
+     * TODO: Remove this one, we needed it only while testing
```
kroky: Can you complete the todo items here?

modules/smtp/modules.php
```diff
@@ -787,10 +787,19 @@ public function process() {
...
-            delete_uploaded_files($this->session, 0);
+        
+        /* Clean up uploaded files
+         * Note: EWS server handles draft deletion differently than IMAP/SMTP.
+         * For EWS, drafts are completely removed during delete_draft operation,
+         * making uploaded file cleanup unnecessary as files are no longer accessible.
+         * For IMAP/SMTP, drafts may remain flagged as deleted but still exist on server,
+         * requiring explicit cleanup of associated uploaded files.
+         */
+        if($mailbox->server_type() !== 'EWS') {
```
kroky: What does draft have to do with files temporarily uploaded to the Cypht server (not EWS or IMAP one)? delete_uploaded_files is doing only unlink operations on the files temporarily stored on the server. Why don't we have to delete them after sending with EWS? I understand EWS handles drafts and temporary sent emails differently but we are dealing with local files stored here only for the process of sending a message and they don't seem to be treated differently by underlying implementations, do they?

modules/imap/hm-ews.php
kroky: It seems other MRs (https://github.com/cypht-org/cypht/pull/1699) are waiting for this one, so let's polish it and merge it now leaving the Archive problem to be solved after upstream is resolved.

marclaporte:
@kambereBr @Shadow243 @kroky 

Thank you!

This was, AFAIK, the last blocker to release 2.5.0!
https://github.com/cypht-org/cypht/issues/1624

PR #1642, opened 2025-08-29
fix(frontend): change msg_actions to look like buttons

pom2ter:
Part of https://avan.tech/item139134

The 'forward' action was a button but not the rest.

### DIFF
--- a/modules/core/site.css
+++ b/modules/core/site.css
@@ -981,13 +981,14 @@ div.unseen,
 .mobile .folder_list {
   margin-right: 0px;
   width: 100%;
+  padding-top: 0px;
 }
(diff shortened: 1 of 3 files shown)


kroky:
Sometimes, changing css or other inline dom elements causes tests to fail. @Shadow243 can you please take a look here and fix before merging?

Shadow243:
> Sometimes, changing css or other inline dom elements causes tests to fail. @Shadow243 can you please take a look here and fix before merging?

Noted, I will test.

marclaporte:
@pom2ter please address merge conflicts

pom2ter:
conflicts should be fixed

Issue #1368, opened 2024-11-19
Move button without functionality in 50% of the cases


ulfgebhardt:
## 🐛 Bugreport

Move button without functionality in 50% of the cases

It happens very often, when selecting an email the move button is correctly shown, but has no functionality. This seems to be click handler register issue.

![image](https://github.com/user-attachments/assets/70654239-af2c-41ac-95be-2f65412ec13c)

### Version & Environment

Rev: [v2.4.0]

OS:  [chromium, alpine]

### Steps to reproduce
1. Mark Mail
2. Click Move Button
3. Nothing happens in ~25-50% of the cases
4. ...
5. Profit

jacob-js:
The move button works perfectly fine. @ulfgebhardt could you please share the details of your device and/or browser?

ulfgebhardt:
![image](https://github.com/user-attachments/assets/167a03b0-cefb-429f-a3a7-57fdc4c00445)

ulfgebhardt:
Further debug info:

If this happens the classes of the move button (an `<a>`) do not change.

1. No message selected (invisible): `imap_move btn btn-sm btn-secondary no_mobile disabled_input`
2. Selecting a message (clickable): `imap_move btn btn-sm btn-secondary no_mobile`
3. Deselecting the message (invisible): `imap_move btn btn-sm btn-secondary no_mobile disabled_input`

Error case:
1. No message selected (invisible): `imap_move btn btn-sm btn-secondary no_mobile disabled_input`
2. Selecting a message (unclickable): `imap_move btn btn-sm btn-secondary no_mobile disabled_input`
3. Deselecting the message (invisible): `imap_move btn btn-sm btn-secondary no_mobile disabled_input`

Removing the class `disabled_input` manually does not make the button clickable

jacob-js:
Let's try to solve this. Please share with me the value of the id attribute for the checked box of a message.

ulfgebhardt:
``` 
<td class="checkbox_cell"><input id="imap_6738e4f407104_5130_494e424f58" type="checkbox" value="imap_6738e4f407104_5130_494e424f58"><label class="checkbox_label" for="imap_6738e4f407104_5130_494e424f58"></label></td>
```

jacob-js:
@IrAlfred @Baraka24 Please try and see if this issue is reproducible. Thanks!

mercihabam:
[...]

PR #1643, opened 2025-08-29
fix(frontend): ensure that selected emails persist after periodic message list refresh, fixing misbehavior of copy/move actions

mercihabam:


marclaporte:
@kroky Here is some context: https://github.com/cypht-org/cypht/issues/1368

PR #1645, opened 2025-08-29
refactor(frontend): update wordpress header layout for improved styling and responsiveness

Grandi0z:
[context ](https://avan.tech/item139406)

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -795,7 +795,7 @@ return [
     //     |
     //     | Github repository tracking.
     //     */
-        'github',
+    //     // 'github',
 
     //     /*
     //     | ---------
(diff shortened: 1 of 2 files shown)


config/app.php
```diff
@@ -795,7 +795,7 @@
     //     |
     //     | Github repository tracking.
     //     */
-    //     // 'github',
+        'github',
```
kroky: We don't use these config values anymore. Enabled modules should come from env variable, like the line 641 suggests.

Grandi0z: Oh, thank you.

I just realized that by default, both modules were commented out—I must have introduced this error while working on the GitHub module."

Issue #1540, opened 2025-06-07
🐛 [Bug] No Pagination in Folders


ulfgebhardt:
## 🐛 Bug

Select Folder, shows `Page 1 in header`

![Image](https://github.com/user-attachments/assets/0584489c-5f1c-4719-93d8-c92bb9072bc8)

Misses a Pagination Control at the bottom.

I did not find any way navigate to page 2.

This is an IMAP account

### Version & Environment

Rev: [v2.4.2]

OS:  []

marclaporte:
@ulfgebhardt Can you please test master? 

We are stabilizing master, from which we will cut Cypht 2.5.0

kroky:
Note that master removed pagination in combined pages which used to be the way it worked in Cypht 1.x. Having paging in combined pages make it rather slow and hard to execute sorting correctly. If combined pages list sources for a specific period of time, it means they contain limited amount of items, by definition. If we still face issues with longer combined pages, we can introduce client-side pagination but server-side, we still need to get all emails in the chosen period of time for all sources in order to sort them correctly.

ulfgebhardt:
@marclaporte I use the project in production and I shy away from using the current master, since if needed its not guaranteed to be able to roll back.

@kroky I have this behaviour in an impa folder, not a combined folder

marclaporte:
> since if needed its not guaranteed to be able to roll back

Right. But can't you run both in parallel? (I do)

marclaporte:
@ulfgebhardt We have a test site for which we could give you access. On this site, you could test Cypht master with our bogus data. 

If you detect and report bugs on this test site, it is super easy for our dev team to address because they have full access to all the code and the server.

What do you think?

ulfgebhardt:
I can try to reproduce problems on your test environment if you like.

I can certainly say that there is many problems on my end your developers cannot reproduce and I wonder why this is.
An example is this issue or https://github.com/cypht-org/cypht/issues/1368 which I still experience.

I am running v2.4.2 - maybe an easy solution could be to release more often? I can certainly update to stable versions regularly.

marclaporte:
> maybe an easy solution could be to release more often?

Yes, that is what we must do, but it's not always easy :-)
https://github.com/cypht-org/cypht/issues/1624

marclaporte:
> I can try to reproduce problems on your test environment if you like.

Super! I will email you later today with credentials and put you in touch with main active devs.

marclaporte:
[...]

PR #1647, opened 2025-08-30
fix(frontend): restore missing pagination in folders

mercihabam:
Related: https://github.com/cypht-org/cypht/issues/1540


marclaporte:
@kroky 

We discussed at the https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings yesterday.

The plan is to release 2.4.3 with this fix

PR #1648, opened 2025-08-30
fix(frontend): redirect to the login screen when the session has expired or has been invalidated

mercihabam:
This addresses an issue reported via email by @ulfgebhardt, in which he wrote:
> One thing I noticed: I switch User Agent every 10 Minutes, which seems to kick me off the platform (invalidate auth). I know that this is a feature of Cypht. But it does not send me to the login screen, just confronts me with a lot of Server errors. This could be a real bug, but needs further investigation whether my guess is correct.

### DIFF
--- a/modules/core/navigation/navigation.js
+++ b/modules/core/navigation/navigation.js
@@ -102,6 +102,16 @@ async function navigate(url, loaderMessage) {
         if (!html || typeof html !== 'string') {
             throw new Error("Failed navigating to page, empty or invalid response body.");
         }
+
+        // load custom javascript
+        const scripts = extractCustomScripts($(html));
+        loadCustomScripts(scripts);
+
+        if (!hm_is_logged()) {
+            logout();
+            return;
+        }
+
         const titleMatch = html.match(/<title[^>]*>([\s\S]*?)<\/title>/i);
         if (!titleMatch) {
             throw new Error("No <title> element found in response HTML");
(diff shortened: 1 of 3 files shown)


modules/core/navigation/navigation.js
```diff
@@ -102,6 +102,16 @@ async function navigate(url, loaderMessage) {
         if (!html || typeof html !== 'string') {
             throw new Error("Failed navigating to page, empty or invalid response body.");
         }
+
+        // load custom javascript
+        const scripts = extractCustomScripts($(html));
+        loadCustomScripts(scripts);
```
kroky, line 108: If we load the custom scripts before the main content is replaced, wouldn't some of them fail by actually working on the old page content?

mercihabam: Seems like none of them actually depend on the DOM content; they are all state data. So, none would fail.

mercihabam:
The Selenium tests pipeline is persistently failing, and I suspect these changes may have triggered the issue. I’m currently investigating and will move this PR out of draft once it’s resolved.

modules/core/navigation/navigation.js
kroky: I see, then we should be good.

mercihabam:
> The Selenium tests pipeline is persistently failing, and I suspect these changes may have triggered the issue. I’m currently investigating and will move this PR out of draft once it’s resolved.

Seems unrelated after all, the pipeline passed on a rerun.

PR #1651, opened 2025-09-01
fix(backend): exclude_auto_bcc causes some unrelated messages to disappear from the inbox

mercihabam:
Some emails were being excluded from the mailbox due to the `NOT HEADER X-Auto-Bcc cypht` filter. After debugging, it was found that this filter matches messages containing the keywords `header`, `auto`, `cypht`, etc., anywhere in the message, including the subject or body. This caused some emails in my inbox not to appear, even though they were visible in earlier versions.

This [thread](https://github.com/cypht-org/cypht/issues/1540#issuecomment-3226086175) is a perfect example: all related emails in it are not shown.

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -2314,7 +2314,7 @@ if (!class_exists('Hm_IMAP')) {
          * @param string $filter can be one of ALL, SEEN, UNSEEN, ANSWERED, UNANSWERED, DELETED, UNDELETED, FLAGGED, or UNFLAGGED
          * @return array list of IMAP message UIDs
          */
-        public function get_message_sort_order($sort='ARRIVAL', $reverse=true, $filter='ALL', $terms=array(), $exclude_deleted=true, $exclude_auto_bcc=true, $only_auto_bcc=false) {
+        public function get_message_sort_order($sort='ARRIVAL', $reverse=true, $filter='ALL', $terms=array(), $exclude_deleted=true, $exclude_auto_bcc=false, $only_auto_bcc=false) {
             if (!$this->is_clean($sort, 'keyword') || !$this->is_clean($filter, 'keyword') || !$this->is_supported('SORT')) {
                 return [];
             }


marclaporte:
Perhaps related: https://github.com/cypht-org/cypht/pull/1152

FYI @Shadow243

kroky:
Ok but we need to make sure that when exclude auto bcc is true, it works as expected. I'll merge this as missing emails is probably a bigger pain than showing irrelevant messages but maybe @Shadow243 can work on a proper fix for exclude auto bcc option?

Shadow243:
> Ok but we need to make sure that when exclude auto bcc is true, it works as expected. I'll merge this as missing emails is probably a bigger pain than showing irrelevant messages but maybe @Shadow243 can work on a proper fix for exclude auto bcc option?

@kroky I think I found it, this parameter was integrated with `true` starting here: https://github.com/cypht-org/cypht/blame/master/modules/core/hm-mailbox.php#L562  in https://github.com/cypht-org/cypht/commit/6bb0bde824c953e8d29d81953731b3f3c8e1ef4f for performance reasons during search.  
From here: https://github.com/cypht-org/cypht/blob/master/modules/imap/functions.php#L803 the value `true` is sent by default, so changing it here: https://github.com/cypht-org/cypht/pull/1651/files#diff-2a16fbb7ff5f2c47ea84a2060f25affdc0fbeb9df48cff9f9438fa7ec705f624R2317 does not solve issue in this PR.

`$exclude_auto_bcc` should have a parameter just like `smtp_auto_bcc`, which can be enabled or disabled.  
It was apparently planned initially, but I cannot find where it was set to `true` except for its default value,  everywhere else it is normally `false`.

@marclaporte @kroky @mercihabam  
I traced it further back: https://github.com/cypht-org/cypht/commit/0820cfba459a77519c1efd4ba99724b4d6eea4ac → the first introduction of the feature.  
We could rather add it as a parameter that the user explicitly enables.

`server_supports_custom_headers` (integer) starting here: https://github.com/cypht-org/cypht/pull/1152/files is only used to check if the server supports custom headers (e.g., Stalwart does not support this).

kroky:
Parameter or not, we still have to make sure that searching for this works. If the server does not support it for some reason - either don't allow user to enavble the feature say that the feature is disabled for the account (if enabled globally) or find a way to search by custom headers. We shouldn't leave it half-broken like this.

Shadow243:
> Parameter or not, we still have to make sure that searching for this works. If the server does not support it for some reason - either don't allow user to enavble the feature say that the feature is disabled for the account (if enabled globally) or find a way to search by custom headers. We shouldn't leave it half-broken like this.

Sure, I've tested it, and I was able to search with [[hm-mailbox.php#L562](https://github.com/cypht-org/cypht/blame/master/modules/core/hm-mailbox.php?utm_source=chatgpt.com#L562)](https://github.com/cypht-org/cypht/blame/master/modules/core/hm-mailbox.php#L562)
 set to false by default, and I still got the result, including the hidden email but when i keep true there as default value I can't anymore get the result i'm looking for. That’s why I would suggest making this parameter dynamic through a configuration setting.

kroky:
That's what I am saying. You should be able to search for those excluding auto bcc... that was the feature implemented but it didn't count the fact that some servers can't search for it (or need another syntax to search for it). Adding an on/off feature is not enough here.

mercihabam:
Correct, Victor. Ideally, we should aim to make the filter work as expected. However, I didn’t have enough time to dig into this deeply and figure out the appropriate syntax to achieve the desired outcome. Either way, I think Steven’s idea of adding a setting to switch this behavior on or off is also worth considering.

Given the current priority of stabilizing master, I think we’ll have to collaborate with @Shadow243 on this a bit later and make sure it’s implemented correctly. I’ll create a task on our intranet. Thanks!

PR #1666, opened 2025-09-10
fix(frontend): message store fails to remove non-existent rows from combined lists on update

mercihabam:
This can be observed when listing emails from the flagged list, opening one email, unflagging it, and then returning to the flagged list — the email still appears until the refresh button is pressed.

### DIFF
--- a/modules/core/js_modules/Hm_MessagesStore.js
+++ b/modules/core/js_modules/Hm_MessagesStore.js
@@ -70,50 +70,37 @@ class Hm_MessagesStore {
         const sourcesToRemove = Object.keys(this.sources).filter(key => !this.currentlyAvailableSources().includes(key));
         sourcesToRemove.forEach(key => delete this.sources[key]);
 
-        Promise.all(
-            this.fetch(hideLoadingState).map((req) => {
-                return new Promise(async (resolve) => {
-                    const { formatted_message_list: updatedMessages, pages, folder_status, do_not_flag_as_read_on_open, sourceId } = await req;
-                    // count and pages only available in non-combined pages where there is only one ajax call, so it is safe to overwrite
-                    this.count = folder_status && Object.values(folder_status)[0]?.messages;
-                    this.pages = parseInt(pages);
-                    this.newMessages = this.getNewMessages(updatedMessages);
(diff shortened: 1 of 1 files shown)


modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -235,12 +248,13 @@ class Hm_MessagesStore {
...
         let store = this;
         return this.getRequestConfigs().map((config) => {
+            const initialConfig = Object.assign([], config);
             return new Promise((resolve, reject) => {
                 Hm_Ajax.request(
                     config,
                     (response) => {
                         if (response) {
-                            response.sourceId = store.hashObject(config);
+                            response.sourceId = store.hashObject(initialConfig); // Do not use this config object because the request appends a "hm_page_key" entry, which would change the hash
```
kroky, line 244: This fix should be sufficient. The problem with message not being removed is that sourceId was artificially changed because of the injected config variables is Hm_Ajax. If the sourceId stays as it should - the original code above correctly removes all messages not part of the original source list and only then adds the new ones. There is no need to do another filter and sort at the end - it will further make the app slower which we don't want to. Please leave only this fix here.

mercihabam: Correct! That one alone achieves the intended outcome. I’ve updated the code.

kroky: Thanks!

modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -104,7 +104,7 @@ class Hm_MessagesStore {
             if (messagesReadyCB) {
                 messagesReadyCB(this);
             }
-        }, this);
+        });
```
kroky, line 107: Wouldn't this change the object context?

mercihabam: I accidentally removed it, though the context isn’t impacted since it refers to the same object as the callback’s parent context.

kroky: Ah, because of the arrow function. Alright.

PR #1670, opened 2025-09-11
fix(frontend): keep track keep track of current page number when sorting, and display correct sort value when reloading

mercihabam:


modules/imap/output_modules.php
```diff
@@ -19,7 +19,7 @@ class Hm_Output_imap_custom_controls extends Hm_Output_Module {
     protected function output() {
         if ($this->get('custom_list_controls_type')) {
             $filter = $this->get('list_filter');
-            $sort = $this->get('list_sort');
+            $sort = $this->get('sort');
```
kroky: Imap handler module for combined pages definitely outputs list_sort while core module outputs both sort and list_sort coming from request. This is very confusing. I am not really sure when sort is used and when list_sort. Can you dig deeper and try to come up with just one variable for sorting? If we really need to, try documenting what is the difference.

mercihabam: `list_sort` is widely used, while `sort` was only used in the combined_sort_dialog function as a fallback alongside `list_sort`. So, I changed everything to use `list_sort`. Note that on the client side, the value is always sent through the key `sort`.

Issue #359, opened 2019-10-25
binding to LDAP with authentication does not work


robert-winkler:
## Switching from anonymous to authenticated binding breaks LDAP contacts

I store my contacts in OpenLDAP. Anonymous binding to the DB to read the contacts in cypht works. However, when changing to `auth=true` in the `ldap.ini`, the connection to the LDAP server breaks. There is no error message.
I tried both: 1) entering the login data in the `ldap.ini` (`"admin"`, `"cn=admin"`, ..); and 2) leaving `admin='' path='' blank` and configuring the access data in the Web-Interface.

I consider this issue **critical**, since **anonymous binding to an LDAP server should be disallowed**.

The admin connection to the LDAP server and reading/writing entries using phpLDAPadmin works; thus **the configuration of the server looks** fine.

### Version and Environment

Rev: I used the install script. There is no obvious version, but this might help:
1664 Jul 19 20:04 composer.json

OS:  Ubuntu 18.04.3 LTS
PHP: PHP 7.2.19-0ubuntu0.18.04.2 (cli) (built: Aug 12 2019 19:34:28) ( NTS )

jasonmunro:
Sounds like a bug. Sorry for the late response. I will check into it as soon as I can (also adding this as a must fix before the next RC release). Thanks for the report!

jasonmunro:
This seems to be working for me here, but there could definitely still be a bug. One thing I noticed is that when I set my credentials via the ini file, I need to setup the entire dn, but when doing so via the settings page only the username and the code builds the dn.

For example my username in the ini file set to "cn=admin,dc=foo,dc=com" translates to just "admin" when setting the username using the settings page.

Best way to troubleshoot this is to make the following code changes locally, then load the contacts page, and check wherever PHP errors are logged (also you must be in debug mode for the errors to be logged).

Code changes to debug: https://gist.github.com/jasonmunro/2afcaae292757c87e7c87a6593b8fc42
Info on debug mode: https://github.com/jasonmunro/cypht/wiki/Troubleshooting-Login-Issues#debug-mode

Hope that helps!

robert-winkler:
Hi Jason,

1. I applied the patch and run> php /usr/local/share/cypht/scripts/config_gen.php.

2. opened mail-debug.

3. The admin is set in the Site settings.

4. and looked for the error.log file after trying to add a contact in LDAP_contacts.

=> no error! (I created a fresh file and it remains empty..)

[...]

PR #1673, opened 2025-09-11
fix(backend): fix LDAP address book and authentication issues

IrAlfred:
# This PR aims to fix these issues

- [x] : ldap with authentication breaks ldap contacts (worked correctly with anonymous binds but failed when using an authenticated service account.) Related issue: [https://github.com/cypht-org/cypht/issues/359](https://github.com/cypht-org/cypht/issues/359)
- [x] : Add support for both cn (Common Name) and uid (User ID) as unique identifiers. Support for other identifiers like sAMAccountName or userPrincipalName (common in Active Directory) is considered out of scope for this PR and requires further research.

### DIFF
--- a/modules/contacts/modules.php
+++ b/modules/contacts/modules.php
@@ -123,6 +123,7 @@ class Hm_Handler_process_send_to_contact extends Hm_Handler_Module {
         if (array_key_exists('contact_id', $this->request->get)) {
             $contacts = $this->get('contact_store');
             $contact = $contacts->get($this->request->get['contact_id']);
+            
             if ($contact) {
                 $to = sprintf('%s <%s>', $contact->value('display_name'), $contact->value('email_address'));
                 $this->out('compose_draft', array('draft_to' => $to, 'draft_subject' => '', 'draft_body' => ''));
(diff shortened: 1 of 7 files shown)


modules/contacts/modules.php
```diff
@@ -123,6 +123,30 @@ public function process() {
         if (array_key_exists('contact_id', $this->request->get)) {
             $contacts = $this->get('contact_store');
             $contact = $contacts->get($this->request->get['contact_id']);
+            if (!$contact && 
+                array_key_exists('contact_type', $this->request->get) &&
+                array_key_exists('contact_source', $this->request->get) &&
+                array_key_exists('dn', $this->request->get) &&
+                $this->request->get['contact_type'] == 'ldap') {
```
kroky: Contacts module itself shown't know about ldap. Try to put this functionality in ldap_contacts module - you can override modules, execute other modules after/before certain ones - there's a lot of flexibility how to do this.

modules/contacts/modules.php
```diff
@@ -396,16 +420,42 @@ protected function output() {
...
+                            }
+                        }
+                        $edit_url = '?page=contacts&amp;contact_id='.$this->html_safe($c->value('id')).'&amp;contact_source='.
                             $this->html_safe($c->value('source')).'&amp;contact_type='.
-                            $this->html_safe($c->value('type')).'&amp;contact_page='.$current_page.
-                            '" class="edit_contact cursor-pointer" title="'.$this->trans('Edit').'"><i class="bi bi-gear ms-2"></i></a>';
+                            $this->html_safe($c->value('type')).'&amp;contact_page='.$current_page;
+                        
+                        // Add DN parameter for LDAP contacts lookup
+                        if ($c->value('type') == 'ldap' && $c->value('all_fields')) {
```
kroky: There is too much ldap exposure here - try to put this in ldap_contacts module and use a specific ldap contact type subclass of Hm_Contact where all this manipulation of dn and other fields will happen.

modules/contacts/modules.php
```diff
@@ -607,6 +626,15 @@ function build_contact_detail($output_mod, $contact, $id) {
     $all_fields = false;
     $contacts = $contact->export();
     ksort($contacts);
+    
+    if (class_exists('Hm_LDAP_Contact') && $contact->value('type') == 'ldap') {
```
kroky: This is a bit better approch than before but still not in line with Cypht code philosophy and a better code pattern could be used. The basic idea is that ldap module is a distinct module (think of it as a distinct class) and making one class look into the internal architecture or use internal functionality of another one breaks a few OOP principles. There are patterns designed to solve these problems. I see a few possibilities here:
- redefine build_contact_details and similar functions in ldap module and replace the default contact's module function as a whole when it is an ldap contact. This way we completely override the behavior but might end up duplicating some code
- if we want to reuse code from this module but insert special code from other modules, we need to refactor and add some kind of a bride or template method - e.g. split this function in several ones and override only the one we really need in ldap module (that adds the additional row of DN content) OR make this function include additional content where it needs to - this is something you tried to do but it should not know if the contact is ldap or if an ldap module exists - it should just call the contact or whatever object is used to check if any additional info must be shown.
I hope you see how you can refactor this to keep OOP principles and make the code easier to work with and more extensible in the future.

IrAlfred: Noted ! Thanks @kroky

IrAlfred: > This is a bit better approch than before but still not in line with Cypht code philosophy and a better code pattern could be used. The basic idea is that ldap module is a distinct module (think of it as a distinct class) and making one class look into the internal architecture or use internal functionality of another one breaks a few OOP principles. There are patterns designed to solve these problems. I see a few possibilities here:
> 
> * redefine build_contact_details and similar functions in ldap module and replace the default contact's module function as a whole when it is an ldap contact. This way we completely override the behavior but might end up duplicating some code
> * if we want to reuse code from this module but insert special code from other modules, we need to refactor and add some kind of a bride or template method - e.g. split this function in several ones and override only the one we really need in ldap module (that adds the additional row of DN content) OR make this function include additional content where it needs to - this is something you tried to do but it should not know if the contact is ldap or if an ldap module exists - it should just call the contact or whatever object is used to check if any additional info must be shown.
>   I hope you see how you can refactor this to keep OOP principles and make the code easier to work with and more extensible in the future.

I found that the dn was already in the contact data, so made some changes add it in the contact details.

PR #1674, opened 2025-09-11
fix(backend): sievefilters sieve_can_delete_folder handler registration for I…

Shadow243:
Discovered while working on https://github.com/cypht-org/cypht/pull/1604

Screenshot 2025-09-12 at 02 29 16


marclaporte:
Good catch!

PR #1679, opened 2025-09-13
fix(frontend): non-existent messages are still not being removed in some cases despite e55265e

mercihabam:


modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -105,6 +102,8 @@ class Hm_MessagesStore {
                 }
             });
 
+            this.rows = this.rows.filter(row => Object.values(this.sources).flat().indexOf(row['1']) !== -1);
```
kroky, line 105: Do you have a specific reproducible use-case? I'd like to avoid costly operations at all costs here as it is already slower than the previous Cypht version. Looping through the whole list of sources doesn't make sense to me and is costly.

mercihabam: Please see [the flagged list for this server](https://elastictrunk.corp1.evoludata.com/sms-at-evoludata-ca?page=message_list&list_path=flagged).

I understand the need to improve Cypht’s current performance, but I’m wondering whether it would make a difference to loop through each source’s content individually, as opposed to executing the same action in one go once all sources are gathered with their values. Wouldn’t the latter be even better?

kroky: I cannot reproduce what you describe. Do you have a reproducible use-case?
The idea of per-source list is killed by your update and I am not sure it is better. I will probably spend some more time optimizing the filters and slower calls in those methods but for now I wanted to see what is the problem here.

mercihabam: I am still able to reproduce this on the mailbox I shared. Here is the pattern: visit the flagged list, open a message (the last one or any other), unflag it, and then go back to the list. After the list auto-refreshes, the email does not disappear. I don’t know if this happens only in my browser, nor do I think that’s the case, but that’s the pattern at least!

kroky: Maybe your browser's cache is old and the fixes we introduced didn't correctly deal with already cached wrong data. Please click 'clear cache' button under the menu and try again. I cannot reproduce this problem and I think the new code works correctly without the fix here.

mercihabam: If the problem can’t be reproduced on your side, then all good. It was likely a caching issue on my end. Let’s close this PR, thanks!

Issue #1078, opened 2024-06-11
Review status with https://coveralls.io/


marclaporte:
https://github.com/cypht-org/cypht/blob/master/README.md links to an outdated https://coveralls.io/github/jasonmunro/cypht?branch=master

PR #1682, opened 2025-09-14
feat(unit): add GitHub Actions workflow for coverage reporting

Shadow243:
Issue reported at https://github.com/cypht-org/cypht/issues/1078


Shadow243:
The test failures have been addressed here in this PR https://github.com/cypht-org/cypht/pull/1683

PR #1687, opened 2025-09-16
fix(frontend): collapsed sidebar layout appears slightly broken

mercihabam:
The sidebar spacing got adjusted in https://github.com/cypht-org/cypht/pull/1653, though it seems the collapsed state was missed.


mercihabam:
@pom2ter it looks good, so going with `130px`. Thanks!

PR #1689, opened 2025-09-16
fix(frontend): add explicit handler for search form submission to pre…

Grandi0z:
[Related task](https://avan.tech/item142916)

Added explicit click handler for `.search_update` button in `route_handlers.js` to ensure search works on first click.

### DIFF
--- a/modules/core/js_modules/route_handlers.js
+++ b/modules/core/js_modules/route_handlers.js
@@ -93,6 +93,21 @@ function applySearchPageHandlers(routeParams) {
     $('.search_reset').on("click", Hm_Utils.reset_search_form);
     $('.combined_sort').on("change", function() { Hm_Message_List.sort($(this).val()); });
 
+    // Add explicit handler for search form submission to fix double-click issue
+    $('.search_update').on("click", function(e) {
+        e.preventDefault();
+        const form = $(this).closest('form');
+        if (form.length) {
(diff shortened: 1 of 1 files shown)


marclaporte:
Super!

Once it's merged in master (and thus will be part of 2.5.0 soon), please also fix in 1.4.x

modules/core/js_modules/route_handlers.js
```diff
@@ -93,6 +93,21 @@ function applySearchPageHandlers(routeParams) {
     $('.search_reset').on("click", Hm_Utils.reset_search_form);
     $('.combined_sort').on("change", function() { Hm_Message_List.sort($(this).val()); });
 
+    // Add explicit handler for search form submission to fix double-click issue
+    $('.search_update').on("click", function(e) {
```
kroky, line 97: .search_update is a submit button. We shouldn't need to emulate its behaviour of form submission. The proper fix here would be to find the reason why clicking it one time doesn't submit the form and fix that - probably other event handlers kicking in?

Grandi0z: This bug seems to be fixed since yesterday. I am looking for the recent PR to identify the fix

marclaporte: FYI: https://dev.tiki.org/How-to-figure-out-which-commit-causes-a-bug

PR #1690, opened 2025-09-17
refactor(frontend): improve contact list display with tooltips and responsive styles

Grandi0z:
[Related task](https://avan.tech/item104817)

trtr


modules/contacts/site.js
```diff
@@ -260,3 +260,10 @@ var check_cc_exist_in_contacts_list = function() {
     }
     return "";
 };
+// Initialize Bootstrap tooltips for contact names
+$(document).ready(function() {
+    var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+    var tooltipList = tooltipTriggerList.map(function (tooltipTriggerEl) {
+        return new bootstrap.Tooltip(tooltipTriggerEl);
```
kroky: Shouldn't these be initialized by default when the page loads and bootstrap initializes? I don't think it is a good idea to manually initialize them here.

Grandi0z: You're right.

 Bootstrap tooltips didn’t auto-init. Because Cypht uses its own system, I replaced them with native tooltips.

modules/local_contacts/modules.php
```diff
@@ -257,14 +257,17 @@ protected function output() {
...
             'value="'.$this->html_safe($phone).'" /><br />'.
-            '<label class="screen_reader" for="contact_group">'.$this->trans('Contact Group').'</label>'.
+            '<label class="form-label" for="contact_group">'.$this->trans('Contact Group').'</label>'.
             '<select class="form-select" id="contact_group" name="contact_group">'.
-            '<option value="'.$this->trans('Personal Addresses').'"'.(isset($group) && $this->html_safe($group) == $this->trans('Personal Addresses') ? ' selected' : '').'>'.$this->trans('Personal Addresses').'</option>'.
-            '<option value="'.$this->trans('Trusted Senders').'"'.(isset($group) && $this->html_safe($group) == $this->trans('Trusted Senders') ? ' selected' : '').'>'.$this->trans('Trusted Senders').'</option>'.
-            '<option value="'.$this->trans('Collected Recipients').'"'.(isset($group) && $this->html_safe($group) == $this->trans('Collected Recipients') ? ' selected' : '').'>'.$this->trans('Collected Recipients').'</option>' .
+            '<option value="'.$this->trans('Personal Addresses').'" selected>'.$this->trans('Personal Addresses').'</option>'.
```
kroky: Why do we remove the check for existing groups?

PR #1693, opened 2025-09-17
fix(frontend/backend): advanced search not displaying any messages

IrAlfred:
This PR resolves a critical bug where the advanced search functionality failed to display messages in the UI, even though the server was returning valid data successfully. The issue was caused by an incorrect handling of the response data structure in the core message list update logic.

### DIFF
--- a/modules/advanced_search/modules.php
+++ b/modules/advanced_search/modules.php
@@ -370,7 +370,20 @@ class Hm_Output_filter_imap_advanced_search extends Hm_Output_Module {
      */
     protected function output() {
         if ($this->get('imap_search_results')) {
-            prepare_imap_message_list($this->get('imap_search_results'), $this, 'advanced_search');
+            $adv_search_result = format_imap_message_list(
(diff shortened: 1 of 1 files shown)


modules/core/site.js
```diff
@@ -624,7 +624,11 @@ function Message_List() {
             msgArray = Object.entries(msgArray);
         }
         for (let row of msgArray) {
-            Hm_Utils.tbody().append(row[0]);
+            if (row[1] && typeof row[1] === 'object' && row[1][0]) {
```
kroky: Can you check all calling code here and make sure it passes the same msg array format? We don't want to deal with arbitrary formats and one place was probably missed with recent refactorings...

IrAlfred: > Can you check all calling code here and make sure it passes the same msg array format? We don't want to deal with arbitrary formats and one place was probably missed with recent refactorings...

@kroky You were right, I've made some changes to align with the correct format in the advanced_search module. Elsewhere we had the same format

PR #1696, opened 2025-09-18
test(unit): add comprehensive unit test suites for core library classes & lib

Shadow243:
####  New Test Suites

- tests/phpunit/modules/core/mailbox.php – Tests for Hm_Mailbox
  - Bridge pattern for IMAP, JMAP, EWS, SMTP
  - Connection/authentication handling
  - Search across protocols
  - Error handling and edge cases

- tests/phpunit/lib/searchable.php – Tests for Searchable trait
  - getBy() method with parameters and filters
  - Edge cases and errors

- tests/phpunit/lib/scram_authenticator.php – SCRAM authentication
  - Hash algorithm detection/validation
  - Client proof for multiple algorithms
  - Authentication flow

- tests/phpunit/lib/ini_set.php – PHP configuration validation
  - Session security, compression, open_basedir
  - Compatibility: PHP 7.3+, 8.0+, 8.1

### DIFF
--- a/tests/phpunit/lib/api_curl.php
+++ b/tests/phpunit/lib/api_curl.php
@@ -1,0 +1,176 @@
+<?php
+
+/**
+ * Unit tests for Hm_API_Curl
+ * @package lib/tests
+ */
+
+use PHPUnit\Framework\TestCase;
+
+class Hm_Test_API_Curl extends TestCase {
+
+    public function setUp(): void {
+        require_once __DIR__.'/../bootstrap.php';
+        // Reset Hm_Functions statics to known defaults before each test
+        Hm_Functions::$exec_res  = '{"unit":"test"}';
+        Hm_Functions::$curl_fail = false;
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_default_format_is_json(): void {
+        $api = new Hm_API_Curl();
+        $this->assertSame('json', $api->format);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_format_can_be_set_in_constructor(): void {
+        $api = new Hm_API_Curl('xml');
+        $this->assertSame('xml', $api->format);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_command_parses_json_response(): void {
+        Hm_Functions::$exec_res = '{"key":"value","num":42}';
+        $api    = new Hm_API_Curl();
+        $result = $api->command('https://example.com');
+        $this->assertSame(['key' => 'value', 'num' => 42], $result);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_command_returns_empty_array_for_invalid_json(): void {
+        Hm_Functions::$exec_res = 'not valid json {{';
+        $api    = new Hm_API_Curl();
+        $result = $api->command('https://example.com');
+        $this->assertSame([], $result);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_command_returns_empty_array_for_null_json(): void {
+        Hm_Functions::$exec_res = 'null';
+        $api    = new Hm_API_Curl();
+        $result = $api->command('https://example.com');
+        $this->assertSame([], $result);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_command_returns_raw_response_for_non_json_format(): void {
+        Hm_Functions::$exec_res = '<html>raw response</html>';
+        $api    = new Hm_API_Curl('html');
+        $result = $api->command('https://example.com');
+        $this->assertSame('<html>raw response</html>', $result);
+    }
+
+    /**
+     * @preserveGlobalState disabled
+     * @runInSeparateProcess
+     */
+    public function test_command_returns_empty_array_when_curl_init_fails(): void {
+        Hm_Functions::$curl_fail = true;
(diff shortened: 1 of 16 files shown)


tests/phpunit/lib/searchable.php
```diff
@@ -0,0 +1,234 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Tests for the Searchable trait
+ */
+class Hm_Test_Searchable extends TestCase {
+
+    public function setUp(): void {
+        if (!defined('APP_PATH')) {
+            require_once __DIR__.'/../bootstrap.php';
```
kroky: Proper bootstrap means we don't call these inside the unit tests themselves. Why would the app path not be defined? Tests should be executed by a single phpunit command which should always properly bootstrap the framework.

tests/phpunit/lib/searchable.php
```diff
@@ -0,0 +1,234 @@
...
+        }
+        // Load mocks first
+        if (!class_exists('Hm_Mock_Session', false)) {
+            require_once APP_PATH.'tests/phpunit/mocks.php';
```
kroky: This is not the correct place to include such files. Do it in the bootstrap system.

tests/phpunit/mocks.php
```diff
@@ -189,6 +189,80 @@ public function __construct($type) {
         $this->type = $type;
     }
 }
+
+trait Mock_Searchable {
```
kroky: None of the classes or traits called mocks are actual mocks - you copy/implement details from the actual code in the test suite - this is counterproductive as code will diverge over time. See the documentation on how to mock objects https://docs.phpunit.de/en/9.6/test-doubles.html#mock-objects  - you add some expectations to methods that should be called and what should be returned but you actually work with the real objects and not some classes that resemble the real classes partially. In theory, it is best to follow this pattern:
- use the object itself whenever possible
- if testing becomes complicated as the original class depends on 3rd party external connections, you can mock the result of some of its methods and return what you want to test
- note that merely testing the return of data that you supplied as a mock is not a productive test. The test objective is to test code that runs in the framework and not code that is written in the tests themselves. So, whenever you want to test a method, see what external connections it uses, mock them and actually call that method on that original class/object and test that it works correctly. There is a subtle but important difference here. The way you did it with these mocks is that you test the mocks themselves which doesn't make sense.

Also, you shouldn't use reflection to set private properties or do similar changes that were not designed to be done with the class in question. If you want to test something that's not available for some reason, do:
- think if it actually needs to be tested. Private methods usually are not tested. Testing the public interface of the class is enough to ensure it works fine as it calls private methods internally, so they become part of the tests.
- if something private really needs to be tested, refactor the original class/interface and make it public or separate it to parts that are testable
Refactoring original code because of testing is a legitimate use-case, so don't be afraid to do it.

Shadow243: @kroky I’ve applied all the requested corrections and fixed the other failing tests.

For Hm_Test_Mailbox, I added some imports in setUp to avoid conflicts. These are only required there and not elsewhere.

Shadow243: In addition, in hm-jmap.php I added a few missing compatibility functions (is_supported, get_message_sort_order, and sort_by_fetch) with TODOs and comments. These provide placeholders so we can later implement proper JMAP behavior and confirm everything works correctly with JMAP as well.

modules/imap/hm-ews.php
```diff
@@ -25,1130 +25,1132 @@
  * public interface to EWS mailboxes
  * @subpackage imap/lib
  */
-class Hm_EWS {
-    protected $ews;
-    protected $api;
-    protected $authed = false;
+if (!class_exists('Hm_EWS')) {
```
kroky: Please remove this check. Require the file just ones and it will be good. You changed the whole file indentation because of this check when PHP already has a mechanism for including class files only once.

Shadow243: Done

IrAlfred:
This is a very big pull request and affects too many modules. A more iterative approach (closing this and submitting smaller) module-specific PRs would likely lead to faster feedback and safer merges. What's your thoughts @kroky @Shadow243 ?

kroky:
I prefer we do all the related/intended work on the tests here in one PR even if it becomes big. It is easier to maintain from my point of view.

Shadow243:
> I prefer we do all the related/intended work on the tests here in one PR even if it becomes big. It is easier to maintain from my point of view.

Noted, Work in progress.

Shadow243:
@kroky please take a look at this again.

tests/phpunit/lib/api_curl.php
```diff
@@ -0,0 +1,176 @@
...
+ * Unit tests for Hm_API_Curl
+ * @package lib/tests
+ */
+
+use PHPUnit\Framework\TestCase;
+
+class Hm_Test_API_Curl extends TestCase {
+
+    public function setUp(): void {
+        require_once __DIR__.'/../bootstrap.php';
```
kroky: This should load for all unit tests during bootstrap process not within each individual unit test.

tests/phpunit/lib/repository.php
```diff
@@ -0,0 +1,260 @@
+<?php
+
+/**
+ * Unit tests for Hm_Repository trait
+ * @package lib/tests
+ */
+
+use PHPUnit\Framework\TestCase;
+
+class Hm_Test_Repository extends TestCase {
```
kroky, line 10: Why do you call it repository test when you test only the taggs wrapper?

kroky:
Overall, it looks good and can be merged after we resolve the 2 issues.

tests/phpunit/lib/api_curl.php
Shadow243: Done

tests/phpunit/lib/repository.php
Shadow243: I applied required changes.

PR #1698, opened 2025-09-20
fix(frontend): empty state renders a few ms before responses are processed, and in some cases some responses remain unprocessed due to a cleared timeout

mercihabam:


modules/core/js_modules/Hm_MessagesStore.js
```diff
@@ -119,17 +119,22 @@ class Hm_MessagesStore {
...
-            return req.then((response) => {
-                pendingResponses.set(response.sourceId, response);
+            return new Promise((resolve) => {
+                req.then((response) => {
+                    pendingResponses.set(response.sourceId, response);
 
-                if (processingTimeout) {
-                    clearTimeout(processingTimeout);
-                }
+                    if (processingTimeouts[response.sourceId]) {
```
kroky: I don't understand the need for this complexity. The previous timeout is designed to clear any pending processing as it is going to initiate a new pending processing each time the process finishes. This allows batching for processes that finish almost at the same time. Making timeouts per-source seems to defate this purpose - why do we set timeouts at all in that case?

mercihabam: It seems to happen that processes that start sequentially (one after the other in a batch), the following process hinders the execution of the preceding one by clearing its timer, so we end up with promises that do not fulfill. That's what I tried to deal with.

kroky: Timer is to process pending responses and is not related to any specific process. If 5 start sequentially, 2 end in 10ms difference, they are processed, the other 3 are processed when they end. If one timer is cleared out, another one is set by the code that follows, so I don't see how it is possible to miss something. If you have a reproducible use-case, I will take a look.

mercihabam: The issue mainly arises intermittently. When you load a combined page, all processes execute smoothly. However, when you leave the page and return to it, some processes are executed, but others are not.

kroky: I can't reproduce the issue you describe. Maybe you can show it to us on the upcoming meeting?

mercihabam: Sure, will show it during the upcoming meeting.

mercihabam: Unfortunately, I didn’t manage to stay until the end of the meeting due to a temporary power outage. Anyway, I figured out that all responses are indeed processed, but some promises don’t resolve because their resolve calls aren’t executed in the pending response execution processes, but rather in timed functions. That said, I’ll come up with a much better approach...

mercihabam: Done in the latest commit. Please take a look at the changes when you get a chance.

kroky: Yes, looks good, thanks!

PR #1699, opened 2025-09-20
fix(other): display messages of type event calendar with ews

### LINKED ISSUE #1672: 🐛 [Bug] EWS: Messages of type event, calendar
A constant has been made, for the EWS server, a message of type event, calendar are not loaded on the list of messages in the inbox folder.

By using exchange OVH to test Exchange Web Services, the messages are clearly visible (ref: capture below), which is not possible with cypht

Image

**view in Cyph**t:

Image

Baraka24:
### Issues

- fixes #1672
- relates #1516

### DIFF
--- a/modules/imap/hm-ews.php
+++ b/modules/imap/hm-ews.php
@@ -290,7 +290,7 @@ class Hm_EWS {
 
             if ($this->is_distinguished_folder($folder)) {
                 $folderObj = new Type\DistinguishedFolderIdType($folder);
-            } else {                
+            } else {
                 $folderObj = new Type\FolderIdType($folder);
             }
             $msg = new Type\MessageType();
@@ -523,9 +523,7 @@ class Hm_EWS {
 
         $itemIds = [];
         foreach ($items as $item) {
-            if (is_object($item)) {
-                $itemIds[] = bin2hex($item->getItemId()->getId());
-            }
+            $itemIds[] = bin2hex($item->getItemId()->getId());
         }
         return [$result->getTotalItemsInView(), $itemIds];
     }
@@ -679,13 +677,6 @@ class Hm_EWS {
             'content-type' => null,
             'references' => null,
             'message_id' => null,
-            // Calendar-specific properties
-            'calendar_start' => $message->getStart() ? $message->getStart()->format('Y-m-d H:i:s') : null,
-            'calendar_end' => $message->getEnd() ? $message->getEnd()->format('Y-m-d H:i:s') : null,
-            'calendar_location' => $message->getLocation() ?: null,
-            'is_all_day' => $message->getIsAllDayEvent() ?: false,
-            'is_meeting' => $message->getIsMeeting() ?: false,
-            'is_recurring' => $message->getIsRecurring() ?: false,
         ];
 
         return $props;
@@ -707,12 +698,6 @@ class Hm_EWS {
             'content-type' => null,
             'references' => $message->getReferences() ?: null,
             'message_id' => $message->getInternetMessageId() ?: null,
-            // Meeting request specific properties
-            'meeting_request_type' => $message->getMeetingRequestType() ?: null,
-            'response_type' => $message->getResponseType() ?: null,
-            'is_delegated' => $message->getIsDelegated() ?: false,
-            'has_been_processed' => $message->getHasBeenProcessed() ?: false,
-            'is_out_of_date' => $message->getIsOutOfDate() ?: false,
         ];
 
         return $props;
@@ -1045,10 +1030,10 @@ class Hm_EWS {
             $struct[$part_num]['lines'] = substr_count($content, "\n");
             $struct[$part_num]['md5'] = '';
             $struct[$part_num]['disposition'] = $part->getContentDisposition();
-            
+
             if ($filename = $part->getFilename()) {
                 $struct[$part_num]['file_attributes'] = ['filename' => $filename];
-                
+
                 if ($part->getContentDisposition() == 'attachment') {
                     $struct[$part_num]['file_attributes']['attachment'] = true;
                 }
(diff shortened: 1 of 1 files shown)


modules/imap/hm-ews.php
```diff
@@ -481,13 +482,39 @@ public function search($folder, $sort, $reverse, $flag_filter, $offset, $limit,
         }
         $request = Type::buildFromArray($request);
         $result = $this->ews->FindItem($request);
-        $messages = $result->getItems()->getMessage() ?? [];
-        if ($messages instanceof Type\MessageType) {
-            $messages = [$messages];
+        $items = [];
+        $responseItems = $result->getItems();
+
+        if (method_exists($responseItems, 'getMessage') && $responseItems->getMessage()) {
```
kroky: Can you spare the 2 calls to getMessage (and similar ones below) - one should be sufficient by assigning it to a variable?

modules/imap/hm-ews.php
```diff
@@ -481,13 +482,39 @@ public function search($folder, $sort, $reverse, $flag_filter, $offset, $limit,
...
+            $meetingRequests = $responseItems->getMeetingRequest();
+            if (!is_array($meetingRequests)) {
+                $meetingRequests = [$meetingRequests];
+            }
+            $items = array_merge($items, $meetingRequests);
+        }
+
+        $itemIds = [];
+        foreach ($items as $item) {
+            if (is_object($item) && method_exists($item, 'getItemId')) {
```
kroky: I don't understand all these method_exist checks. Isn't the lib defining these per the API? Maybe we need to merge the current EWS fixes PR and then these will appear automatically?

modules/imap/hm-ews.php
```diff
@@ -522,24 +550,25 @@ public function get_message_list($itemIds, $include_preview = false) {
...
             $msg = [
                 'uid' => $uid,
                 'flags' => implode(' ', $flags),
-                'internal_date' => $message->getDateTimeCreated(),
-                'size' => $message->getSize(),
-                'date' => $message->getDateTimeReceived(),
-                'from' => $this->extract_mailbox($message->getFrom()),
-                'to' => $this->extract_mailbox($message->getToRecipients()),
-                'subject' => $message->getSubject(),
+                'internal_date' => method_exists($message, 'getDateTimeCreated') && $message->getDateTimeCreated() ?
```
kroky: We can do it more intelligently here - instead of all these method_exists checks - make a check of the message type and have a specific method that returns the proper format of the message based on type - e.g. getMessageProperties($message) and then based on message type return the same set of array keys but with the proper values.

modules/imap/hm-ews.php
```diff
@@ -924,9 +957,25 @@ protected function extract_mailbox($data) {
             }
             return $result;
         } elseif (is_object($data) && $data->Mailbox) {
-            return $data->Mailbox->getName() . ' <' . $data->Mailbox->getEmailAddress() . '>';
-        } elseif (is_object($data) && $data->getMailbox()) {
-            return $data->getMailbox()->getName() . ' <' . $data->getMailbox()->getEmailAddress() . '>';
+            if(is_array($data->Mailbox)) {
```
kroky, line 1095: I'm pretty sure the other open EWS PR has these updates and the ones below. Please reuse them by pushing your changes to that PR directly or wait for it to be merged and then update this one.

Baraka24:
Thanks @kroky  for reviews, we kept the logic as you requested, please review again.

modules/imap/hm-ews.php
kroky: This has not been resolved. How do you see it? We shouldn't go over the same or similar code review 2 times and then also risk running into conflicts. Furthermore, I still see method_exists calls here and there and that's most probably because the other PR has not been merged (https://github.com/cypht-org/cypht/pull/1635). I have added comments to the other MR and we can merge it soon. Then, you merge master and go ahead with a final review here.

modules/imap/hm-ews.php
```diff
@@ -481,13 +482,36 @@ public function search($folder, $sort, $reverse, $flag_filter, $offset, $limit,
...
+        if (method_exists($responseItems, 'getMeetingRequest') && $meetingRequests = $responseItems->getMeetingRequest()) {
+            if (!is_array($meetingRequests)) {
+                $meetingRequests = [$meetingRequests];
+            }
+            $items = array_merge($items, $meetingRequests);
+        }
+
+        $itemIds = [];
+        foreach ($items as $item) {
+            if (is_object($item)) {
```
kroky: I am really curious what else would that be than an object?

modules/imap/hm-ews.php
```diff
@@ -522,58 +547,191 @@ public function get_message_list($itemIds, $include_preview = false) {
...
+                $message->getDateTimeReceived()->format('Y-m-d H:i:s.u') :
+                ($message->getStart() ? $message->getStart()->format('Y-m-d H:i:s.u') : date('Y-m-d H:i:s.u')),
+            'from' => $message->getOrganizer() ? $this->extract_mailbox($message->getOrganizer()) : '',
+            'to' => $message->getRequiredAttendees() ? $this->extract_mailbox($message->getRequiredAttendees()) : '',
+            'subject' => $message->getSubject() ?: '',
+            'content-type' => null,
+            'references' => null,
+            'message_id' => null,
+            // Calendar-specific properties
+            'calendar_start' => $message->getStart() ? $message->getStart()->format('Y-m-d H:i:s') : null,
```
kroky: Are these additional properties going to be used somewhere? If they are not, I prefer we don't return extra data, use more memory and make it slower just for the sake of returning what is provided by EWS.

modules/imap/hm-ews.php
marclaporte: @Baraka24 https://github.com/cypht-org/cypht/pull/1635 has been merged

marclaporte:
FYI: https://github.com/cypht-org/cypht/pull/1635 has been merged.

PR #1701, opened 2025-09-23
fix(backend): skip servers with failed connections in IMAP message list handler

Grandi0z:
[Related task](https://avan.tech/item140149)

**Problem:** PHP Fatal error: Call to a member function search() on bool when visiting message list pages with no IMAP servers configured.
**Solution:** Added error checking to skip failed mailbox connections before calling ->search() method.


modules/imap/handler_modules.php
```diff
@@ -1338,6 +1338,11 @@ public function process() {
         foreach ($ids as $key => $id) {
             $details = Hm_IMAP_List::dump($id);
             $mailbox = Hm_IMAP_List::get_connected_mailbox($id, $this->cache);
+
+            if (!$mailbox || !$mailbox->authed()) {
+                continue; // Skip this server if connection failed
```
kroky: Can we also add a feedback msg (warning or error) that server named xyz failed to connect?

PR #1705, opened 2025-09-24
fix(other): remove envelope-to address exclusion from reply-all logic

IrAlfred:
## Description

This change improve consistency in multi-profile environments. Previously, the system excluded email addresses based on delivery the header Delivered-To, which caused inconsistent behavior when emails were copied between different mailboxes. 

The exclusion of the current mailbox is already handled in JavaScript.

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -142,11 +142,6 @@ function reply_to_address($headers, $type) {
     $msg_cc = '';
     $headers = lc_headers($headers);
     $parsed = array();
-    $delivered_address = false;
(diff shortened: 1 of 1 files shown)


modules/core/message_functions.php
```diff
@@ -143,10 +143,6 @@ function reply_to_address($headers, $type) {
     $headers = lc_headers($headers);
     $parsed = array();
     $delivered_address = false;
```
kroky: You can also clean up the delivered_address variable usage as it is no longer needed.

PR #1706, opened 2025-09-25
fix(other): remove envelope-to address exclusion from reply-all logic

IrAlfred:
(cherry picked from commit d1874f878d526753c0c5d76219ae77593458ca55)

### DIFF
--- a/modules/core/message_functions.php
+++ b/modules/core/message_functions.php
@@ -120,11 +120,6 @@ function reply_to_address($headers, $type) {
     $msg_cc = '';
     $headers = lc_headers($headers);
     $parsed = array();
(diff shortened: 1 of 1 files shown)


kroky:
Are you sure the removal logic of the current address is in 1.4 as well, namely your comment "The exclusion of the current mailbox is already handled in JavaScript."?

IrAlfred:
> Are you sure the removal logic of the current address is in 1.4 as well, namely your comment "The exclusion of the current mailbox is already handled in JavaScript."?

Yes, here is the commit https://github.com/cypht-org/cypht/commit/5c4840b4960df8c30b016bf3e1dd4561cbf12a70

Issue #1658, opened 2025-09-05
🐛 [Bug] config/database.php - CREATE TABLE uses mysql / mariadb key work "lock" and won't create table.


jeremiah:
## 🐛 Bug

https://github.com/cypht-org/cypht/blame/a5498a45e67525379aa9d1d17418f5534c00fef2/config/database.php#L78

The instructions for creating a table for MySQL / MariaDB / SQLite include the column name "lock" which is a reserved keyword in mysql. Has this schema been tested since MySQL / MariaDB on Debian 13 will not create this table. 

### Version & Environment

Rev: [master-cypht]

OS:  [Debian 13]

Baraka24:
Thanks for your report, we'll look into this.

PR #1707, opened 2025-09-25
fix(other): remove mistaken lock column from hm_user_session schema

Baraka24:
### Issues

- #1658
- Remove the mistakenly added 'lock INTEGER DEFAULT 0' column from the MySQL/SQLite CREATE TABLE statement for hm_user_session table in the database configuration comments.

### DIFF
--- a/config/database.php
+++ b/config/database.php
@@ -74,7 +74,10 @@ return [
     |  Postgresql:
     |   CREATE TABLE hm_user_session (hm_id varchar(250) primary key not null, data text, hm_version INTEGER DEFAULT 1, date timestamp);
     |
-    |  MySQL or SQLite:
+    |  MySQL:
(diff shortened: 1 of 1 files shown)


marclaporte:
@Shadow243 Seems to have been added here: https://github.com/cypht-org/cypht/pull/1380

Shadow243:
> @Shadow243 Seems to have been added here: #1380

Exactly, the lock is only used with SQLite, and we now have migrations here:  
https://github.com/cypht-org/cypht/tree/master/database/migrations  

The lock migration is here:  
https://github.com/cypht-org/cypht/blob/master/database/migrations/sqlite/20241209010201_add_lock_columns.sql  

We highly recommend using migrations, like this one does:  
https://github.com/cypht-org/cypht/blob/master/scripts/setup_database.php

Shadow243:
@Baraka24  The documentation probably needs an update regarding this, in  https://github.com/cypht-org/cypht/blob/master/config/database.php

Baraka24:
@Shadow243, Understod. Thanks. I have updated the doc.

Issue #341, opened 2019-04-26
Logging in after changing password leads to broken settings


HanPrower:
## 🐛 Bugreport
After changing account password with` php ./scripts/update_password.php username password` can't login without deleting cookies. Once logged in settings of account no longer seem to be attached.

Changing the password back to what it was before allows settings to load correctly.

I see no way to change the password internally.

### Version & Environment
Commit: https://github.com/jasonmunro/cypht/commit/8f76381981376fc5b045c3bda37acaa3f434bf87
OS: Arch Linux (rolling)
Auth: Database (mariadb) + Redis
Environment:
```
PHP version	7.3.4
Zend version	3.3.4
SAPI	fpm-fcgi
Enabled Modules	core, pop3, imap, smtp, account, idle_timer, desktop_notifications, themes, nux, developer, history, saved_searches, advanced_search, profiles, inline_message, imap_folders, keyboard_shortcuts
```

### Steps to reproduce
1. Make account and set up some settings, save & logout
2. Change password using `php ./scripts/update_password.php username password` (the password hash in the db does change)
3. Try and login; prompted is wrong
4. Delete cookies and login again; this time it logs in
5. Notice any saved settings are missing (general, email servers, etc.)
6. Use script to set password back to what it was, delete cookies and login and settings are back

I did see https://github.com/jasonmunro/cypht/issues/98, but it doesn't seem relevant to this.

jasonmunro:
Hello, thanks for the feedback. A couple thoughts:

- When using built in authentication you should have a Settings -> Password page in the main menu that allows you to change your password without losing your settings.
- It's a known issue that changing your password externally from Cypht (including the cli script, which I sort of forgot about), causes settings to be lost. There is a recover_settings module set however that you can enable to recover those presuming the old password is still known.

We could also improve that cli script to take both the old and new password so that it can decrypt then re-encrypt settings so this does not happen when using it.

HanFox:
Hm, not sure how I missed the Settings -> Password. I swear it wasn't there before... probably just blind.

When you do use that form, however, it does throw an error. When looking at the error log it seems related to a missing translation. The password does change though.

jasonmunro:
We log missing translations as a way to easily know what needs to be added to the translation files, but those messages are information only and will not cause bad behavior (aside from not being translated properly).

marclaporte:
Lost settings reminds me of https://github.com/jasonmunro/cypht/issues/349

jasonmunro:
[...]

PR #1719, opened 2025-10-02
feat(other): preserve user settings during password update

IrAlfred:
[Related issue](https://github.com/cypht-org/cypht/issues/341)

### DIFF
--- a/scripts/update_password.php
+++ b/scripts/update_password.php
@@ -2,17 +2,38 @@
 
 /**
  * CLI script to update a user password
+ *
+ * Usage:
+ * - With old password (preserves user data): php ./scripts/update_password.php <username> <old_password> <new_password>
+ * - Without old password (data loss warning): php ./scripts/update_password.php <username> <new_password>
  */
 if (mb_strtolower(php_sapi_name()) !== 'cli') {
     die("Must be run from the command line\n");
 }
 
-if (is_array($argv) && count($argv) == 3) {
-    $user = $argv[1];
-    $pass = $argv[2];
-}
-else {
-    die("Incorrect usage\n\nphp ./scripts/update_password.php <username> <password>\n\n");
+$user = '';
+$old_pass = '';
+$new_pass = '';
+$incorrect_usage_msg = "Incorrect usage\n\nUsage:\n" .
+            "  With old password (preserves user config and needs [sudo]):\n    php ./scripts/update_password.php <username> <old_password> <new_password>\n\n" .
+            "  Without old password (user config will be lost):\n    php ./scripts/update_password.php <username> <new_password>\n\n";
+
+if (is_array($argv)) {
+    if (count($argv) == 4) {
+        // Mode 1: With old password
+        $user = $argv[1];
+        $old_pass = $argv[2];
+        $new_pass = $argv[3];
+        $preserve_data = true;
+    } elseif (count($argv) == 3) {
+        $user = $argv[1];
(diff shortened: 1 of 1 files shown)


scripts/update_password.php
```diff
@@ -2,17 +2,38 @@
...
-    $user = $argv[1];
-    $pass = $argv[2];
-}
-else {
-    die("Incorrect usage\n\nphp ./scripts/update_password.php <username> <password>\n\n");
+$user = '';
+$old_pass = '';
+$new_pass = '';
+$incorrect_usage_msg = "Incorrect usage\n\nUsage:\n" .
+            "  With old password (preserves user config and needs [sudo]):\n    sudo php ./scripts/update_password.php <username> <old_password> <new_password>\n\n" .
```
kroky: Why sudo?

scripts/update_password.php
```diff
@@ -40,11 +61,57 @@
 }
 
 $auth = new Hm_Auth_DB($config);
-if ($user && $pass) {
-    if ($auth->change_pass($user, $pass)) {
-        die("Password Updated\n\n");
+
+if (! $auth->check_credentials($user, $old_pass)) {
```
kroky: The check here will fail when user didn't supply an old password - it needs to be in the block that handles conversion using the old password.

scripts/update_password.php
```diff
@@ -40,11 +61,57 @@
...
+            $user_config->save($user, $new_pass);
+            echo "Password updated successfully.\n\n";
+        } catch (Exception $e) {
+            echo "Warning: Password updated but could not save user configuration: " . $e->getMessage() . "\nUser may need to reconfigure their settings.\n\n";
+        }
+
+    } else {
+        echo "Warning: Could not decrypt existing user configuration with provided old password.\nUser data may not be preserved. Continue? (y/N): ";
+        $handle = fopen("php://stdin", "r");
+        $input = trim(fgets($handle));
```
kroky: Please move this input retrieval to a function that can be reused.

scripts/update_password.php
```diff
@@ -40,11 +61,57 @@
...
+    $input = trim(fgets($handle));
+    fclose($handle);
+
+    if ($input !== 'YES') {
+        die("Password update cancelled.\n\n");
     }
-    else {
-        die("An error occured\n\n");
+
+    if ($auth->change_pass($user, $new_pass)) {
```
kroky: This actual change with the error message if it failed repeats several times. Please dry up the code a bit.

scripts/update_password.php
IrAlfred: The user config being saved in the location `/var/lib/...` will need sudo to overwrite

kroky: It is not Cypht's responsibility to check if user running the command has permission to write to that file. I suggest you just keep the php commands here and let user decide when to use sudo. This way, it doesn't look weird when you list 2 similar commands one with sudo and one not. Just ensure that we show a proper error message before changing the password if the new settings cannot be saved due to permission issue.
Also, user can configure the location of this settings file, so there are many variables not in Cypht's code control that determine the permission checks at the end.

kroky:
@Shadow243 please check this one as well - we still have intermittent selenium test problems. The pipeline ran correctly before merging but failed after merging in master:
https://github.com/cypht-org/cypht/actions/runs/18274901513/job/52024917036

We need to make these more stable.

Issue #398, opened 2020-05-06
Configurable custom actions (buttons)


marclaporte:
Cypht has standard actions (delete, archive, etc.). Let's take this up a notch.

I do the same actions again and again. It would be nice to configure these actions. A bit like macros.

Part of the logic for filters: https://github.com/jasonmunro/cypht/issues/9 but manual trigger of actions.

Ex.: 
* move email to a folder XYZ in another account
* forward to ABC and delete
* forward to DEF and archive

Thanks!

dumblob:
@marclaporte you mean something like a button **run now** next to each filter chain?

marclaporte:
@dumblob: yes, that as well. Any set of actions really.

dumblob:
@marclaporte hm, I mean, having a button **run now** next to each filter is very easy to implement, but if there are any other actions (which? I think there do not exist any other meaningful actions except for [sending emails](https://github.com/jasonmunro/cypht/issues/115 )), it'd get more complicated to implement.

marclaporte:
Maybe I am misunderstanding. Can you elaborate on "filter chain"?

"which?": This is a long term goal: Parse emails for structured data and move to a database (http://doc.tiki.org/Trackers). Imagine invoices by email that are added to the accounting system.

I would like a button to bounce an email back to sender, simulating a non existant account.

Some more ideas here:
http://james.apache.org/mailet/standard/mailet-report.html

For the short term, I want buttons like "move to accounting inbox", "move to accounting archive"

marclaporte:
Also, one of our goals is to support the use case of generic mailboxes like info@, support@, accountspayable@, etc. Please see: http://doc.tiki.org/Groupmail

Actions could be:
* Add a threaded discussion about this email to coordinate on how to handle/reply.
* Convert to a task

I realize a lot of this will go beyond the use case of standalone Cypht. Configurable buttons could have 10 possible actions in standalone and 20 when embedded within an app like Tiki. And sooner or later, other web apps will follow our footsteps for JMAP and programmable emails!

dumblob:
@marclaporte now I see what you mean. IMHO that has not much to do with Cypht, but rather with modules which Cypht already supports. Cypht itself has no need for such actions - they don't exist now, so there is no need for any buttons/... (i.e. no user interface).

But of course, you're free to write your own extension modules for Cypht and you'll get your buttons in the UI as well.

[...]

PR #1726, opened 2025-10-03
feat(frontend): start from one or many emails , create an action filter

camilevahviraki:
### Issues
 
- https://avan.tech/item143515
- https://github.com/cypht-org/cypht/issues/398

- None

### Todo

- [X] None


modules/imap/site.js
```diff
@@ -922,7 +922,7 @@ var search_selected_for_imap = function() {
         $('.imap_move').on("click", function() { return false; });
         bootstrap.Dropdown.getOrCreateInstance($('.move_to_location')[0]).hide();
         $('.move_to_location').html('');
-    }
+    }es
```
marclaporte: This looks odd.

camilevahviraki: Thanks. It has been fixed.

modules/core/site.js
```diff
@@ -2588,6 +2587,379 @@ function setupActionSnooze(callback) {
     $(document).on('change', '.nexter_input_snooze', callback);
 }
 
+const Hm_Filters = (function (my) {
```
kroky: Please move this to sievefilters or whatever module is responsible for the functionality. We don't want to pollute the core module with functionality that might not even be enabled.

modules/core/site.js
```diff
@@ -2588,6 +2587,379 @@ function setupActionSnooze(callback) {
...
+          header_fields +=
+            '<option value="' +
+            value.name +
+            '">' +
+            value.description +
+            "</option>";
+        }
+      });
+
+      let extra_options =
```
kroky: If we are building all this html inline, can you at least use template literals and spare the quotes on each line?

modules/imap/handler_modules.php
```diff
@@ -1933,6 +1933,12 @@ public function process() {
             $this->out('imap_accounts', $this->user_config->get('imap_servers'), array());
             $this->out('show_pagination_links', $this->user_config->get('pagination_links_setting', true));
             $this->out('auto_advance_email_enabled', $this->user_config->get('auto_advance_email_setting', true));
+
+            $parts = explode('_', 'imap_' . $form['imap_server_id'] . '_' . $form['folder'], 3);
+            $details = Hm_IMAP_List::dump($parts[1]);
```
kroky: Why don't you simply use $form['imap_server_id'] here? I don't see a reason for the explode...

modules/imap/output_modules.php
```diff
@@ -420,6 +423,57 @@ protected function output() {
                     $txt .= '<span class="text-decoration-none btn btn-sm btn-outline-danger" data-bs-toogle="tooltip" title="This functionality requires the email server support &quot;Sieve&quot; technology which is not provided. Contact your email provider to fix it or enable it if supported."><i class="bi bi-lock-fill"></i> <span id="filter_block_txt">'.$this->trans('Block Sender').'</span></span>';
                 }
             }
+
+            if ($this->get('sieve_filters_enabled')) {
```
kroky: No need to check for this. Whenever you find yourself in the situation to check if a module is enabled, you should immediately think if you put the code in the right module. This code needs to go in sievefilters module - it can be an output module that is executed after this one here. If you need to put the content in a specific place in the middle of the headers section, add a placeholder for it or insert after an element.

modules/imap/site.js
```diff
@@ -960,6 +960,86 @@ var imap_move_copy = function(e, action, context) {
...
+            "danger"
+        );
+        $btn.prop("disabled", false);
+        return;
+    }
+
+    edit_filter_modal.open();
+
+    // After modal content is ready, auto-fill conditions
+    setTimeout(() => {
```
kroky: Please avoid timeouts as they are inconsistent - e.g. it might take more than 250ms for the modal to load and then your function will fail. Instead, use proper callbacks after modal has loaded.

modules/sievefilters/site.js
```diff
@@ -186,6 +190,455 @@ var hm_sieve_possible_actions = function() {
...
+    const add_filter_action = Hm_Filters.add_filter_action;
+
+
+/**************************************************************************************
+*                                      MODAL EVENTS
+**************************************************************************************/
+var hm_sieve_button_events = function (edit_filter_modal, edit_script_modal) {
+
+  $(document)
+    .off("click")
```
kroky: You are removing all 'click' event handlers on the document element - this will have severe side-effects. If you want to avoid attaching duplicate event handlers, use proper event filtering when removing event handlers - they need to be the same as the ones you use in the 'on click' handler.

modules/sievefilters/site.js
```diff
@@ -186,6 +190,455 @@ var hm_sieve_possible_actions = function() {
...
+    hm_sieve_current_account = $(this).attr("account");
+    edit_script_modal?.open();
+  });
+
+  /**
+   * Delete action Button
+   */
+  $(document).on("click", ".delete_else_action_modal_button", function (e) {
+    e.preventDefault();
+    $(this).parent().parent().remove();
```
kroky: This is usually a bad idea. Try closest('.something-that-is-less-lilkely-to-change')

modules/core/site.js
camilevahviraki: Thanks, good point — I moved this logic to the sievefilters module

modules/imap/site.js
camilevahviraki: Done, Timeout removed as suggested.

modules/sievefilters/site.js
camilevahviraki: Done — updated to use closest() with a stable selector.

modules/imap/output_modules.php
camilevahviraki: Done, I have placed the button in the sievefilters module

modules/core/site.js
```diff
@@ -2528,3 +2527,122 @@ window.addEventListener('page-change', () => {
...
+window.addEventListener('page-change', () => {
+    document.getElementById("cypht-upgrade-alert")?.addEventListener("close.bs.alert", function () {
+        Hm_Utils.save_to_local_storage('cypht_upgrade_alert_dismissed', '1');
+    });
+    if (Hm_Utils.get_from_local_storage('cypht_upgrade_alert_dismissed') !== '1') {
+        $("#cypht-upgrade-alert").addClass("show");
+    } else {
+        $("#cypht-upgrade-alert").addClass("hide");
+    }
+})
```
kroky, line 2548: The above 2 event handlers seem to repeat 3 times. Not sure if you resolved conflicts after merge but please check what you are committing - repeating code should be relatively easy to spot.

modules/imap/site.js
```diff
@@ -961,6 +963,87 @@ var imap_move_copy = function(e, action, context) {
     return false;
 };
 
+$(function () {
+    $(document).on("submit", "#create-filter-form", function (e) {
```
kroky: If this is connected with sievefilters module, why is it in imap module js file?

modules/sievefilters/site.js
```diff
@@ -186,6 +190,798 @@ var hm_sieve_possible_actions = function() {
...
+          }
+        });
+        edit_filter_modal.setTitle(current_editing_filter_name);
+        edit_filter_modal.open();
+      }
+    );
+  });
+  return true;
+};
+
```
kroky, line 934: Same comment here as your other big MR - if you do code style updates, do them in a separate MR and list the reasons you do so. Copy-pasting large amounts of code and changing them only slightly makes it difficult to review the actual changes. What did you try to do here with these 700+ lines of moved code?

modules/imap/site.js
camilevahviraki: This was mistakenly placed there during development and has now been moved to the correct sievefilters module.

modules/sievefilters/site.js
camilevahviraki: In this case, this was not a code style update. The code was moved to make the sieve filters functions accessible outside the sieveFiltersPageHandler() scope, since they were previously defined inside that function and could not be reused by other parts of the module. The intent was to expose and reuse this functionality, not to change style. I agree that the diff is large, but the main goal here was to change the scope/structure so the functions can be shared.

modules/core/site.js
camilevahviraki: Thank you.
It has been fixed.

modules/sievefilters/site.js
kroky: OK, sounds good.

modules/core/site.js
```diff
@@ -2528,3 +2527,35 @@ window.addEventListener('page-change', () => {
     }
 })
 
+class Hm_Filter_Modal extends Hm_Modal {
```
kroky: This seems to use methods from sievefilters module. Why don't you define the whole Hm_Filter_Modal there?

camilevahviraki: Moved the method to sieve filter

PR #1727, opened 2025-10-04
fix(backend): clear session cache for IMAP folders when enabled on load …

Shadow243:
Cypht is not updating the folder list even after clicking reload. If you create a folder outside of Cypht, you can't see it even after reload because of the cached data.

This PR adds cache deletion on reload click, so that expanding the server folder will reload from the server and reset the cache.

### DIFF
--- a/modules/core/site.css
+++ b/modules/core/site.css
@@ -1261,7 +1261,6 @@ div.unseen,
 }
 [data-title] {
   position: relative;
-  z-index: 10 !important;
 }
 
 #list_controls_menu {


modules/imap/handler_modules.php
```diff
@@ -882,6 +882,9 @@ public function process() {
         $folders = array();
         if (!empty($servers)) {
             foreach ($servers as $id => $server) {
+                if ($this->config->get('allow_session_cache', false)) {
+                    $this->cache->del('imap_folders_imap_'.$id.'_');
```
kroky, line 888: I think this module is called in more places than the reload button. Shouldn't we keep the cache in that case? Maybe pass a parameter to update_folder_list JS function that calls the ajax_hm_folders ajax action?

Shadow243: @kroky I updated

PR #1731, opened 2025-10-08
fix(frontend): mobile responsiveness ajustments

pom2ter:
Some fixes to the mobile responsiveness

### DIFF
--- a/modules/feeds/site.css
+++ b/modules/feeds/site.css
@@ -25,7 +25,7 @@
 }
 
 /* Mobile responsiveness */
-@media (max-width: 768px) {
+@media (max-width: 767.98px) {
     .msg_headers .fw-semibold {
         min-width: 60px;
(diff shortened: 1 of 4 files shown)


modules/feeds/site.css
```diff
@@ -25,7 +25,7 @@
 }
 
 /* Mobile responsiveness */
-@media (max-width: 768px) {
+@media (max-width: 767.98px) {
```
marclaporte, line 28: Sometimes, 0.02px makes a difference. Pixel perfect! :-)

pom2ter: Thats the recommended value for Bootstrap.
https://getbootstrap.com/docs/5.0/layout/breakpoints/#max-width

It isnt that important to be precise but i got an issue while testing and had to change it.

PR #1733, opened 2025-10-08
fix(backend): exclude auto-bcc'd messages php-side, preventing inaccurate filter results from IMAP, which does not seem to perform negative exact header match

mercihabam:


modules/imap/hm-imap.php
```diff
@@ -1110,6 +1112,15 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
                     }
                 }
             }
+
+            if ($exclude_auto_bcc) {
+                foreach ($headers as $k => $v) {
+                    if ($v['x_auto_bcc']) {
```
kroky: Previously, it tried to exclude only those matching 'cypht' and now it is excluding everything with that header. Can we keep previous functionality?

modules/imap/hm-imap.php
```diff
@@ -1110,6 +1112,15 @@ public function get_message_list($uids, $raw=false, $include_content_body = fals
                     }
                 }
             }
+
+            if ($exclude_auto_bcc) {
+                foreach ($headers as $k => $v) {
```
kroky: Maybe you can just continue and skip adding it to the headers array if you detect this in a loop that already happens above. Why do we loop twice, otherwise?

modules/imap/hm-imap.php
mercihabam: Sure!

modules/imap/hm-imap.php
mercihabam: No need to loop twice. Thanks.

modules/core/hm-mailbox.php
```diff
@@ -559,25 +559,25 @@ public function set_search_charset($charset) {
...
+    public function search($folder, $target='ALL', $terms=array(), $sort=null, $reverse=null, $exclude_deleted=true, $only_auto_bcc=false) {
         if (! $this->select_folder($folder)) {
             return [];
         }
         if ($this->is_imap()) {
             if ($sort) {
                 if ($this->connection->is_supported('SORT')) {
                     // use fast sort extension and search simultanously
-                    $uids = $this->connection->get_message_sort_order($sort, $reverse, $target, $terms, $exclude_deleted, $exclude_auto_bcc, $only_auto_bcc);
+                    $uids = $this->connection->get_message_sort_order($sort, $reverse, $target, $terms, $exclude_deleted, $only_auto_bcc);
```
kroky, line 570: @mercihabam I need to reopen this one as it seems like a big regression.

Previously, there were 2 settings - to exclude auto-bcc ones and return only auto-bcc ones. Now, when this PR got merged in master and backported back to 2.x, we have the problem that whenever `DEFAULT_SETTING_ENABLE_EXCLUDE_AUTO_BCC=true` is set, it sends true to the `only_auto_bcc` flag and thus returns zero results on all combined pages... which is pretty nasty.

Can you push a fix here? I don't think you should remove exclude_auto_bcc params on all these functions. Instead, we should keep both options and use whatever is appropriate. I guess there are cases where only auto-bcc ones are needed and where auto-bcc ones need to be excluded...

We need this backported to 2.x and to Tiki 27.x branch...

mercihabam: @kroky I think this worked fine until the introduction of https://github.com/cypht-org/cypht/pull/1740. The relevant parameter here (exclude auto-bcc) was moved to the `get_message_list()` function so that the filtering can be done PHP-side and not server-side (which caused issues).

The follow-up PR did not pass the new setting to the appropriate function, causing the current issue. I've fixed that here: https://github.com/cypht-org/cypht/pull/2015.

kroky: Ah, maybe merge conflict or merged these 2 PRs simultaneously that worked fine on their own but not when combined... thanks for fixing it!

Issue #337, opened 2019-04-12
MTA-STS with TLS Reporting


dumblob:
## 🚀 Feature

In the light of current news about [Gmail implementing MTA-STS and TLS Reporting](https://www.zdnet.com/article/gmail-becomes-first-major-email-provider-to-support-mta-sts-and-tls-reporting/ ), I'm opening this feature request tracking the potentially needed bits to support [MTA-STS](https://tools.ietf.org/html/rfc8461 ) and [TLS Reporting](https://tools.ietf.org/html/rfc8460 ) also in the MUA Cypht.

Some more information about MTA-STS e.g. on https://www.hardenize.com/blog/mta-sts and also a [validator](https://aykevl.nl/apps/mta-sts/ ) for MTA-STS. See also APNIC information about TLS Reporting - [TLS-RPT](https://blog.apnic.net/2018/11/16/introducing-smtp-tls-reporting/ ).

### Design, Layout, Architecture

1. My first request would be to tell the Cypht user whether when sending someone an email would leverage MTA-STS or not (maybe something similar like the validator above?).
2. As in point (1), but for TLS-RPT.

marclaporte:
Related: https://github.com/jasonmunro/cypht/issues/241

marclaporte:
@Danelif please advise.

marclaporte:
@Danelif so what should be done in Cypht? (vs other parts of the mail stack).

Danelif:
I think we should implement a script that can handle the set up of MTA-STS with TLS RPT automatically at the installation of Cypht.

marclaporte:
@neustradamus @dumblob Any thoughts?

marclaporte:
Related: https://github.com/virtualmin/virtualmin-gpl/issues/808

Danelif:
@marclaporte i started this feature in [#1736](https://github.com/cypht-org/cypht/pull/1736) still draft but we can start discussion

PR #1736, opened 2025-10-11
feat(frontend/backend): show recipient MTA-STS and TLS-RPT policy status

Danelif:
This PR implements comprehensive MTA-STS (Mail Transfer Agent Strict Transport Security) and TLS-RPT (TLS Reporting) support in Cypht, addressing https://github.com/cypht-org/cypht/issues/337.


lib/mta_sts.php
```diff
@@ -0,0 +1,398 @@
...
+    /**
+     * Fetch MTA-STS policy file from domain
+     *
+     * @param string $domain The domain to fetch policy from
+     * @return array|false Parsed policy or false on failure
+     */
+    private static function fetch_policy($domain) {
+        $policy_url = "https://mta-sts.{$domain}/.well-known/mta-sts.txt";
+
+        // Use cURL if available, otherwise use file_get_contents with SSL context
```
kroky: You should try to reuse what is available in Cypht. There are already Hm_Functions methods to use with curl, curl extension is required via composer, so it is safe to assume you can use it. Also, see other modules, hm-feed, for example, users curl and falls back to file_get_contents. Most of the modules use Hm_API_Curl. Only when we need really low-level stream functionality like smtp/imap connections - there we use stream socket client.

modules/mta_sts/module.php
```diff
@@ -0,0 +1,226 @@
...
+
+        if (!empty($compose_to)) {
+            $recipients = $this->parse_recipients($compose_to);
+            $mta_sts_status = array();
+
+            foreach ($recipients as $email) {
+                $domain = Hm_MTA_STS::extract_domain($email);
+                if ($domain) {
+                    $result = Hm_MTA_STS::check_domain($domain);
+                    $tls_rpt = Hm_MTA_STS::check_tls_rpt($domain);
```
kroky: Please refactor the sts class to get rid of all static methods - using a class with static methods like this is not much different than using plain simple functions. I'd like to see class instantiation and you call methods on the object. For example, you can initialize the object with the domain and then just call $sts->check_domain(), $sts->check_tls_rpt while the class itself keeps the data it needs to share locally and privately.

modules/mta_sts/module.php
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        if (!empty($compose_to)) {
+            $recipients = $this->parse_recipients($compose_to);
+            $mta_sts_status = array();
+
+            foreach ($recipients as $email) {
+                $domain = Hm_MTA_STS::extract_domain($email);
+                if ($domain) {
+                    $result = Hm_MTA_STS::check_domain($domain);
```
kroky: This will be quite slow to run on each and every compose page. I think it deserves a setting or a checkbox on the compose page if user wants this behavior.

modules/mta_sts/modules.php
```diff
@@ -0,0 +1,226 @@
...
+
+        return array_unique($emails);
+    }
+}
+
+/**
+ * Output MTA-STS status indicator in compose form
+ * @subpackage mta_sts/output
+ */
+class Hm_Output_mta_sts_status_indicator extends Hm_Output_Module {
```
kroky, line 230: Besides the checks and status indicator don't we need to update our hm-smtp functionality to work with STS? Example concerning point that I don't see implemented: https://datatracker.ietf.org/doc/html/rfc8461#section-5

scripts/setup_mts_sts.sh
```diff
@@ -0,0 +1,460 @@
+#!/usr/bin/env bash
+
+# MTA-STS (Mail Transfer Agent Strict Transport Security) Setup Script
+# This script helps configure MTA-STS and TLS-RPT for email security
+#
+# MTA-STS (RFC 8461): Enables mail servers to declare their ability to receive TLS-secured connections
+# TLS-RPT (RFC 8460): Provides a reporting mechanism for TLS connection issues
+#
+# Usage: ./setup_mta_sts.sh [domain] [mx-servers]
+# Example: ./setup_mta_sts.sh example.com "mail.example.com,mail2.example.com"
```
kroky: This whole script is setting something up for the server. Cypht is a client. I don't think its place is in Cypht. It is more oriented for sysadmins supporting their own smtp servers but they probably know enough to set up the servers themselves.

Danelif: I have removed the script!

modules/mta_sts/modules.php
Danelif: Implementing RFC 8461 Section 5 (actual SMTP client enforcement) would require modifying the hm-smtp.php SMTP client to:

1. Verify MX records match the MTA-STS policy
2. Enforce TLS connections only
3. Validate certificate names against policy
4. Handle policy fetch failures

This is a significantly larger feature that would need to be addressed as a separate issue/PR, as it changes core email sending behavior and could break existing functionality if not implemented carefully.

modules/mta_sts/module.php
Danelif: Added "Enable MTA-STS security status checking" checkbox in Settings > General

modules/mta_sts/module.php
Danelif: Done

lib/mta_sts.php
Danelif: i have changed and  used Hm_Functions::c_init() instead of curl_init(), Hm_Functions::c_setopt() instead of curl_setopt(), Hm_Functions::c_exec() instead of curl_exec(), adn Hm_Functions::c_status() instead of curl_getinfo()

modules/mta_sts/modules.php
kroky: I don't see the purpose of implementing the RFC in 20 or 30%. It will be confusing to the user to enable something that's not fully implemented/working. It is better to address the full implementation in one MR here and test it well before merging.

Makefile
```diff
@@ -30,5 +30,10 @@ setup:  ## locally setup app and users. presumes env vars are set
 	echo "Creating directories and configs"
 	./scripts/setup_system.sh
 
+.PHONY: setup-mta-sts
+setup-mta-sts:  ## setup MTA-STS and TLS-RPT. usage: make setup-mta-sts DOMAIN=example.com MX=mail.example.com EMAIL=admin@example.com
+	@if [ -z "$(DOMAIN)" ]; then echo "Error: DOMAIN is required. Usage: make setup-mta-sts DOMAIN=example.com MX=mail.example.com EMAIL=admin@example.com"; exit 1; fi
+	@if [ -z "$(MX)" ]; then echo "Error: MX is required. Usage: make setup-mta-sts DOMAIN=example.com MX=mail.example.com EMAIL=admin@example.com"; exit 1; fi
+	./scripts/setup_mta_sts.sh -d $(DOMAIN) -m "$(MX)" $(if $(EMAIL),-e $(EMAIL),)
```
kroky: If you removed this script, you should remove this section in the makefile as well, right?

Danelif: yes please  i should remove this aswell

Makefile
```diff
@@ -29,6 +29,3 @@ setup:  ## locally setup app and users. presumes env vars are set
 	./scripts/setup_database.php
 	echo "Creating directories and configs"
 	./scripts/setup_system.sh
-
-help:  ## get help
```
marclaporte, line 33: Is this removal related to MTA-STS?

Danelif: oups wrong removal! let me fix it

Makefile
Danelif: done

modules/mta_sts/modules.php
Danelif: Got it, let me work on that !

Danelif: I reworked this so the PR no longer appears to be full MTA-STS enforcement.
After checking RFC 8461 section 5, the actual enforcement path belongs to the sending MTA that connects to the recipient MX hosts. Cypht normally submits mail to a configured SMTP relay, so implementing full MTA-STS delivery enforcement in the compose UI would be misleading and would not truly comply with the RFC.

I updated the PR to make the scope explicit: it now provides visibility into opt-in recipient-domain MTA-STS/TLS-RPT policies only. I also changed the UI/README wording to avoid implying that Cypht enforces delivery security.

Alongside that, I addressed the related implementation issues:
- renamed `module.php` to `modules.php` so the module loads correctly
- moved inline CSS to `site.css.`
- used the `_setting` user config convention
- removed redirect following when fetching policies
- shared the in-request policy cache across instances
- removed remaining static helper usage
- added tests for policy parsing and cache behavior


So this PR is now intentionally scoped as a policy visibility feature, not a partial RFC enforcement implementation.

modules/mta_sts/modules.php
```diff
@@ -0,0 +1,220 @@
...
+
+        if (empty($compose_to)) {
+            // Check if there are draft recipients
+            $draft = $this->get('compose_draft', array());
+            if (!empty($draft) && isset($draft['draft_to'])) {
+                $compose_to = $draft['draft_to'];
+            }
+        }
+
+        if (empty($compose_to)) {
```
kroky: Shouldn't we get all the recipient domains and check their STS status? I assume you should add some hooks into the smtp module JS, so when user adds email addresses to TO, CC, BCC, etc. fields, there is an Ajax request that checks the STS status (if enabled) and shows that status in a suitable place in the form. At the moment you seem to catch only the case where someone is replying to somebody or the TO field is prefilled.

modules/mta_sts/modules.php
kroky: Yes, I see. We are in display status mode only for now.

modules/mta_sts/modules.php
Danelif: I have updated the implementation to check all compose recipients across To, Cc, and Bcc, including typed recipients and recipient bubbles. The MTA-STS/TLS-RPT status now refreshes through an AJAX check when recipients change, and the initial compose load also handles drafts/replies more completely.

Issue #306, opened 2018-11-28
Integrate advanced search into saved searches module set


jasonmunro:
## 🚀 Feature

Saved searches should be updated to allow saving and rerunning of advanced search parameters
### Design, Layout, Architecture
This should be a pretty easy update. Advanced search parameters for the current search are stored in browser session local storage. The advanced search save feature can read this in and save them server side. It needs to be updated to know the difference between normal and advanced searches when loading one to run. There may need to be some small changes to the advanced search code itself to facilitate the rerunning of saved searches.

marclaporte:
While we work on this, let's keep in mind that an action / filter can be desired as well: https://github.com/jasonmunro/cypht/issues/9

PR #1738, opened 2025-10-11
feat(other): integrate advanced search into saved searches module set

IrAlfred:
Related issue https://github.com/cypht-org/cypht/issues/306

### DIFF
--- a/tests/phpunit/modules/saved_searches/class.php
+++ b/tests/phpunit/modules/saved_searches/class.php
@@ -1,0 +1,264 @@
+<?php
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Tests for Hm_Saved_Searches class
+ * Comprehensive coverage of both search types (simple and advanced):
+ */
+class Hm_Test_Saved_Searches_Class extends TestCase {
+
+    private $saved_searches;
+    private $sample_data;
+    private $advanced_search_data;
+
+    public function setUp(): void {
+        $this->sample_data = array(
+            'Simple Search' => array(
+                'test terms',
+                'SINCE 1-Jan-2023', 
+                'TEXT',
+                'Simple Search'
+            ),
+            'Another Simple' => array(
+                'more terms',
+                'SINCE 1-Jun-2023',
+                'SUBJECT', 
+                'Another Simple'
+            )
+        );
+
+        $this->advanced_search_data = array(
+            'terms' => array(
+                array('term' => 'project meeting', 'condition' => 'and'),
+                array('term' => 'urgent deadline', 'condition' => 'or'),
+                array('term' => 'quarterly report', 'condition' => 'and')
+            ),
+            'targets' => array(
+                array('target' => 'SUBJECT', 'orig' => 'TEXT', 'condition' => 'and'),
+                array('target' => 'FROM', 'orig' => 'SUBJECT', 'condition' => 'or'),
+                array('target' => 'TO', 'orig' => 'FROM', 'condition' => 'and')
+            ),
+            'sources' => array(
+                array('source' => 'imap_0_INBOX', 'label' => 'Gmail - Inbox'),
+                array('source' => 'imap_1_INBOX.Sent', 'label' => 'Yahoo - Sent Mail'),
+                array('source' => 'imap_2_Work.Projects', 'label' => 'Work Email - Projects')
+            ),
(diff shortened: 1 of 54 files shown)


modules/advanced_search/js_modules/route_handlers.js
```diff
@@ -13,6 +13,37 @@ function applyAdvancedSearchPageHandlers() {
...
 
+    $(document).off('click.adv_search_save').on('click.adv_search_save', '.show_save_advanced_search', function(e) {
+        e.preventDefault();
+        $('.save_advanced_search_form').show();
+        $('.show_save_advanced_search').hide();
+        $('.advanced_search_name').focus();
+        return false;
+    });
+
+    $(document).off('click.adv_search_cancel').on('click.adv_search_cancel', '.cancel_save_advanced_search', function(e) {
```
kroky: Why do you need the event specifier here? Can't we just use 'click' as the event name and target the specific button by class name like you do?

kroky:
@IrAlfred thanks, it looks good but can you please add some unit/integration tests to cover the new functionality?

tests/phpunit/modules/saved_searches/class.php
```diff
@@ -0,0 +1,283 @@
...
+class Hm_Test_Saved_Searches_Class extends TestCase {
+
+    private $saved_searches;
+    private $sample_data;
+    private $advanced_search_data;
+
+    public function setUp(): void {
+        require_once __DIR__.'/../../bootstrap.php';
+        require_once APP_PATH.'modules/core/modules.php';
+        require_once APP_PATH.'modules/saved_searches/modules.php';
```
kroky: These three lines are inappropriately placed here. They should be part of the test suite bootstrap process.

tests/phpunit/modules/saved_searches/class.php
```diff
@@ -0,0 +1,283 @@
...
+                array('from' => '2024-01-01', 'to' => '2024-12-31')
+            ),
+            'other' => array(
+                'charset' => 'UTF-8',
+                'limit' => 100,
+                'flags' => array('SEEN')
+            )
+        );
+
+        $this->saved_searches = new Hm_Saved_Searches($this->sample_data);
```
kroky, line 57: If you initialize this here, why do you keep initializing it in individual tests and not just reuse?

tests/phpunit/modules/saved_searches/class.php
IrAlfred: Currently, bootstrap.php is included directly in the test classes. For now, I've kept it because otherwise we'd have to modify all these classes and each time we'd either run the tests with the --bootstrap=bootstrap.php flag or include it in the phpunit.xml configurations.

What do you suggest for this ?

kroky: Hmm, I see. Well, that's a bit weird to include bootstrap over and over again. Bootstrapping processes is meant to be run only once when the test suite starts, it sets up everything and then the tests are executed. setUp class methods are meant to set up the environment for the tests defined in that file and shouldn't do global bootstrapping tasks. I suggest you make sure bootstrap.php is called ones, put it in phpunit.xml, remove run-time require statements and make sure the bootstrap process includes all modules and files, so we don't have to do random require statements here and there.

IrAlfred: Can we do that in a separate PR as it's some kind of refactoring ?

kroky: We won't merge this before the 2.5 release, so you have time to do it here. If you prefer, refactor in another branch and PR but then it will have conflicts as you will have to start from current master and you will end up having to resolve the conflicts.

kroky:
@IrAlfred thanks, I will need some time to review. Will get back to you next week.

kroky:
@IrAlfred please take a look at the coverage report failing: https://github.com/cypht-org/cypht/actions/runs/19029638869/job/54340356825

The github action should be updated to match the refactoring you have done. Can you take a look, please?

IrAlfred:
> @IrAlfred please take a look at the coverage report failing: https://github.com/cypht-org/cypht/actions/runs/19029638869/job/54340356825
> 
> The github action should be updated to match the refactoring you have done. Can you take a look, please?

Yes, I work on it !

PR #1739, opened 2025-10-12
fix(other): fix some GitHub module issues

IrAlfred:
This PR addresses the following issues:
- "Undefined array key 1" errors in GitHub "All" page (?page=message_list&list_path=github_all). The Individual repo pages are in the following format (?page=message_list&list_path=github_owner/repo). The issue arises  when GitHub repositories are not in expected "owner/repo" format.
- "Call to a member function search() on false" when no IMAP server is not configured. The fix allows GitHub module to work independently without IMAP setup


modules/imap/handler_modules.php
```diff
@@ -1317,6 +1317,10 @@ public function process() {
             $folders = array_map(function($ds) { return $ds['folder']; }, $data_sources);
         }
 
+        if (empty($ids)) {
+            return;
```
kroky: This method outputs some data which other modules expect to use. If you return early, that output data will not be available. What is the problem of ids being empty? The code that follows should just work and output some empty/default data.

modules/imap/handler_modules.php
```diff
@@ -1363,6 +1367,9 @@ public function process() {
         foreach ($ids as $key => $id) {
             $details = Hm_IMAP_List::dump($id);
             $mailbox = Hm_IMAP_List::get_connected_mailbox($id, $this->cache);
+            if (!$mailbox) {
```
kroky: This seems like we are masking a deeper problem. Why would the mailbox be not present? If there is a problem connecting or something else, we should send feedback to the user, at least.

IrAlfred: > This seems like we are masking a deeper problem. Why would the mailbox be not present? If there is a problem connecting or something else, we should send feedback to the user, at least.

Let me investigate more deeper about this !

modules/imap/handler_modules.php
```diff
@@ -1394,6 +1394,9 @@ public function process() {
         foreach ($ids as $key => $id) {
             $details = Hm_IMAP_List::dump($id);
             $mailbox = Hm_IMAP_List::get_connected_mailbox($id, $this->cache);
+            if (!$mailbox || !$mailbox->authed()) {
```
kroky, line 1501: If a mailbox is configured but inaccessible at the moment, we should show an error to the user with details about which mailbox is inaccessible.

modules/github/modules.php
```diff
@@ -438,123 +448,274 @@ protected function output() {
...
+
+        $limit = $this->user_config->get('github_limit_setting', DEFAULT_GITHUB_PER_SOURCE);
+        $token = $details['access_token'];
+
+        // Serve cache hits immediately; collect cache misses for parallel fetch
+        $results = array();
+        $to_fetch = array();
+        foreach ($repos as $repo) {
+            $cache_key = 'github_events_' . md5($repo . '_' . $limit);
+            $cached = $this->cache->get($cache_key, false);
```
kroky, line 599: What is the cache invalidation strategy? If we are caching all github event calls, what triggers the cache invalidation, so that we can re-fetch newer events? Especially that we are caching paged responses ordered by some mechanism which means they can differ over time (e.g. if github returns latest 20 events on the first page)...

IrAlfred: Yeah, good catch on the cache invalidation thing. So we cache the first page of repo events with a 5-minute TTL. No event-driven invalidation, just the same TTL-only setup as the **feeds** module. If there's a cache miss or the TTL expires, we just call the GitHub API fresh.

(new changes) For manual refresh, that's the explicit bypass: hit the refresh button, it sends github_force_refresh=1, deletes the cache entry, and fetches right away. Background refreshes still use the cache while it's valid, which matters when github_all fans out across a bunch of repos. So stale data is capped at about 5 minutes in normal use, and users can always force a fresh load with the refresh button.

PR #1740, opened 2025-10-13
feat(backend): add option to exclude auto BCC from messages

Shadow243:
spinoff https://github.com/cypht-org/cypht/pull/1733


kroky:
Looks good, we can merge. Why it is still a draft?

Shadow243:
> Looks good, we can merge. Why it is still a draft?

Sure, it's ready for review.

PR #1741, opened 2025-10-14
fix(frontend): close other dropdowns when one is opened

camilevahviraki:
[When you pick one action with a drop down, it should close the others](https://avan.tech/item145740)

### Issues

- relates : https://avan.tech/item145740

### Current Behavior

[Screencast from 2025-10-14 12-39-11.webm](https://github.com/user-attachments/assets/9ae67837-b26e-4d4a-94c2-bab07b8f64a8)

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -193,17 +193,17 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
             if (!array_key_exists('subject', lc_headers($headers)) || !trim(lc_headers($headers)['subject'])) {
                 $headers['subject'] = $this->trans('[No Subject]');
             }
-            
+
             // Start Bootstrap container
             $txt .= '<div class="container-fluid p-0 ml-0 border-bottom border-secondary-subtle text-muted">';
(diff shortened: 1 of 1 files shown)


modules/imap/site.js
```diff
@@ -1553,3 +1553,13 @@ $(document).on('submit', '#shareForm', function(e) {
         }
     );
 });
+
+document.addEventListener("show.bs.dropdown", function (event) {
```
kroky: This seems like a global handler - it is better to put it in core module that is always loaded.
Also, are we sure we want to close all dropdowns when another one opens. Do we have nested dropdown lists when opening one also opens sub-elements that are also dropdowns?

camilevahviraki: Thanks @kroky , for the feedback!

I’ve moved the global show.bs.dropdown listener to modules/core/site.js so it’s always loaded and available.

I updated the logic to skip nested dropdowns (submenus), while still closing other top-level dropdowns when one is opened.

kroky: ok, thanks!

PR #1748, opened 2025-10-16
feat(other): improve feedback on user creation

IrAlfred:


scripts/create_account.php
```diff
@@ -59,5 +59,13 @@
 }
 
 if ($user && $pass) {
-    $auth->create($user, $pass);
+    $res = $auth->create($user, $pass);
+    switch ($res) {
+        case 1:
+            die("Error: Unable to create user account.\n");
+        case 2:
+            die("User account created successfully.\n");
```
kroky: We shouldn't die when the action is successful. CLI scripts have specific error codes that are very useful when external tools call them. We should exit with 0 success code if everything is good.

PR #1750, opened 2025-10-17
refactor(frontend): maintain a dedicated logout page to separate session termination from navigation handling, preventing related issues

mercihabam:


kroky:
Hey, isn't this a dupe of https://github.com/cypht-org/cypht/pull/1742 ?

mercihabam:
> Hey, isn't this a dupe of #1742 ?

They intend to fix the same issue, but I just checked the current behavior in master, when the user session expires, a `Server error` message is shown instead of a login screen, which I think isn't better.

kroky:
Ok, I have checked this deeper - can we do a more robust fix here? I suggest we use normal browser navigation for the logout action - it should be cleaning up any session, cookie, etc. data like normal logout procedure does and it doesn't make sense to be part of an inline JS navigation - e.g. we face the problem of other ajax calls failing but there can be more than that. I suggest we simply turn the logout functionality into a normal browser visit of the logout page.

mercihabam:
@kroky, those changes involve quite a bit of refactoring. What do you say about keeping the minimal fix in https://github.com/cypht-org/cypht/pull/1750/commits/9f85c4cd695e0e130ceeb95618d019252313d48d, which won't be definitive, and continuing with the changes you suggested once the next release is made and stabilized (something between 2.5.1...2.5.3)?

kroky:
Yes, we don't want a full refactor now. Let's keep this for after the release.

mercihabam:
So, I opened: https://github.com/cypht-org/cypht/pull/1758.

mercihabam:
Hi @kroky, I thought this would be a good time to revisit this, so I’ve worked on getting it ready. Please take a look when you have a chance.

PR #1755, opened 2025-10-21
feat(other): persist Bcc header in sent and draft messages

IrAlfred:
This PR fixes the issue where Bcc recipients were missing from saved email headers.

### DIFF
--- a/modules/imap/output_modules.php
+++ b/modules/imap/output_modules.php
@@ -183,7 +183,7 @@ class Hm_Output_filter_message_headers extends Hm_Output_Module {
       protected function output() {
         if ($this->get('msg_headers')) {
             $txt = '';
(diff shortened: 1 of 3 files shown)


kroky:
Not sure I understand the purpose here. What's the problem of storing a draft with BCC header that makes you use a custom X- header?

IrAlfred:
> Not sure I understand the purpose here. What's the problem of storing a draft with BCC header that makes you use a custom X- header?

I used custom x- header instead of standard bcc header for privacy protection. Standard BCC headers must never appear in the actual email sent to recipients. This actual x- header is needed only for the users to see their own bcc information when viewing drafts and sent items. I tried to follow the pattern existing in cypht for X-Auto-Bcc.

What do you suggest, I may be wrong.

PR #1756, opened 2025-10-22
fix(frontend): ensure that `replace_cursor_position` function is properly called in pageHandler, fixing missing new lines at the beginning of the textarea in reply and forward mode

mercihabam:


marclaporte:
For the record: this is to restore the behavior from Cypht 1.4.x

PR #1761, opened 2025-10-24
fix(backend): ews display attachment filenames for received messages

kambereBr:
1. Properly extract attachment filenames from received messages.
2. Uncaught Error: Call to undefined method Hm_EWS::supports_dsn() in /cypht/modules/smtp/modules.php:584

### DIFF
--- a/modules/imap/hm-ews.php
+++ b/modules/imap/hm-ews.php
@@ -267,8 +267,6 @@ class Hm_EWS {
             $msg = new Type\MessageType();
             $msg->setFrom($from);
             $msg->setToRecipients($recipients);
-
-            $message = $this->use_iso_if_safe($message);
             $mimeContent = Type\MimeContentType::buildFromArray([
                 'CharacterSet' => 'UTF-8',
                 '_' => base64_encode($message)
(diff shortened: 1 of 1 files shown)


modules/imap/hm-ews.php
```diff
@@ -285,6 +297,44 @@ public function send_message($from, $recipients, $message, $delivery_receipt = f
...
     }
 
+    /**
+     * Convert UTF-8 to ISO-8859-1 if the content is safe to do so.
+     * This prevents Exchange from converting quoted-printable to base64 encoding.
+     * 
+     * @param string $message The MIME message from Hm_MIME_Msg
+     * @return string Modified message with ISO-8859-1 where safe
+     */
+    private function use_iso_if_safe($message) {
```
kroky: This whole function seems very hacky and low-level. Why do you need to convert a universally available encoding like UTF-8 to something that doesn't even support more than a few hundreds of characters?
If you experience a specific edge case with Exchange, let's discuss what can be done but this conversion certainly looks very suspicious to do on the client side.

kambereBr: Hi @kroky, thanks for the review.

As discussed in the task thread, since the message sent from Cypht through EWS fully follows the relevant RFCs (5322, 2045–2047, 3676) and aligns with Microsoft’s EWS specification for `<MimeContent>`, I’ve withdrawn the encoding-related changes.

This PR now contains only the changes related to the filename issue.

Thanks!

PR #1772, opened 2025-11-02
fix(module/config): read postgres byte streams before decryption

affinitytechworks:
PostgreSQL can hand `hm_user_settings.settings` back as a stream resource when PDO is built with libpq support. Passing that resource straight into `base64_decode()` causes a PHP 8 fatal error before libsodium gets a chance to decrypt the payload. Normalizing the value with `stream_get_contents()` ensures the stored settings are always a string before decode/decrypt, which allows users to log in again.

### DIFF
--- a/lib/config.php
+++ b/lib/config.php
@@ -357,7 +357,15 @@ class Hm_User_Config_DB extends Hm_Config {
      */
     private function decrypt_settings($data, $key) {
         if (!$this->crypt) {
-            $data = $this->decode($data['settings']);
+            $settings = $data['settings'];
+            if (is_resource($settings)) {
+                $settings = stream_get_contents($settings);
+            }
+            if ($settings === false) {
+                $this->decrypt_failed = true;
+                return false;
+            }
+            $data = $this->decode($settings);
         } else {
             $data = $this->decode(Hm_Crypt::plaintext($data['settings'], $key));
         }
--- a/modules/smtp/hm-smtp.php
+++ b/modules/smtp/hm-smtp.php
@@ -111,13 +111,33 @@ class Hm_SMTP {
         else {
             $this->port = 25;
         }
-        if (isset($conf['tls']) && $conf['tls']) {
-            $this->tls = true;
+        $this->tls = false;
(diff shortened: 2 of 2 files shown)


lib/config.php
```diff
@@ -357,7 +357,15 @@ private function new_settings($username) {
      */
     private function decrypt_settings($data, $key) {
         if (!$this->crypt) {
-            $data = $this->decode($data['settings']);
+            $settings = $data['settings'];
+            if (is_resource($settings)) {
```
kroky, line 361: Can we do this at the place where stream is defined/used? It seems awkward to work with streams here. THis method should receive a string.

modules/smtp/hm-smtp.php
```diff
@@ -111,13 +111,33 @@ function __construct($conf) {
         else {
             $this->port = 25;
         }
-        if (isset($conf['tls']) && $conf['tls']) {
-            $this->tls = true;
-        }
-        else {
-            $this->tls = false;
+        $this->tls = false;
```
kroky, line 114: This seems to be part of another PR (which I already merged). Can you push one change at a time to a PR? Thanks.

marclaporte:
@affinitytechworks  We need you :-)

IrAlfred:
Seems related to https://github.com/cypht-org/cypht/pull/1790

marclaporte:
@affinitytechworks we need you :-)

affinitytechworks:
Hi Marc,Apologies, total zeitnot on my side - am I in critical path for your team? I’ll update/retest cypth by the end of the month. Kind regardsDMSent from my iPhoneOn Jan 14, 2026, at 04:43, Marc Laporte ***@***.***> wrote:﻿marclaporte left a comment (cypht-org/cypht#1772)
@affinitytechworks we need you :-)

marclaporte:
No problem. Please just come back when you can :-)

marclaporte:
@affinitytechworks  Not a particular rush but should be handled at one point :-)

marclaporte:
I converted this PR to draft because of lack of activity

PR #1776, opened 2025-11-04
refactor(backend): expose imap instance

bienvenuushindi:
I need to use Imap object to  be able to use it in Tiki  check this MR https://gitlab.com/tikiwiki/tiki/-/merge_requests/8961/diffs


kroky:
I've replied in the other PR - it is better to use Hm_IMAP directly.

PR #1777, opened 2025-11-05
fix(other): improve user feedback for saved searches to include settings save reminder

IrAlfred:
When a user logout without saving his settings, the saved searches are lost, so we need a reminder

### DIFF
--- a/modules/saved_searches/modules.php
+++ b/modules/saved_searches/modules.php
@@ -49,7 +49,12 @@ class Hm_Handler_update_search extends Hm_Handler_Module {
                 $this->user_config->set('saved_searches', $searches->dump());
                 $this->session->set('user_data', $this->user_config->dump());
                 $this->out('updated_search', true);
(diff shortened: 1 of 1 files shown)


modules/saved_searches/modules.php
```diff
@@ -49,7 +49,12 @@ public function process() {
                 $this->user_config->set('saved_searches', $searches->dump());
                 $this->session->set('user_data', $this->user_config->dump());
                 $this->out('updated_search', true);
-                Hm_Msgs::add('Saved search updated', 'info');
+                if (isPageConfigured('save')) {
+                    Hm_Msgs::add("Saved search updated.  To preserve your searches after logout, please go to <a class='alert-link' href='/?page=save'>Save Settings</a>.", 'info');
```
marclaporte: There are many spaces between two sentences. Intentional?

IrAlfred: > There are many spaces between two sentences. Intentional?

Oh no, it's an error. Let me fix

Issue #1556, opened 2025-06-21
🐛 [Bug] Cannot login when SESSION_TYPE=REDIS


diogovalentte:
## 🐛 Bug
When I set `SESSION_TYPE` to `REDIS` and try to login, it says that my credentials are wrong. Setting it to `DB` works, but it seems cypht doesn't use redis for caching.

I tested and the cypht container can connect to the redis container.

### Version & Environment
Using docker compose:
```yml
services:
  db:
    container_name: cypht-db
    image: mariadb:10
    volumes:
      - /Apps/cypht/db:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=root_password
      - MYSQL_DATABASE=cypht
      - MYSQL_USER=cypht
      - MYSQL_PASSWORD=cypht_password
    restart: unless-stopped

  redis:
    container_name: cypht-redis
    image: redis:7-alpine
    volumes:
      - /Apps/cypht/redis/data:/data
    restart: unless-stopped

  cypht:
    container_name: cypht
    image: cypht/cypht:2.4.0
    volumes:
      - /Apps/cypht/users:/var/lib/hm3/users
    ports:
      - 46185:80
    environment:
      - DEBUG_LOG=true
      - AUTH_USERNAME=${AUTH_USERNAME}
      - AUTH_PASSWORD=${AUTH_PASSWORD}
      - DB_CONNECTION_TYPE=host
      - DB_DRIVER=mysql
      - DB_HOST=db
      - DB_NAME=cypht
      - DB_USER=cypht
      - DB_PASS=cypht_password
      - DB_SOCKET=/var/lib/mysqld/mysqld.sock
      - USER_CONFIG_TYPE=DB
      - SESSION_TYPE=REDIS
      - AUTH_TYPE=DB
      - ENABLE_REDIS=true
      - REDIS_SERVER=redis
      - REDIS_PORT=6379
      - REDIS_INDEX=1
      - REDIS_SOCKET=
      - GMAIL_CLIENT_ID=${GMAIL_CLIENT_ID}
      - GMAIL_CLIENT_SECRET=${GMAIL_CLIENT_SECRET}
      - GMAIL_CLIENT_URI=${GMAIL_CLIENT_URI}
    restart: unless-stopped
```

### Logs
```
Attempting to connect to database ... (1/10)

Attempting to connect to database ... (2/10)

Database connection successful ...

Creating database table hm_user ...

[...]

PR #1779, opened 2025-11-06
fix(other): apply cache-first strategy to accelerate folder browsing

IrAlfred:
https://github.com/cypht-org/cypht/issues/1556

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -650,16 +650,36 @@ class Hm_Handler_imap_folder_expand extends Hm_Handler_Module {
                 $folder = $this->request->post['folder'];
             }
             $path = sprintf("imap_%s_%s", $form['imap_server_id'], $folder);
-            $page_cache = $this->cache->get('imap_folders_'.$path);
+            $with_subscription = isset($this->request->post['subscription_state']) && $this->request->post['subscription_state'];
+
             if (array_key_exists('imap_prefetch', $this->request->post)) {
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -56,7 +56,7 @@ REDIS_SERVER='127.0.0.1'
 REDIS_PORT=6379
 REDIS_INDEX=1
 REDIS_PASS=
-REDIS_SOCKET=/var/run/redis/redis-server.sock
+REDIS_SOCKET=
```
kroky: Can you leave the default here - it is an example, after all.

modules/imap/handler_modules.php
```diff
@@ -650,13 +650,25 @@ public function process() {
...
+                $this->out('imap_expanded_folder_id', $form['imap_server_id']);
+                $this->out('imap_expanded_folder_path', $path);
+                $this->out('with_input', $with_subscription);
+                $this->out('folder', $folder);
+                return;
+            }
+
             $mailbox = Hm_IMAP_List::get_connected_mailbox($form['imap_server_id'], $this->cache);
             if ($mailbox && $mailbox->authed()) {
                 $this->out('can_share_folders', stripos($mailbox->get_capability(), 'ACL') !== false);
```
kroky: The proposed approach will skip sending 'can_share_folders' and quota variables which might result in unexpected behavior. Can you make sure you cache this data as well, when storing the imap_folders cache and then reuse it here, so indeed, you can skip authenticating to each server but still use values that were cached?

Issue #1501, opened 2025-05-01
🚀 [Feature] Add Redis Service to Docker Compose Files


Revisto:
## 🚀 Feature
Add a Redis service definition to the development (`docker-compose.dev.yaml`) and production example (`docker/docker-compose.yaml`) Docker Compose files to simplify setup for development and document the option for production.

marclaporte:
So the main benefit is to improve performance?

What type of gains could we expect?

Revisto:
For me, when I was working on emoji reactions, I added new functions to fetch data from IMAP servers and needed to make sure they worked properly with cached data. Without testing with Redis, it would have caused errors in the `get_message_parts_content` function (which relies on `search_bodystructure`) when the data came from cache. Having Redis made it possible to test this properly and ensure everything behaves as expected with cached results, without errors.

IrAlfred:
[https://github.com/cypht-org/cypht/wiki/Cypht%E2%80%90Redis-setup](https://github.com/cypht-org/cypht/wiki/Cypht%E2%80%90Redis-setup)

PR #1783, opened 2025-11-10
feat(other): add redis service to Docker Compose Files

IrAlfred:
Related to : https://github.com/cypht-org/cypht/issues/1501

### DIFF
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -14,12 +14,32 @@ services:
       - MYSQL_USER=cypht
       - MYSQL_PASSWORD=cypht_password
     stop_grace_period: 30s
+
+  redis:
+    image: redis:7-alpine
+    ports:
+      - "6379:6379"
+    volumes:
+      - ./data/redis:/data
+    command: redis-server --appendonly yes
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+      interval: 10s
+      timeout: 3s
+      retries: 5
+    stop_grace_period: 30s
+
   cypht:
     build:
       context: .
       dockerfile: ./docker/Dockerfile
       args:
         WITH_DEBUG: true
+    depends_on:
+      db:
+        condition: service_started
+      redis:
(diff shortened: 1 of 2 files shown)


docker-compose.dev.yaml
```diff
@@ -14,12 +14,32 @@ services:
...
       - MYSQL_PASSWORD=cypht_password
     stop_grace_period: 30s
+
+  redis:
+    image: redis:7-alpine
+    ports:
+      - "6379:6379"
+    volumes:
+      - ./data/redis:/data
+    command: redis-server --appendonly yes
```
kroky, line 24: Why appendonly?

kroky:
I am not sure we want to force redis for all docker deployments. Maybe we can suggest using it, add an example and let user decide if they want to?

docker-compose.dev.yaml
IrAlfred: For data persistence between container restarts

IrAlfred:
> I am not sure we want to force redis for all docker deployments. Maybe we can suggest using it, add an example and let user decide if they want to?

Noted, I will include this in the documentation

docker-compose.dev.yaml
kroky: Isn't a volume used for data persistence? Appendonly seems like a totally different mode of operation for redis.

IrAlfred: > Isn't a volume used for data persistence? Appendonly seems like a totally different mode of operation for redis.

Without --appendonly Redis would run in-memory only, losing all cache data if the container restarts. Of course for data persistence we use volume and combined with the --appendonly yes option  the data are logged in a file on that volume. Both work together to ensure data survives container restarts.

IrAlfred: But we could simplify by removing persistence since we're not caching critical data

kroky: Yes, I think for our use-case where we use it for caching, we can skip redis persistence.

Issue #1544, opened 2025-06-11
🐛 [Bug] USER_CONFIG_TYPE=BD cause 500 Error


ljcbaby:
## 🐛 Bug

it seems config decode error.

logs here
```bash
NOTICE: PHP message: PHP Warning:  Undefined variable $config in /usr/local/share/cypht/site/index.php on line 57
NOTICE: PHP message: PHP Stack trace:
NOTICE: PHP message: PHP   1. {main}() /usr/local/share/cypht/site/index.php:0
127.0.0.1 -  11/Jun/2025:22:40:44 +0800 "GET /index.php" 200
NOTICE: PHP message: Array
(
    [0] => INFO: Using Hm_PHP_Session with Hm_Auth_DB
    [1] => INFO: Using DB user configuration
    [2] => WARNING: Already registered module for ajax_combined_message_list re-attempted: load_user_data
    [3] => WARNING: Already registered module for ajax_imap_debug re-attempted: profile_data
    [4] => WARNING: Already registered module for servers re-attempted: compose_profile_data
    [5] => INFO: Using sapi: fpm-fcgi
    [6] => INFO: Request type: HTTP
    [7] => INFO: Request path: /
    [8] => INFO: TLS request: 0
    [9] => INFO: Mobile request: 0
    [10] => INFO: Page ID: home
    [11] => INFO: Using Hm_Cache for cache
    [12] => DANGER: Redis connect failed
    [13] => INFO: CACHE backend using: memcache
    [14] => WARNING: TRANSLATION NOT FOUND :Cypht:
    [15] => WARNING: TRANSLATION NOT FOUND :Are you sure you want to delete this server? :
    [16] => WARNING: TRANSLATION NOT FOUND :You need to have Javascript enabled to use %s, sorry about that!:
    [17] => INFO: PHP version 8.1.32
    [18] => INFO: Zend version 4.1.32
    [19] => INFO: Peak Memory: 16384
    [20] => INFO: PID: 1874
    [21] => INFO: Included files: 147
)
192.168.2.8 - - [11/Jun/2025:22:40:44 +0800] "GET / HTTP/1.1" 200 167633 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36 Edg/137.0.0.0"
127.0.0.1 -  11/Jun/2025:22:40:47 +0800 "POST /index.php" 500
[...]

PR #1790, opened 2025-11-12
fix(other): fix PostgreSQL BYTEA resource handling in user config loading

IrAlfred:
When using PostgreSQL with USER_CONFIG_TYPE=DB, the settings column
(BYTEA type) is returned by PDO as a resource stream rather than a
string. This caused a fatal error when the resource was passed to
base64_decode() in the decryption process.

Exact error
```
NOTICE: PHP message: PHP Fatal error:  Uncaught TypeError: base64_decode(): Argument #1 ($string) must be of type string, resource given in /usr/local/share/cypht/lib/crypt_sodium.php:28
127.0.0.1 -  12/Nov/2025:09:02:23 +0800 "POST /index.php" 500
Stack trace:
#0 /usr/local/share/cypht/lib/crypt_sodium.php(28): base64_decode()
#1 /usr/local/share/cypht/lib/config.php(362): Hm_Crypt::plaintext()
#2 /usr/local/share/cypht/lib/config.php(387): Hm_User_Config_DB->decrypt_settings()
#3 /usr/local/share/cypht/modules/core/handler_modules.php(719): Hm_User_Config_DB->load()
#4 /usr/local/share/cypht/lib/modules_exec.php(177): Hm_Handler_load_user_data->process()
#5 /usr/local/share/cypht/lib/modules_exec.php(155): Hm_Module_Exec->run_handler_module()
#6 /usr/local/share/cypht/lib/dispatch.php(218): Hm_Module_Exec->run_handler_modules()
#7 /usr/local/share/cypht/lib/dispatch.php(188): Hm_Dispatch->process_request()
#8 /usr/local/share/cypht/site/index.php(55): Hm_Dispatch->__construct()
#9 {main}
  thrown in /usr/local/share/cypht/lib/crypt_sodium.php on line 28
```

Related issue: https://github.com/cypht-org/cypht/issues/1544

### DIFF
--- a/lib/db.php
+++ b/lib/db.php
@@ -80,6 +80,24 @@ class Hm_DB {
     }
 
     /**
+     * Convert any resource values in a database row to strings
+     * Ensures consistent scalar return types across different database engines
+     * @param mixed $row database row or query result
+     * @return mixed row with any resource values converted to strings
+     */
+    static private function convert_resources_to_strings($row) {
+        if (!is_array($row)) {
+            return $row;
+        }
+        foreach ($row as $key => $value) {
+            if (is_resource($value)) {
+                $row[$key] = stream_get_contents($value);
+            }
+        }
+        return $row;
+    }
+
+    /**
      * @param object|false $dbh PDO connection object
      * @param string $sql sql with placeholders to execute
      * @param array $args values to insert into the sql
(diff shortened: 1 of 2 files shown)


lib/config.php
```diff
@@ -384,6 +384,10 @@ public function load($username, $key) {
         if (!$data || !array_key_exists('settings', $data)) {
             return $this->new_settings($username);
         }
+
+        if (is_resource($data['settings'])) {
```
kroky: Fix this in Hm_DB class - we should have a stable interface there returning scalar values for db queries rather than streams regardless of underlying db engine.

PR #1792, opened 2025-11-12
refactor(frontend):  harmonize special folders selection on the advanced search page, allowing to select all folders or each individually

mercihabam:
Screenshot 2025-11-12 at 22 19 40

Screenshot 2025-11-13 at 17 16 54


modules/advanced_search/js_modules/route_handlers.js
```diff
@@ -62,4 +62,77 @@ function applyAdvancedSearchPageHandlers() {
...
+    $('body').on("click", ".pick_special_folders", function(e) {
+        e.preventDefault();
+
+        const modal = new Hm_Modal({
+            modalId: 'pick_special_folders_modal',
+            title: 'Pick Special Folders',
+            size: 'lg',
+        });
+
+        const serverId = $(this).closest('li').attr("class").split('_')[1];
```
kroky: Can you please use data attributes for something holding JS or business logic?

modules/advanced_search/js_modules/route_handlers.js
```diff
@@ -62,4 +62,77 @@ function applyAdvancedSearchPageHandlers() {
...
+        const specialFolders = [];
+        ['archive', 'draft', 'junk', 'sent', 'trash'].forEach(folderType => {
+            const folderName = hm_special_folders()[serverId][folderType];
```
kroky: Please guard against TypeError here where you access nested level elements.

modules/advanced_search/js_modules/route_handlers.js
```diff
@@ -62,4 +62,77 @@ function applyAdvancedSearchPageHandlers() {
         }
     }
     Hm_Message_List.check_empty_list();
+
+    $('body').on("click", ".pick_special_folders", function(e) {
```
kroky, line 66: Doesn't it need 'off' as well?

modules/advanced_search/js_modules/route_handlers.js
```diff
@@ -62,4 +62,77 @@ function applyAdvancedSearchPageHandlers() {
...
+        const accountId = account.data('target');
+
+        const isFolderSelected = (folder) => get_adv_sources().find(source => source.label === getFolderLabel(folder));
+
+        const getFolderLabel = (folder) => accountLabel + ' > ' + folder;
+
+        modal.setContent(`
+        <div class="d-flex gap-3 flex-wrap">
+            <div class="form-check form-switch">
+                <input class="form-check-input" type="checkbox" id="all">
```
kroky: Please use a bit more specific identifier here - `all` might conflict with another one in current or future version of Cypht or Tiki.

modules/advanced_search/modules.php
```diff
@@ -176,7 +159,13 @@ private function validate_source($val) {
         }
         $source_parts = explode('_', $val);
         $this->imap_id = $source_parts[1];
-        $this->folder = hex2bin($source_parts[2]);
+
+        $folder = $source_parts[2];
+        if (ctype_xdigit($folder)) {
```
kroky: I'd rather do the conversion here unconditionally - either always do hex2bin or never do it. Otherwise, we risk introducing bugs. ctype_xdigit returns true for numeric only strings which might be actual folder IDs not hex encoded. If you face a bug where the folder here is not hex encoded - please trace it to the proper place and fix that. I think Cypht expects the folder ids to be hex encoded everywhere it touches the frontend/UI.

modules/advanced_search/js_modules/route_handlers.js
mercihabam: Sure! 'off' is done down on line 136.

Issue #1143, opened 2024-08-02
Indicate new releases in Cypht, to alert end users that an upgrade is available


marclaporte:
Cypht showed a list of latest commits, which has now been removed: https://github.com/cypht-org/cypht/pull/1138 because it was a little weird to show non-devs a list of commits.

But showing that a new version is available would be relevant.

PR #1799, opened 2025-11-21
feat(frontend/backend): alert end users when potential upgrades are detected

mercihabam:
Screenshot 2025-11-23 at 17 45 40

# A small note

This will start functioning from the next release onward. So users who are using Cypht 2.5.2 while 2.5.3 has already been released will be notified about the available upgrade.

~~This fully depends on https://github.com/cypht-org/cypht/pull/1800~~, and will require consistently updating the `CYPHT_VERSION` defined [here](https://github.com/cypht-org/cypht/pull/1799/files#diff-36ac85783267877ba06b979d66deb816f6f5b5dfdd6107f10d121092f7cc89b7R3) to the next upcoming release version after each release is made.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /site.js
 /site.css
 /site/
+/fonts/
 /todo
 /tests/selenium/creds.py
 /tests/selenium/local_creds.py
--- a/scripts/config_gen.php
+++ b/scripts/config_gen.php
@@ -318,14 +318,24 @@ function write_config_file($settings, $filters) {
  * @return void
  */
 function append_bootstrap_icons_files() {
+    // Ensure target font directories exist for both deployment modes:
+    // - Running from site/: fonts are expected under site/fonts relative to site/site.css
+    // - Running from the main directory: fonts are expected under fonts relative to site.css
     if (!is_dir("site/fonts")) {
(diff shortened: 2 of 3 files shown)


IrAlfred:
Related issue: https://github.com/cypht-org/cypht/issues/1143

lib/version.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+define('CYPHT_VERSION', '2.5.2');
+
+$releases = json_decode(file_get_contents('https://github.com/cypht-org/cypht/blob/master/releases.json?raw=true'), true);
```
kroky: Since you are fetching github already, why not using the data available there?
https://api.github.com/repos/cypht-org/cypht/releases

mercihabam: Thanks! Did not know that source was available.

releases.json
```diff
@@ -0,0 +1,39 @@
+{
```
kroky: Please remove this file as it is no longer used.

marclaporte:
Cypht master produces an incorrect report:

Baraka24:
Relates to https://github.com/cypht-org/cypht/issues/1838

mercihabam:
I opened: https://github.com/cypht-org/cypht/pull/1849

PR #1800, opened 2025-11-23
chore(backend): gather all releases in a json file

mercihabam:
This will allow the implementation in https://github.com/cypht-org/cypht/pull/1799 to have a source from which to retrieve the released-versions information. Ideally, this file should be updated whenever a new version is released in the future.


kroky:
I prefer we use github releases data directly.

mercihabam:
Great! So, closing this PR in favor of data fetched from GitHub.

Issue #1532, opened 2025-05-31
🐛 [Bug] Sample YAML/CSV bulk import file links are incorrect and sample files don't seem to exist


knightsg:
## 🐛 Bug
I downloaded the sample YAML bulk account import file from my Cypht installation but the file it downloaded is actually a block of HTML content with what appeared to be unicode characters embedded. I attempted to open it in multiple web browsers but it won't display the HTML and just triggers the browser's file download feature. I tried the CSV sample as well and got the same thing. However I added it to a gist which formatted it correctly and found that it was just the content of the Cypht server settings page.

On further investigation I realised that the link for both downloads is like `https://<mail-server-host>/usr/local/cypht/assets/data/server_accounts_sample.yaml`. I'm running cypht from Docker so I checked in the container but I couldn't find this file. Obviously the linked URL is incorrect but I thought if I could find the file I could try modifying the URL to the correct path.

### Version & Environment
Rev: `daily (3e5864dea5f8)`
OS:  `Ubuntu 24.04.2 LTS/docker-ce 28.1.1-1~ubuntu.24.04~noble`

Shadow243:
Big thanks to @knightsg  for reporting this issue and helping improve the project!

Shadow243:
@knightsg the error has been fixed. You can confirm: it will be available in the daily image starting with the next daily release.

knightsg:
Thanks, I'll take a look sometime in the next day or two and confirm.

Baraka24:
@knightsg have you tested? If it works, please close this.

marclaporte:
@knightsg We are stabilizing master, from which we will release version 2.5.0.

knightsg:
Apologies, life got in the way. I'll do my best to verify by the end of this week 😬

knightsg:
I updated my docker image to the latest daily image (`sha256:b482cd2583cb30db885aa5f0cd4f798d515eaf4b1a40a348102b23819efd307f` created at `2025-09-02T00:59:19+0000`) but unfortunately I'm still seeing the same issue here. The link to download the bulk account YAML is still `https://<mail-server-host>/usr/local/cypht/assets/data/server_accounts_sample.yaml` and the downloaded file content is still just the Cypht server settings page.

Shadow243:
[...]

PR #1801, opened 2025-11-24
fix(other): fix asset download links when Cypht is installed outside web root

IrAlfred:
When Cypht is installed in a non-web-accessible directory (e.g., `/usr/local/share/cypht`) for security purposes, asset download links (such as server account sample files) fail with 404 errors. 

The issue occurs because `ASSETS_PATH` is defined using the absolute filesystem path (`APP_PATH.'assets/'`), which results in invalid URLs like:

`http://localhost/usr/local/share/cypht/assets/data/server_accounts_sample.yaml`

This works fine when Cypht is installed in the document root.

Related issue: https://github.com/cypht-org/cypht/issues/1532

### DIFF
--- a/scripts/config_gen.php
+++ b/scripts/config_gen.php
@@ -372,8 +372,15 @@ function create_production_site($assets, $settings, $hashes) {
     copy('site.js', 'site/site.js');
     append_bootstrap_icons_files();
 
+    // Copy main assets directory
+    if (is_readable('assets/')) {
+        printf("copying assets directory...\n");
+        copy_recursive('assets');
+    }
+
     $index_file = file_get_contents('index.php');
     $index_file = preg_replace("/APP_PATH', ''/", "APP_PATH', '".APP_PATH."'", $index_file);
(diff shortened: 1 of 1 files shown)


kroky:
I'd rather simplify this and just move the whole assets/data directory to our "public" facing directory called "site". I assume people pointed the webroot to site dir and that's why it was inaccessible. If we move the assets to the site dir, we should be fine. I suggest you use a little bit different name to make things clear, like site/sample_data/server_accounts_sample.*

IrAlfred:
> I'd rather simplify this and just move the whole assets/data directory to our "public" facing directory called "site". I assume people pointed the webroot to site dir and that's why it was inaccessible. If we move the assets to the site dir, we should be fine. I suggest you use a little bit different name to make things clear, like site/sample_data/server_accounts_sample.*

Understood. The site/ directory is generated by the php ./scripts/config_gen.php command and is not tracked in git, which is why I handled it that way.

kroky:
Actually, that makes sense. We can keep it this way for now and when we start to have more version-controlled assets, we can keep the site dir in the repo and some of the data there be static.

PR #1802, opened 2025-11-27
fix(other): fix Bootstrap Icons not visible in production mode (DEBUG=false)

IrAlfred:
## Issue
Bootstrap Icons are not displaying when running Cypht in production mode (`DEBUG=false`). Icons worked correctly in debug mode but were missing in production.

When combining CSS files for production, the Bootstrap Icons CSS was included in `site.css`, but the font file paths remained as relative paths (`./fonts/`) that were correct for the original vendor location but incorrect for the combined production CSS file location. The fonts are copied to `site/fonts/` during the build process, but the CSS was still referencing `./fonts/`, causing 404 errors for the font files.

### DIFF
--- a/scripts/config_gen.php
+++ b/scripts/config_gen.php
@@ -318,14 +318,24 @@ function write_config_file($settings, $filters) {
  * @return void
  */
 function append_bootstrap_icons_files() {
+    // Ensure target font directories exist for both deployment modes:
+    // - Running from site/: fonts are expected under site/fonts relative to site/site.css
+    // - Running from the main directory: fonts are expected under fonts relative to site.css
     if (!is_dir("site/fonts")) {
         mkdir('site/fonts', 0755);
+    }
+    if (!is_dir("fonts")) {
+        mkdir('fonts', 0755);
     }
     $source_folder = VENDOR_PATH.'twbs/bootstrap-icons/font/fonts/';
     $files = glob("$source_folder*.*");
     foreach($files as $file){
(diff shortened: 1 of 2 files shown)


scripts/config_gen.php
```diff
@@ -257,6 +257,7 @@ function combine_includes($js, $js_compress, $css, $css_compress, $settings) {
     $css_hash = '';
     if ($css) {
         $css_out = file_get_contents(VENDOR_PATH . "twbs/bootstrap-icons/font/bootstrap-icons.css");
+        $css_out = preg_replace('/url\("\.\/fonts\//', 'url("site/fonts/', $css_out);
```
kroky: site.css is copied to site/site.css later in the same script. Some people run cypht from site/ while others from main directory. That's really confusing and we should disabled it one day but until then, we have to take care of the consequences. I think here, we need distinct versions of the site.css depending on the location if you are going to use relative paths to the fonts.

IrAlfred: > site.css is copied to site/site.css later in the same script. Some people run cypht from site/ while others from main directory. That's really confusing and we should disabled it one day but until then, we have to take care of the consequences. I think here, we need distinct versions of the site.css depending on the location if you are going to use relative paths to the fonts.

I agree with you! However, I found that instead of creating separate versions of site.css, it would be much better to generate the fonts/ directory at the root level.

This would ensure that regardless of whether you run Cypht from the main directory or from the site/ subdirectory, the relative path ../fonts/ would consistently point to a valid location, eliminating the need for duplicate CSS files. This won't be tracked by Git too.

PR #1805, opened 2025-11-28
fix(backend): handle undefined array keys username and password

Baraka24:
### Issues

- Two issues:
```
[26-Nov-2025 12:31:03 UTC] PHP Warning:  Undefined array key "username" in /var/www/html/cypht/modules/imap/hm-ews.php on line 45
[26-Nov-2025 12:31:03 UTC] PHP Warning:  Undefined array key "password" in /var/www/html/cypht/modules/imap/hm-ews.php on line 45
```

### Todo

- Lgn in into your Cypht
- Click your EWS account Inbx(should request _ajax_imap_folder_display_ request)
- Check your logs according to your configuration.

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -559,7 +559,7 @@ class Hm_Handler_imap_message_list_type extends Hm_Handler_Module {
                     } else {
                         $folder = hex2bin($parts[2]);
                     }
-                    
+
                     $mailbox = Hm_IMAP_List::get_mailbox_without_connection($details);
                     $label = $mailbox->get_folder_name($folder);
                     if(!$label) {
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -52,7 +52,7 @@ class Hm_IMAP_List {
         if (self::$use_cache && $cache && is_array($cache)) {
             self::$server_list[$id]['object']->get_connection()->load_cache($cache, 'array');
         }
-        
+
         return self::$server_list[$id]['object']->connect();
     }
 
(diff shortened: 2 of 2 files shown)


modules/imap/hm-ews.php
```diff
@@ -42,7 +42,12 @@ class Hm_EWS {
 
     public function connect(array $config) {
         try {
-            $this->ews = ExchangeWebServices::fromUsernameAndPassword($config['server'], $config['username'], $config['password'], ['version' => ExchangeWebServices::VERSION_2016, 'trace' => 1]);
+            $password = $config['password'] ?? '';
```
kroky: Why would the config be in such an unexpected format? Isn't it written by cypht? Is it probably an old version of cypht that is not properly upgraded? We should find out and keep the config file properly encoded - i.e. servers array having the right keys...

modules/imap/handler_modules.php
```diff
@@ -559,7 +559,9 @@ public function process() {
                     } else {
                         $folder = hex2bin($parts[2]);
                     }
-                    
+
+                    $details['password'] =  "";
+                    $details['username'] = $details['user'];
```
kroky: This is still not the correct place to fix it. Here, we are reading already saved details. Who is writing these details and why they are written in this way? You should find out this and fix there...

Baraka24: As the  `Hm_IMAP_List::dump()`   method stores server credentials using `user/pass `keys, but `get_mailbox_without_connection()` expects `username/password` keys. By adding the true parameter to get full server details and mapping the keys right before the mailbox call, we're fixing the mismatch at the precise location where it occurs.  We expected this to be better than changing the storage format (which would break existing functionality) or modifying the mailbox method (which would require updating all callers).

kroky: if user/pass are expected "almost" everywhere, change get_mailbox_without_connection to use those as well. Otherwise, use username/password. At any  rate, we should be using one and the same storage format. If something is using a different one - refactor to use the same one. That's the goal here.

Baraka24: You're right. The ge`t_mailbox_without_connection` method already handles this correctly by checking for `user`
 key and mapping it to `username` internally, so we just need to ensure we call dump() with `true` parameter to get the credentials and let the method handle the key standardization automatically.
Thanks for the review @kroky .

kroky: Nice catch, thanks!

Issue #1020, opened 2024-05-14
Unraid Docker Compose access denied for user 'cypht'@'172.17.0.1'


ansonryan:
## 💬 Question

During the installation through Unraid compose, the log reported the following warnings:

> Warning: Use of undefined constant DB - assumed 'DB' (this will throw an Error in a future version of PHP) in /tmp/cypht_setup_database.php on line 3
> 
> Warning: Use of undefined constant DB - assumed 'DB' (this will throw an Error in a future version of PHP) in /tmp/cypht_setup_database.php on line 4
> 
> Warning: Use of undefined constant file - assumed 'file' (this will throw an Error in a future version of PHP) in /tmp/cypht_setup_database.php on line 5
> 
> Warning: Use of undefined constant mysql - assumed 'mysql' (this will throw an Error in a future version of PHP) in /tmp/cypht_setup_database.php on line 6
> Waiting for database connection ... (SQLSTATE[HY000] [1045] Access denied for user 'cypht'@'172.17.0.1' (using password: YES))
> 

My configuration on Unraid compose is as follows:

```
version: '3'
services:
  cypht:
    image: sailfrog/cypht-docker:latest
    container_name: cypht
    network_mode: bridge
    restart: always
    #privileged: true
    volumes:
      - /mnt/user/Personal/cypht:/var/lib/hm3/users
    ports:
      - "8280:80"
    environment:
      - CYPHT_AUTH_USERNAME=admin
      - CYPHT_AUTH_PASSWORD=admin_password
      - CYPHT_DB_CONNECTION_TYPE=host
      - CYPHT_DB_HOST=192.168.3.10
      - CYPHT_DB_DRIVER=mysql
      - CYPHT_DB_NAME=cypht
      - CYPHT_DB_USER=cypht
      - CYPHT_DB_PASS=Z
      - CYPHT_SESSION_TYPE=DB
      - CYPHT_DEFAULT_SETTING_TIMEZONE=Asia/Shanghai
      - CYPHT_MODULE_POP3=disable
      - CYPHT_MODULE_IMAP=enable
    labels:
      net.unraid.docker.webui: https://mail.domain.com/
      net.unraid.docker.icon: https://raw.githubusercontent.com/cypht-org/cypht/6d0e50a1778cd63a77130e44f4b46bb9c2a1cb82/modules/core/assets/images/logo.svg
      traefik.enable: true
      traefik.http.routers.cypht.rule: Host(`mail.domain.com`)
[...]

PR #1806, opened 2025-12-03
fix(other): resolve MySQL access denied in container networking with init script

IrAlfred:
## 🍰 Pullrequest
This PR fixes `Access denied for user 'cypht'@'<ip>'` errors when running Cypht with MySQL/MariaDB in Docker by ensuring the application DB user is created with the correct host permissions for container networking.

### Issue
Related to https://github.com/cypht-org/cypht/issues/1020

### DIFF
--- a/docker/docker-compose.yaml
+++ b/docker/docker-compose.yaml
@@ -7,6 +7,8 @@ services:
       - "3306:3306"
     volumes:
       - ./data/mysql:/var/lib/mysql
+      # Mount the initialization script to create user with Docker network permissions
+      - ./init-db.sh:/docker-entrypoint-initdb.d/01-init-user.sh:ro
     environment:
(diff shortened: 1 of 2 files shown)


kroky:
see https://github.com/MariaDB/mariadb-docker/blob/5ad00a82b5bb2705825c7afb670f6547b1bed316/10.11/docker-entrypoint.sh#L442

Passing MARIADB_USER and MARIADB_PASSOWRD + MARIADB_DATABASE should give enough permissions for the user to access the database regardless of the docker container. @% host is usually enough. If we can use something from the official dockerfile or entrypoint, let's use it instead of creating our own.

IrAlfred:
There were some errors in the docker-compose.yml file. See related comment https://github.com/cypht-org/cypht/issues/1020#issuecomment-4097179020. So I'm closing this pull request

PR #1816, opened 2025-12-28
fix(other): allow sending emails with only Cc/Bcc recipients

IrAlfred:
## 🍰 Pullrequest

This PR allows users to send emails with only Cc or Bcc recipients, removing the requirement for the To field to be filled. This aligns with email standards (RFC 5322) where a message is valid as long as it has at least one recipient in any field (To, Cc, or Bcc).

### Technical Details

- The existing validation logic already checks for at least one valid recipient across To, Cc, and Bcc 
- Error message "No valid recipients found" will still be shown if all recipient fields are empty

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -678,7 +678,7 @@ class Hm_Handler_process_compose_form_submit extends Hm_Handler_Module {
         }
 
         /* missing field */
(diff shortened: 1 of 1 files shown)


marclaporte:
I see you updated https://github.com/cypht-org/cypht/wiki/Comprehensive-review-of-to-cc-bcc-when-reply-reply%E2%80%90to%E2%80%90all-forward-in-all-possible-contexts

Well done!

PR #1817, opened 2025-12-28
fix(other): email validation require fqdn tld

IrAlfred:
### Issue
- Email addresses without proper TLD were accepted (e.g., `alfred@test`, `patrick@example`)
- Valid RFC 5322 formats were rejected (IPv6 addresses, quoted strings, internationalized domains)

### Changes: Added RFC 5322 compliant validation
- Client-side: `modules/core/site.js`, `modules/smtp/site.js`
- Server-side: `modules/core/functions.php`
- Support for domain literals (IPv4/IPv6), quoted local-parts, IDN, group syntax
- Enforces minimum 2 domain labels (fixes TLD issue)

### DIFF
--- a/modules/contacts/js_modules/route_handlers.js
+++ b/modules/contacts/js_modules/route_handlers.js
@@ -3,8 +3,8 @@ function applyContactsPageHandlers() {
     $('.add_contact_form').on('submit', function(e) {
         var emailField = $('#contact_email');
         var email = emailField.val();
-        
-        if (email && Hm_Utils.is_valid_email(email)) {
+
+        if (email && !Hm_Utils.is_valid_email(email)) {
             e.preventDefault();
             e.stopPropagation();
             emailField.focus();
@@ -17,14 +17,14 @@ function applyContactsPageHandlers() {
     // Real-time validation feedback on email field
     $('#contact_email').on('blur', function() {
         var email = $(this).val();
+        // Always remove existing error messages first
+        $(this).siblings('.invalid-feedback').remove();
+
         if (email && !Hm_Utils.is_valid_email(email)) {
             $(this).addClass('is-invalid');
-            if ($(this).next('.invalid-feedback').length === 0) {
-                $(this).after('<div class="invalid-feedback">' + hm_trans('Please enter a valid email address with a proper domain (e.g., user@example.com)') + '</div>');
-            }
+            $(this).after('<div class="invalid-feedback">' + hm_trans('Please enter a valid email address with a proper domain (e.g., user@example.com)') + '</div>');
         } else {
             $(this).removeClass('is-invalid');
-            $(this).next('.invalid-feedback').remove();
         }
     });
 
@@ -32,7 +32,7 @@ function applyContactsPageHandlers() {
     $('#contact_email').on('input', function() {
         if ($(this).hasClass('is-invalid')) {
             $(this).removeClass('is-invalid');
-            $(this).next('.invalid-feedback').remove();
+            $(this).siblings('.invalid-feedback').remove();
         }
     });
 
--- a/modules/ldap_contacts/site.js
+++ b/modules/ldap_contacts/site.js
@@ -128,15 +128,15 @@ $(function() {
 
     $('#ldap_mail').on('blur', function() {
         var email = $(this).val();
+        // Always remove existing error messages first
+        $(this).siblings('.invalid-feedback').remove();
+
         if (email) {
             if (!Hm_Utils.is_valid_email(email)) {
                 $(this).addClass('is-invalid');
-                if ($(this).next('.invalid-feedback').length === 0) {
-                    $(this).after('<div class="invalid-feedback">' + hm_trans('Please enter a valid email address with a proper domain (e.g., user@example.com)') + '</div>');
-                }
+                $(this).after('<div class="invalid-feedback">' + hm_trans('Please enter a valid email address with a proper domain (e.g., user@example.com)') + '</div>');
             } else {
                 $(this).removeClass('is-invalid');
-                $(this).next('.invalid-feedback').remove();
             }
         }
     });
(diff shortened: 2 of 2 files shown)


modules/smtp/site.js
```diff
@@ -336,7 +336,8 @@ var copy_text_to_clipboard = function(e) {
 }
 
 var is_valid_recipient = function(recipient) {
-    var valid_regex = /^[\p{L}|\d' ]*(<)?[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*(>)?$/u;
+    // Require FQDN with TLD (at least 2 characters after the last dot)
+    var valid_regex = /^[\p{L}|\d' ]*(<)?[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*\.[a-zA-Z]{2,}(>)?$/u;
```
kroky: can we use one and the same string for all those email regexes you do in this PR? There are at least 3-4 repeating regex strings.

modules/ldap_contacts/site.js
```diff
@@ -109,4 +109,45 @@ $(function() {
     }
 
     $(document).ready(enhanceLdapContacts);
+
+    // Validate LDAP contact email on form submit
+    $('.add_contact_form').on('submit', function(e) {
+        var emailField = $('#ldap_mail');
+        if (emailField.length && emailField.val()) {
+            var email = emailField.val().trim();
+            var emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*\.[a-zA-Z]{2,}$/;
```
kroky: I still see repeated email regexes at least on 4 distinct places.

modules/contacts/js_modules/route_handlers.js
```diff
@@ -1,4 +1,53 @@
+// Validate email address with FQDN requirement (matches backend validation)
+var is_valid_contact_email = function(email) {
+    if (!email || typeof email !== 'string') {
+        return false;
+    }
+    email = email.trim();
+    // Require FQDN with TLD (at least 2 characters after the last dot)
+    // This matches the backend validate_domain_full() function
+    var emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*\.[a-zA-Z]{2,}$/;
```
kroky: Maybe we misunderstand each other in some way. I see you request my review 3 times now and I return it 3 times with the same comment. Here, I will list all the places I see email regex that should be unified - ultimately, we want to specify this regex only once in the whole app. Here's the first place.

modules/core/site.js
```diff
@@ -1744,7 +1744,9 @@ var Hm_Utils = {
     },
 
     is_valid_email: function (val) {
-        return /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|.(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/.test(val)
+        var email = val.trim();
+        var emailRegex = /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*\.[a-zA-Z]{2,}$/;
```
kroky: Second place...

modules/smtp/site.js
```diff
@@ -336,8 +336,16 @@ var copy_text_to_clipboard = function(e) {
...
 var is_valid_recipient = function(recipient) {
-    var valid_regex = /^[\p{L}|\d' ]*(<)?[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*(>)?$/u;
-    return recipient.match(valid_regex);
+    if (!recipient) {
+        return false;
+    }
+    var val = recipient.trim();
+
+    // Support "Name <email@example.com>" format by extracting email from angle brackets
+    var match = val.match(/<([^>]+)>/);
```
kroky: Another regex for email matching...

IrAlfred:
In fact, the `is_valid_recipient` function, used on the Compose page for example, uses different validation logic compared to a simple email field on the contact form, which only accepts the email address. It must accept these different formats:
- `Display Name <email@example.com>`,
- `Display Name email@example.com`, or
- `email@example.com`

modules/smtp/site.js
```diff
@@ -336,8 +336,27 @@ var copy_text_to_clipboard = function(e) {
...
 var is_valid_recipient = function(recipient) {
-    var valid_regex = /^[\p{L}|\d' ]*(<)?[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*(>)?$/u;
-    return recipient.match(valid_regex);
+    if (!recipient) {
+        return false;
+    }
+    var val = recipient.trim();
+
+    // Check format: "Display Name <email@example.com>", "Display Name email@example.com", or "email@example.com"
+    var match = val.match(/^(.+)\s+<([^>]+)>$/);
```
kroky: Yes, here it starts to get complicated. Can you check that this validation is broad enough to support all RFC 5322 cases? E.g. emails can be separated by comma or semi-column, can include various types of brackets, etc. I just want to make sure we don't block a valid use-case which is annoying for the user. Imagine someone pasting a list of emails in the CC field, for example.

IrAlfred: Added RFC 5322 compliance. This required too many changes

PR #1820, opened 2025-12-29
fix(frontend): false alert on the existence of the mail address in the block list

christer77:


modules/imap/site.js
```diff
@@ -1400,7 +1400,8 @@ $('.screen-email-like').on("click", function() {
         }
     });
 
-    if (email_existing_in_blocked_senders) {
+    
+    if (email_existing_in_blocked_senders.length > 0) {
```
kroky, line 1404: What would be a non-falsy value that failed before but would pass now?

christer77: Even for an empty array, the JavaScript test returns "true".

PR #1825, opened 2026-01-08
fix(frontend): fix incorrect unsaved changes count when adding multip…

espoirbaraka:
When adding servers through cypht/?page=servers, the "Unsaved Changes" list on the save page displayed incorrect counts:

- SMTP server added: (2X) instead of (1X)

- IMAP server added: (1X) 

- Special folders assigned: (1X)

### DIFF
--- a/modules/nux/modules.php
+++ b/modules/nux/modules.php
@@ -216,7 +216,6 @@ class Hm_Handler_process_nux_add_service extends Hm_Handler_Module {
                     }
                     Hm_IMAP_List::clean_up();
                     $this->session->record_unsaved('IMAP server added');
-                    $this->session->record_unsaved('SMTP server added');
                     $this->session->secure_cookie($this->request, 'hm_reload_folders', '1');
                     if (isPageConfigured('save')) {
                         Hm_Msgs::add("E-mail account successfully added, To preserve these settings after logout, please go to <a class='alert-link' href='/?page=save'>Save Settings</a>.");


modules/nux/modules.php
```diff
@@ -214,16 +214,15 @@ public function process() {
...
                     $this->session->record_unsaved('IMAP server added');
-                    $this->session->record_unsaved('SMTP server added');
                     $this->session->secure_cookie($this->request, 'hm_reload_folders', '1');
                     if (isPageConfigured('save')) {
                         Hm_Msgs::add("E-mail account successfully added, To preserve these settings after logout, please go to <a class='alert-link' href='/?page=save'>Save Settings</a>.");
                     } else {
                         Hm_Msgs::add("E-mail account successfully added.");
                     }
                     $this->session->close_early();
+                    Hm_IMAP_List::clean_up();
```
kroky: Why is this moved after the session closing?

modules/nux/modules.php
```diff
@@ -214,16 +214,15 @@ public function process() {
                             $this->session->record_unsaved('SMTP server added');
                         }
                     }
-                    Hm_IMAP_List::clean_up();
                     $this->session->record_unsaved('IMAP server added');
-                    $this->session->record_unsaved('SMTP server added');
```
kroky, line 219: Here, we seem to add one smtp and one imap server. Why do you remove the smtp msg? If they are doubled, maybe something else is adding another smtp msg?

modules/nux/modules.php
espoirbaraka: Moved `clean_up()` before `close_early()` to match the pattern in `Hm_Handler_process_oauth2_authorization` and the codebase.

modules/nux/modules.php
espoirbaraka: The SMTP message isn’t removed; it’s conditional in `Hm_Handler_process_nux_add_service` (line 214). It only records if `can_save_last_added_server()` returns true, which prevents recording when a duplicate server is detected and removed.

modules/nux/modules.php
kroky: This is actually not doing anything productive. It was still called before close_early and could stay there is is. Furthermore, the pattern in the process_oauth2_authroization is that you clean up, then record messages, then close session. The change you did here actually breaks this pattern. Another point is that it doesn't seem related the the issue you are fixed. Please avoid such changes.

espoirbaraka: You're right. I've reverted it.

PR #1828, opened 2026-01-11
fix(other): fix adding screen_emails param in searching request

christer77:


marclaporte:
@christer77 A test has failed.

kroky:
Selenium tests are still not running smoothly with random failures which have nothing to do with the committed code. We should really fix this one day...

Issue #555, opened 2022-02-23
Warn about missing dependency (was: Nothing sent with compose method "markdown")


smitgd:
## 🐛 Bugreport

I sent a plain text message ok and it arrived at the destination. Then I wanted to see if the markdown setting under Site/General would work. After composing the message and clicking send, the screen goes completely blank and nothing is sent and nothing received. I went back to plain text mode and it sent again ok. Tried markdown mode sending couple more times and still wouldn't send and screen went blank. Didn't try html.

### Version & Environment

Rev: [not sure, got it around Feb 7 using the install script]

OS:  [linux]

jasonmunro:
Sounds like a bug, let me try to reproduce and I will let you know. Thanks for the report!

jasonmunro:
I was able to send with markdown without issue. I did find a bug after send causing a 500 error/blank page however I just pushed a fix for that and it was happening after send so probably not what you are seeing.

We use a third party library called parsedown to convert the message body markdown to html on send (clients don't normally natively render markdown), since your site is failing before sending I suspect something want wrong with the "composer install" step of the installation. When you get a blank page it's almost always a server side 500 error and there maybe a clue in your webserver log about what is going wrong. I would try rerunning the composer step and checking the logs if it still fails.

kk6mrp:
I'm getting this error when trying to reply to messages:

```bash AH01071: Got error 'PHP message: PHP Fatal error: Uncaught Error: Class "DOMDocument" not found in /usr/local/share/cypht/modules/core/message_functions.php:427\nStack trace:\n#0
/usr/local/share/cypht/modules/core/message_functions.php(38): HTMLToText->__construct()\n#1 /usr/local/share/cypht/modules/core/message_functions.php(318): convert_html_to_text()\n#2 /usr/local/share/cypht/modules/core/message_functions.php(274): format_reply_as_text()\n#3 /usr/local/share/cypht/modules/core/message_functions.php(382)
[...]

PR #1830, opened 2026-01-11
feat(other): add module-aware PHP dependency checker to build script

IrAlfred:
Related issue: https://github.com/cypht-org/cypht/issues/555

### DIFF
--- a/scripts/config_gen.php
+++ b/scripts/config_gen.php
@@ -4,7 +4,7 @@
  * CLI script to build the site configuration
  */
 
-if (mb_strtolower(php_sapi_name()) !== 'cli') {
+if (strtolower(php_sapi_name()) !== 'cli') {
     die("Must be run from the command line\n");
 }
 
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -38,6 +38,9 @@
 $environment = Hm_Environment::getInstance();
 $environment->load();
 
+/* check dependencies */
+$environment->check_dependencies();
```
kroky: Let's do this in config_gen or something that sets up the site rather than on each and every web request.

lib/environment.php
```diff
@@ -107,6 +107,73 @@ private function set_required_environment_variables($file) {
...
+            'curl' => 'cURL (required for HTTP requests)',
+            'fileinfo' => 'fileinfo (required for file handling)',
+            'filter' => 'filter (required for input validation and sanitization)',
+            'iconv' => 'iconv (required for character encoding)',
+            'json' => 'json (required for JSON operations)',
+            'mbstring' => 'mbstring (required for multibyte string operations)',
+            'openssl' => 'openssl (required for encryption and secure connections)',
+            'session' => 'session (required for session management)',
+            'xml' => 'XML (required for HTMLPurifier, Gmail contacts, CardDAV, and feeds)',
+            'dom' => 'DOM (required for HTMLPurifier and XML processing)',
```
kroky: Are you sure we are using all of those extensions?

lib/environment.php
```diff
@@ -107,6 +107,73 @@ private function set_required_environment_variables($file) {
...
+        }
+
+        // Log missing optional extensions to debug log (not shown to users)
+        if (!class_exists('PDO')) {
+            Hm_Debug::add('Optional PHP extension missing: pdo - Database features will not work', 'warning');
+        }
+        if (!class_exists('Redis')) {
+            Hm_Debug::add('Optional PHP extension missing: redis - Redis caching or sessions will not work', 'warning');
+        }
+        if (!class_exists('Memcached')) {
```
kroky: Some of these are false-positive warnings. It might be worth checking the actually enabled modules and session config to see if these modules would cause conflicts or not.

lib/environment.php
IrAlfred: They may change depending on which modules are activated. Let me investigate more deeply

PR #1844, opened 2026-01-24
fix(backend): handle undefined array key name

Baraka24:
### Problem

Opening any  folder triggered:

```
PHP Warning: Undefined array key "name" in modules/core/hm-mailbox.php on line 133
```

The page title/label still rendered because a separate fallback in `handler_modules.php` recovered the name via `get_connected_mailbox`. But the warning was logged on every request.

### Root cause

`Hm_Mailbox::get_folder_name()` internally called `connect()` when `get_folder_name_quick()` returned false. For EWS non-distinguished folders (opaque binary IDs like `AAMkADY3...`) this is always the case, because unlike IMAP/JMAP, the folder identifier is not the display name — resolving it requires a live Exchange server call.

The mailbox in this code path is created via `Hm_IMAP_List::get_mailbox_without_connection()`, which intentionally strips the password. So `connect()` failed silently, `get_folder_status()` caught the resulting exception and returned `[]`, and `$result['name']` on an empty array produced the warning.

This was an architectural violation: `get_folder_name()` was calling `connect()` behind the caller's back on a mailbox explicitly created without a connection.

### Fix

**`modules/core/hm-mailbox.php`**

- `get_folder_name()`: for non-IMAP types, now simply returns `get_folder_name_quick() ?: null`. It no longer attempts to connect or call `get_folder_status()` on its own. The caller is responsible for obtaining a connected mailbox when a live server call is needed.

**`modules/imap/handler_modules.php`**

- EWS folder name resolution is now a top-level branch in the fallback, not buried under the "session cache off" path.
- For EWS non-distinguished folders, the handler now directly uses `get_connected_mailbox()` (the already-established session connection) to resolve the display name — with no warning, since this is the normal expected path for EWS, not a slow fallback.
- IMAP/JMAP logic is unchanged (IMAP never reaches the fallback since `get_folder_name()` returns the path string directly).

### DIFF
--- a/modules/core/hm-mailbox.php
+++ b/modules/core/hm-mailbox.php
@@ -122,16 +122,12 @@ class Hm_Mailbox {
         if ($this->is_imap()) {
             return $folder;
         } else {
-            $result = $this->connection->get_folder_name_quick($folder);
-            if ($result) {
-                return $result;
-            } else {
-                if (! $this->connection->authed()) {
-                    $this->connect();
-                }
-                $result = $this->connection->get_folder_status($folder);
-                return $result['name'];
-            }
+            // EWS identifies folders by opaque binary IDs rather than human-readable path
+            // strings like IMAP/JMAP. get_folder_name_quick() only resolves distinguished
+            // folder names (e.g. inbox, sentitems). For all other folders the caller must
+            // use an authenticated connection (e.g. via get_connected_mailbox()) to resolve
+            // the display name — this method does not attempt to connect on its own.
+            return $this->connection->get_folder_name_quick($folder) ?: null;
         }
     }
 
--- a/modules/core/hm-mailbox.php
+++ b/modules/core/hm-mailbox.php
@@ -118,16 +118,23 @@ class Hm_Mailbox {
         }
     }
 
-    public function get_folder_name($folder) {
+    public function get_folder_name($folder, $quick = false) {
         if ($this->is_imap()) {
             return $folder;
         } else {
-            // EWS identifies folders by opaque binary IDs rather than human-readable path
-            // strings like IMAP/JMAP. get_folder_name_quick() only resolves distinguished
-            // folder names (e.g. inbox, sentitems). For all other folders the caller must
-            // use an authenticated connection (e.g. via get_connected_mailbox()) to resolve
-            // the display name — this method does not attempt to connect on its own.
-            return $this->connection->get_folder_name_quick($folder) ?: null;
+            $result = $this->connection->get_folder_name_quick($folder);
+            if ($result) {
+                return $result;
+            }
+
+            if ($quick) {
+                return null;
+            }
+            if (!$this->connection->authed()) {
+                $this->connect();
+            }
+            $result = $this->connection->get_folder_status($folder);
+            return $result['name'];
         }
     }
 
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -616,24 +616,21 @@ class Hm_Handler_imap_message_list_type extends Hm_Handler_Module {
                     }
 
                     $mailbox = Hm_IMAP_List::get_mailbox_without_connection($details);
-                    $label = $mailbox->get_folder_name($folder);
+                    $label = $mailbox->get_folder_name($folder, true);
                     if (!$label) {
-                        if (isset($details['type']) && $details['type'] === 'ews') {
-                            // EWS folders have opaque binary IDs; get_folder_name() only resolves
-                            // distinguished names (inbox, sentitems, etc.) without a connection.
-                            // Use the established session connection for all other folders.
-                            $connected_mailbox = Hm_IMAP_List::get_connected_mailbox($parts[1], $this->cache);
-                            if ($connected_mailbox && $connected_mailbox->authed()) {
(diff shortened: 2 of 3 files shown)


modules/core/hm-mailbox.php
```diff
@@ -130,7 +130,7 @@ public function get_folder_name($folder) {
                     $this->connect();
                 }
                 $result = $this->connection->get_folder_status($folder);
-                return $result['name'];
+                return $result['name'] ?? null;
```
kroky: This doesn't handle the issue properly but simply ignores it. Why is the name not defined? Shouldn't a folder always have a name?
If the ajax call requests non-existing folder, then trace back the code to see why does it do it? Ultimately, if it is a user error (e.g. setting up custom folders for something that doesn't exist), we should show an error to the user, so they can fix. If it is Cypth code handling data incorrectly, we should fix the code error.

Baraka24: We discovered that an exception is normally thrown at line 182 in the **"modules\imap\hm-ews.php"** file, but unfortunately its `$e->getMessage()` is empty, making it impossible to capture the real error behind this behavior. This explains why the **get_folder_status()** method returns an empty result array in the catch block, which then causes the **"undefined array key 'name'"** error in the **get_folder_name()** method - the EWS API is failing silently without providing any meaningful error message, so we need to enhance the exception handling to log more details like the exception type, code, and stack trace to properly diagnose what's causing the folder lookup to fail.

kroky: Then, go for it and debug further the EWS client or API. There should be an underlying issue somewhere. I don't like the current approach. Folder not found means something - if it is a normal folder name, then this is a real issue to debug further. Note that EWS and the client we use is a bit pesky about folder naming having both constants, names and base64 encoded IDs with various ways to search for a folder.

Baraka24:
@IrAlfred, I assign this to you. Thanks to continue the work around.

modules/core/hm-mailbox.php
```diff
@@ -50,7 +50,20 @@ public function connect() {
...
             return false;
         }
-        return $this->connection->connect($this->config);
+        $config = $this->config;
+        // EWS identifies folders by opaque binary IDs (e.g. AAMkADY3...) rather than
+        // human-readable path strings like IMAP/JMAP (e.g. INBOX.Sent). Resolving a
+        // display name from an ID therefore requires a live Exchange server call, which
+        // needs authentication. However, get_mailbox_without_connection() intentionally
+        // strips credentials. Fetch them from the session-backed server list so EWS
+        // can connect on demand.
```
kroky: One thing I don't understand is why do we have to connect when the call is get_mailbox_without_connection? Who is calling the connect method? I think that was meant to fix performance issues when connection is not needed. Maybe some other code needs to be tweaked to not call connect method in that case?

IrAlfred: @kroky This is a really crucial point, thank you for raising it.
You're right to question why get_mailbox_without_connection would still need to call connect. The naming strongly suggests no connection should be established, so either the method name is misleading, or there's some hidden coupling that forces a connect despite the name.

We will deeply investigate this.

modules/imap/handler_modules.php
```diff
@@ -617,21 +617,23 @@ public function process() {
...
-                        if ($this->config->get('allow_session_cache', false)) {
+                    if (!$label) {
+                        if (isset($details['type']) && $details['type'] === 'ews') {
+                            // EWS folders have opaque binary IDs; get_folder_name() only resolves
+                            // distinguished names (inbox, sentitems, etc.) without a connection.
+                            // Use the established session connection for all other folders.
+                            $connected_mailbox = Hm_IMAP_List::get_connected_mailbox($parts[1], $this->cache);
+                            if ($connected_mailbox && $connected_mailbox->authed()) {
+                                $folder_status = $connected_mailbox->get_folder_status($folder, false);
+                                $label = $folder_status['name'] ?? null;
```
kroky: Why do you move this code from the central hm-mailbox where it belongs to a random handler module? What about all other usages? If you meet a place where get_folder_name is executed on a non-connected mailbox and it shouldn't connect, add a parameter to that get_folder_name method and call it with a parameter like light=true or quick=true and then use the quick method.

modules/imap/handler_modules.php
```diff
@@ -616,22 +616,18 @@ public function process() {
...
                     $mailbox = Hm_IMAP_List::get_mailbox_without_connection($details);
-                    $label = $mailbox->get_folder_name($folder);
-                    if(!$label) {
-                        if ($this->config->get('allow_session_cache', false)) {
+                    $label = $mailbox->get_folder_name($folder, true);
+                    if (!$label) {
+                        $connected_mailbox = Hm_IMAP_List::get_connected_mailbox($parts[1], $this->cache);
+                        if ($connected_mailbox && $connected_mailbox->authed()) {
+                            $label = $connected_mailbox->get_folder_name($folder);
+                        } elseif ($this->config->get('allow_session_cache', false)) {
```
kroky: The code here changes behavior. Previously, if the label couldn't be fetched, it was tried from the cache and then from the server. I think you should reverse the order of operations here - try quick first, then load from cache and if not available in the cache as well, try to long version.

PR #1854, opened 2026-02-08
feat(frontend/backend): add feature report spam

Grandi0z:
[Related task](https://avan.tech/item78936)

This PR introduces a spam reporting system in Cypht, allowing users to report unwanted messages to multiple external services.


modules/spam_reporting/functions.php
```diff
@@ -0,0 +1,1254 @@
...
+
+use ZBateson\MailMimeParser\MailMimeParser;
+
+if (!defined('DEBUG_MODE')) { die(); }
+
+/**
+ * Internal spam report model
+ * @subpackage spam_reporting/lib
+ */
+class Hm_Spam_Report {
```
kroky: Please move class definitions to their own files and leave only functions in this file.

modules/spam_reporting/functions.php
```diff
@@ -0,0 +1,1254 @@
...
+        );
+    }
+    return $out;
+}}
+
+/**
+ * Build one effective-target descriptor (public fields only); adapter/instance_config added by caller.
+ */
+if (!hm_exists('spam_reporting_build_effective_descriptor')) {
+function spam_reporting_build_effective_descriptor($adapter, $id, $label, array $instance_config = array()) {
```
kroky: This and surrounding functions seem way overcomplicated to me. Can you simplify the code? I understand some abstraction is needed because of the multiple types of spam reporting but please don't nest/mix so many patterns here as registry, descriptors, resolver, provider, etc. I am pretty sure all this can be coded with just a couple of classes and nice usage/reusage of code.

modules/spam_reporting/site.js
```diff
@@ -0,0 +1,741 @@
...
+                return;
+            }
+            spam_reporting_targets = res.spam_report_targets || [];
+            spam_reporting_platforms = res.spam_report_platforms || [];
+            spam_reporting_suggestion = res.spam_report_suggestion || {};
+            spam_reporting_render_targets(spam_reporting_targets, spam_reporting_suggestion, spam_reporting_platforms);
+            spam_reporting_render_suggestion(spam_reporting_suggestion);
+            spam_reporting_render_platforms(spam_reporting_platforms);
+            spam_reporting_update_data_summary('');
+            spam_reporting_apply_preview(res.spam_report_preview || {});
```
kroky: Chunks like this of 5 consecutive function calls should be avoided. If you simplify the spam reporting on php side, this can be simplified as well. Don't try to overengineer this and have so many configuration options, etc. We need to start simple and build on top of that. Otherwise, it comes unmanageable. Furthremore, Cypth started with simplicity and conciseness in mind, we should keep these virtues.

modules/spam_reporting/site.js
```diff
@@ -0,0 +1,741 @@
...
+    var enabled = document.getElementById('spam_reporting_enabled') && document.getElementById('spam_reporting_enabled').checked;
+    var html = '';
+    var order = ['abuseipdb', 'spamcop', 'custom_email'];
+    order.forEach(function(serviceKey) {
+        var def = spam_reporting_service_definitions[serviceKey];
+        if (!def || !spam_reporting_settings_has_adapter(def.adapter_id)) return;
+        var configs = spam_reporting_settings_get_configs_for_service(serviceKey);
+        var connected = configs.length > 0;
+        var cardClass = 'spam-reporting-service-card card mb-3' + (enabled ? '' : ' spam-reporting-card-disabled');
+        html += '<div class="' + cardClass + '" data-service="' + serviceKey.replace(/"/g, '&quot;') + '">';
```
kroky: I'd like to avoid such large html building in JS. Build on php side and provide as templates. One day, we will move to a template engine and it will be easy to convert the php side to use the template engine but not the JS side.

tests/phpunit/bootstrap.php
```diff
@@ -55,7 +55,9 @@
 $environment = Hm_Environment::getInstance();
 // Load test environment configuration from .env.test
 // Create this file by copying .env.test.example: cp .env.test.example .env.test
+$prev = error_reporting(E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED);
```
kroky: Why would we need different error reporting when running tests? The goal is to see as close results as possible to actually running cypht, so the same error reporting as running in dev or prod makes sense to me.

modules/spam_reporting/functions.php
Grandi0z: I decided to create a dedicated lib/ folder for them, but I’m wondering if I could just put them all in a single file.

kroky: One class per file is cleaner, thanks!

kroky:
@Grandi0z since this MR is 3100+ line changes most of which are additions, please help me review/understand it more easily - what is the approach you took, what did you implement and why? Just some guidance will be nice.

marclaporte:
This is quite a big feature. @IrAlfred please test and advise.

Grandi0z:
> @Grandi0z since this MR is 3100+ line changes most of which are additions, please help me review/understand it more easily - what is the approach you took, what did you implement and why? Just some guidance will be nice.

## Approach

The goal of this MR is to introduce a **modular spam reporting system**, by separating:

* presentation (UI + AJAX)
* application logic (handlers)
* domain logic (parsing, detection, configuration)
* delivery (API or email)

The system is built around an **adapter pattern**, where each reporting target implements a common interface.

---

## `adapters/` (delivery layer)

This folder contains the implementations of the different reporting targets.

* `report_target_interface.php`
  Defines the common contract (`build_payload`, `deliver`, `is_available`, etc.)

* `abstract_report_target.php`
  Provides a minimal base implementation

* `abstract_api_report_target.php`
  Shared logic for API-based targets (HTTP requests, headers, error handling)

* `abuseipdb_report_target.php`
  Concrete API implementation

* `email_report_target.php`
  Generic email-based reporting (attaches original message as `message/rfc822`)

* `spamcop_email_report_target.php`
  Specific implementation for SpamCop (requires a different MIME format)

Each adapter acts as a strategy selected dynamically at runtime.

---

## Domain / service layer (`lib/` + `functions.php`)

This contains the core business logic:

* data objects (`Hm_Spam_Report`, payload, result, context)
* MIME message parsing (headers, body)
* extraction of relevant data (IP, domains, raw headers)
* provider detection based on headers
* mapping provider → platforms (for suggestions)
* transformation and validation of user configuration

The goal is to keep this logic independent from the delivery layer.

---

## `SpamReportingManager` (central orchestrator)

This is the central component that connects configuration, domain logic, and adapters.

Responsibilities:

* register available adapters
* load and normalize user configuration
* filter adapters based on global config (`allowed_target_types`)
* resolve available targets for a given message
* return the adapter + instance configuration for execution

It acts as a policy layer between raw configuration and executable targets.

---

Global flow:

```
UI → handler → manager → adapter → result
```
---

## Architecture goals

* extensibility (adding a new provider = adding a new adapter)
* clear separation of responsibilities
* support for provider-specific constraints (e.g. SpamCop MIME format)
* flexible configuration (admin-level vs user-level)

---

## Note

This is a fairly large first iteration, so I’m open to feedback

---

marclaporte:
Since this is a fairly big feature, it is OK to take time to analyze and should not block the release of 2.8.0.

Grandi0z:
Hi @IrAlfred, the conflicts have been fixed.
Waiting for your feedback

marclaporte:
Joseph Kausi is no longer available. So I put this in draft, and we can re-explore when other devs are less busy (ex.: Alfred and Steven)

Issue #1873, opened 2026-02-17
🚀 [Feature] Allow to rename the Email folder account for the default IMAP


PeopleInside:
Cyphy version
2.6.0

## 🚀 Feature

Allow to rename the Email default account IMAP folder.
When I login into Cypht I see the Email menu on the left side, if I expand it I found as the first voice the `IMAP_AUTH_NAME= MyIMAP`
[that I set on the env file](https://github.com/cypht-org/cypht/blob/c0ccd9100f2c7dd97566777ee7ef3251d1a514b1/.env.example#L17C1-L28C22)

but this is the email address with I logged in and I want to see the email name.

If I add other accounts I can set the name same as the email address that I add.

For better example suppose I want setup three email address

address1@domain
address2@domain
address3@domain

I should control all this three account when i login on the first one.
When I log in with the first one address1@domain

On the email voice I will have
```
MyIMAP
address2@domain
address3@domain
```

I'm looking to rename MyIMAP with address1@domain but seems I can't
In the .env settings I set IMAP as authentication and I insert settings for IMAP
https://github.com/cypht-org/cypht/blob/c0ccd9100f2c7dd97566777ee7ef3251d1a514b1/.env.example#L15

I changed from 
`AUTH_TYPE=DB`

to:
`AUTH_TYPE=IMAP`

then

PeopleInside:
The issue is not maybe about rename manually.
What Crypht show should be not what set in the .env file but should show the email name.

If I rename in the .env file this will be renamed on all account I login.
Instead of showing the name set in the .env file maybe the main default address with one has been done the login should be showed.

If I login with email1@domain.ext and in the Email I see something like "domain.ext IMAP" and in the future I can rename in email1@domain.ext

the issue will be present for another user that login with email2@domain.ext . They should see email2@domain.ext under email menu and not "domain.ext IMAP"

IrAlfred:
> What Crypht show should be not what set in the .env file but should show the email name.

I agree with this

IrAlfred:
Hello @PeopleInside

The fix has been applied and will be available in the Cypht 2.7.0 release.

PR #1879, opened 2026-02-18
fix(other): use user email as IMAP account name instead of static config

IrAlfred:
In this PR we replace hardcoded IMAP_AUTH_NAME with the authenticated user's email
address to provide personalized account names in the folder list.

Related issue: https://github.com/cypht-org/cypht/issues/1873

### DIFF
--- a/modules/imap/handler_modules.php
+++ b/modules/imap/handler_modules.php
@@ -1804,7 +1804,8 @@ class Hm_Handler_load_imap_servers_from_config extends Hm_Handler_Module {
                 $name = $auth_server['name'];
             }
             else {
-                $name = $this->config->get('imap_auth_name', 'Default');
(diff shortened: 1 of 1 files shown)


modules/imap/handler_modules.php
```diff
@@ -1804,7 +1804,8 @@ public function process() {
                 $name = $auth_server['name'];
             }
             else {
-                $name = $this->config->get('imap_auth_name', 'Default');
+                // Use the username (email address) as the account name instead of static config
+                $name = $auth_server['username'];
```
kroky, line 1808: Are we sure username always exists and is defined?

IrAlfred: Yes, it's actually the email used when we login

PR #1885, opened 2026-02-27
fix(other): update version reference to 2.7.0 for upcoming release

IrAlfred:


Baraka24:
Okay. Thanks @IrAlfred. When is next release? I think you should merge this some hours before.

PR #1889, opened 2026-03-02
fix(backend): prioritize exact, case-insensitive match over substring-based when searching for a message body struct. Preventing falsy match of subtype (e.g., 'x-amp-html' to 'html')

mercihabam:


modules/imap/hm-imap-bodystructure.php
```diff
@@ -403,7 +403,7 @@ public function recursive_search($struct, $flds, $all, $res, $parent=false) {
                 }
             }
             foreach ($flds as $name => $fld_val) {
-                if (isset($vals[$name]) && mb_stristr($vals[$name], $fld_val)) {
+                if (isset($vals[$name]) && strcasecmp($vals[$name], $fld_val) === 0) {
```
kroky: I think the reason it uses a substring match is because of subparts... what is the problem you are trying to solve?

mercihabam: A subpart is checked as a field in its own as per [hm-imap.php#L2515](https://github.com/cypht-org/cypht/blob/master/modules/imap/hm-imap.php#L2515). The problem I'm trying to solve is that the default message part being shown is `text/x-amp-html` instead of the `text/html` one when both are available.

kroky: ok, I still think there might be valid cases to search by substring. Can't we safeguard here with:
- search for substring or exact match
- prefer exact matches first and then substrings
So, getting the first part matching something (like content type header) will return the part if there is an exact match and only then return substring matches even if they appear before the others.

mercihabam: Done. I made substring matching occur only when no exact match is found, to prevent both conditions from being true at once, which would result in the earliest occurrence being validated.

Issue #1027, opened 2024-05-15
Tame error logging


jonocodes:
## 🗣 Suggestion

Discussion began here: https://github.com/cypht-org/cypht/pull/1001#pullrequestreview-2049777476

I am genuinely confused with the logging setup and it has bitten me several times. There is a lot of good info being sent go Hm_Debug::add, but its hard to get the data out from there. I lost several hours trying to fix a sqlite bug that would have been minutes if those logs were simply displayed.

Its possible PHP logging is different because one of the output streams is meant to screen/browser output. I hear Laraval is a commonly used PHP framework, so I looked up their logging setup: [laravel.com/docs/11.x/logging](https://laravel.com/docs/11.x/logging)

which seems quite standard. It uses [monolog](https://github.com/Seldaek/monolog) which provides simple, immediate, customizable outputs.

Also DEBUG_MODE is only hard-coded in scripts for now. It will be a good idea to extend that to come from the environment.

jonocodes:
@kroky said:
> Hm_Debug uses Hm_list which has `get` method to get the messages and `show` method to send them to the error log. PHP error logging might be complex and not everything logged, indeed, it has settings for error_reporting - what kind of errors to report, then displaying errors on screen or writing errors to a log file - thus third option also depends on your backend server setup. So, a lot of moving parts.
> 
> Integrating something like monolog is an option but note that cypht started with minimalism in mind while Laravel is enormous, so we still try to depend on fewer external libs. That being said, I think an ENV setting for DEBUG_MODE which is observed and makes sure debugging messages are written to the php error log at the end of a request + screen/cli if necessary seems like a better approach to me.

jonocodes:
> Hm_Debug uses Hm_list which has `get` method to get the messages and `show` method to send them to the error log. PHP error logging might be complex and not everything logged, indeed, it has settings for error_reporting - what kind of errors to report, then displaying errors on screen or writing errors to a log file - thus third option also depends on your backend server setup. So, a lot of moving parts.

@kroky
I think I should understand the part where it "displays errors on the screen". I presume that's to the browser? Or is it hidden in the browser source so its not displayed? That may clarify to me why we are queuing it up. Can you show me an example of where I can see that happening?

kroky:
[...]

PR #1891, opened 2026-03-11
feat(other): add Monolog for structured logging with environment-based configuration

IrAlfred:
Related issue: https://github.com/cypht-org/cypht/issues/1027


kroky:
Looks good, merging it!

PR #1897, opened 2026-03-15
refactor(frontend/backend): propose sieve filter actions according to server extensions

josaphatim:
## 🍰 Pullrequest
This PR aims to propose to users actions according to supported extensions on the server. Users won't see what is not available for them to use.


marclaporte:
This is a good principle: https://doc.tiki.org/WYSIWYCA

PR #1898, opened 2026-03-17
fix(other): only init KindEditor for signature in HTML compose mode

IrAlfred:
### Issue

The stepper (add server) and EWS (add Exchange server) forms were unconditionally initializing KindEditor on the signature textarea by targeting the element by ID alone. This caused the rich-text editor to appear even when **Outbound mail format** was set to **Plain Text**.


kroky:
Oh, nice catch, thanks!

PR #1899, opened 2026-03-20
fix(backend): allow custom sources

josaphatim:
This PR reintroduce a way of adding custom sources from apps integrating Cypht. Recent refactoring removed the callback that where allowing apps to add sources to message list pages.


marclaporte:
Will be used here: https://gitlab.com/tikiwiki/tiki/-/merge_requests/9805

Issue #682, opened 2023-02-26
Handle winmail.dat


marclaporte:
I some old emails with this:
https://support.mozilla.org/en-US/kb/what-winmaildat-attachment

Roundcube can read them
https://docs.roundcube.net/doc/phpdoc/classes/rcube_message.html#method_tnef_decode

marclaporte:
https://packagist.org/packages/qualityunit/tnef-decoder

dumblob:
Hm, my first thought is that this is way too specific to implement & maintain for a web MUA. There are plenty web browser extensions which extract the content and automatically import it into e.g. Gmail. Such extensions could be extended (pun intended) to support Cypht.

Or alternatively it could be just part of Cypht wiki listing external tooling like those TNEF decoders/exporters. I will also add one in Python:

https://github.com/delimitry/tnef_data_extractor

marclaporte:
The above shared link to Roundcube is now 404.

Roundcube added support in 2009:
https://roundcube.net/news/2009/03/10/update-021-released

marclaporte:
I received an email today and Cypht outputs winmail.dat

Thunderbird recommends "The best solution is to contact the message sender and inform them that their copy of Outlook is incorrectly configured as suggested in the Microsoft Support article."
https://support.mozilla.org/en-US/kb/what-winmaildat-attachment

Roundcube handles it well.
https://github.com/roundcube/roundcubemail/blob/master/program/lib/Roundcube/rcube_tnef_decoder.php but license is GPL so we can't easily add to Cypht

Same licensing issue for:
https://packagist.org/packages/qualityunit/tnef-decoder

Horde is LGPL (compatible with Cypht)
https://github.com/horde/Mime_Viewer/blob/FRAMEWORK_6_0/lib/Horde/Mime/Viewer/Tnef.php

At the very least, we should add an error message in Cypht pointing to online resources. Even better is to add support to Cypht.

marclaporte:
I saw another winmail.dat message in my mailbox the other day. I don't remember how this happened.

Idea: Have a tool/action to scan a mailbox and find them all.

marclaporte:
I had this today. I received an email with many attachments on a webmail with KolabNow. Then, I moved the email via Cypht (I don't remember if it was to another IMAP account or to Tiki Trackers) and the attachments became winmail.dat. I moved the email back and all was well.

marclaporte:
I got this again a few times lately. It is annoying.

[...]

PR #1902, opened 2026-03-23
feat(backend): implement support for application/ms-tnef message

mercihabam:
Implements: #682 

This implementation relies on [tnef](https://linux.die.net/man/1/tnef) and [unrtf](https://linux.die.net/man/1/unrtf) being available on the system. Also, requires that `ENABLE_MSTNEF_VIEWER` be enabled via the environment configuration variables.

See also: https://github.com/verdammelt/tnef


composer.json
```diff
@@ -58,7 +60,11 @@
         "twbs/bootstrap": "^5.3",
         "twbs/bootstrap-icons": "^1.11",
         "webklex/composer-info": "^0.0.1",
-        "zbateson/mail-mime-parser": "^3.0"
+        "zbateson/mail-mime-parser": "^3.0",
+        "horde/mime_viewer": "3.0.0alpha7",
```
marclaporte: So it means we are pulling in some dependencies from Horde.

How many distinct dependencies (with the sub-dependencies)?
How many megs in total?

mercihabam: We have quite a number of dependencies pulled from Horde, 23 in total (including sub-dependencies), weighing 16.9 MB.

marclaporte: Understood. This is too much, especially for such a rarely used feature.

Please try to find a lightweight alternative. Either:
- by cherry-picking only what we need from Horde or
- by picking another dependency. If the license is not compatible, we can ask them to dual license or re-license. Cypht is the same license as Tiki: https://tiki.org/License

Thanks!

@kroky @josaphatim @Shadow243

marclaporte: Roundcube is GPL so we can't easily use.

But the original code in Roundcube seems to come from Horde LGPL code: https://github.com/roundcube/roundcubemail/commit/f19d86b96da4a73efa88042dd0bcf0da1a214762

So maybe we could collaborate with Roundcube contributors to extract TNEF into a shared ressource?

kroky: yes, I agree - something more lightweight is needed.

mercihabam: Great! Given the licensing challenge and lib size issue, I've tried replacing the Horde implementation with [tnef](https://linux.die.net/man/1/tnef), which is a lightweight CLI utility. Let me know how you opine about that. Thanks.

marclaporte: Works for me!

I look forward to @kroky's advice

.env.example
```diff
@@ -234,3 +234,7 @@ WORKER_CUSTOM_IMPORTS=
 #Glitchtip errors capturing
 GLITCHTIP_DSN=
 GLITCHTIP_TRACES_SAMPLE_RATE=0.01
+
+# To be able to view Microsoft TNEF decoded messages, you need to enable the following setting
+# and ensure that `tnef` and `unrtf` are installed on your system.
```
kroky, line 239: besides the comment, can you also make the corresponding checks in config_gen where system is set up and warn the user if they are missing or give user hints on how to install them?

modules/imap/functions.php
```diff
@@ -1685,3 +1690,62 @@ function is_imap_trash_folder($handler, $server_id, $folder) {
...
+
+                    unlink($file);
+                } elseif (strpos($mimeType, 'image/') === 0) {
+                    // display images
+                    $base64Data = base64_encode(file_get_contents($file));
+                    $attachmentsOutput .= '<div><img src="data:' . $mimeType . ';base64,' . $base64Data . '" alt="' . htmlspecialchars($fileInfo['basename']) . '" class="img-fluid"></div>';
+                    unlink($file);
+                } else {
+                    // other attachments
+                    $fileUrl = "?page=message&imap_msg_uid=$msgId&download_attachment=" . urlencode($fileInfo['basename']);
```
kroky: This prevents us from being able to remove the file and temporary directory, so proper cleanup is impossible. Over time, we will get many directories in the attachments dir. Can we prioritize cleanup here and just provide the links to the "other" types of files.

Then, when download attachment is executed, we can re-parse the tnef and get the proper file to return to the user. Re-parsing is taking more time but we keep the filesystem clean, so I think it is a good tradeoff. Furthermore, it is up to the user to click those links, so this re-parsing might not actually happen at all.

modules/imap/functions.php
```diff
@@ -1685,3 +1690,62 @@ function is_imap_trash_folder($handler, $server_id, $folder) {
...
+if (!hm_exists('parse_mstnef')) {
+    function parse_mstnef($tnefBinary, $attachmentDir, $msgId) {
+        if (! is_writable($attachmentDir)) {
+            return '<div class="alert alert-danger" rol="alert">Unable to parse TNEF message. Please ensure that "attachment_dir" is configured in your environmment and is writable!</div>';
+        }
+
+        $filename = $attachmentDir . DIRECTORY_SEPARATOR . 'message_' . $msgId . '.dat';
+        file_put_contents($filename, $tnefBinary);
+
+        $unpackDir = $attachmentDir . DIRECTORY_SEPARATOR . 'message_' . $msgId;
```
kroky: Maybe we use unique names here? There might be two processes (users) looking at the same message at the same time.

composer.json
kroky: yes, that's quite better, thanks!

PR #1903, opened 2026-03-25
refactor(frontend): move some JS code of the sievefilter module for execution in route handlers instead of a global execution

mercihabam:
This PR fixes runtime errors occurring in the Tiki integration with Cypht, caused by the execution of JS code globally instead of a per-route handled execution. This is problematic because not all external JS dependencies will have been loaded (Bootstrap particularly) when the browser parses JS files.

Furthermore, some global variables are declared both in Cypht and in Tiki, and they cause runtime errors. So, opting for `var` instead of `let` when declaring global state variables. Much better would be to block-scope those variables, but this can happen only after gathering the use case of each one. I will try to find time for this in the future, but the current approach seems useful as well.

Thanks!

### DIFF
--- a/modules/core/js_modules/route_handlers.js
+++ b/modules/core/js_modules/route_handlers.js
@@ -156,6 +156,8 @@ function applyMessageListPageHandlers(routeParams) {
         Hm_Message_List.sort($(this).val());
     });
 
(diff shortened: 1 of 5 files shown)


kroky:
Looks good, thanks!

PR #1301, opened 2024-10-26
Allow the exclusion of certain JS libraries from being bundled when integrating Cypht with third-party software

jacob-js:

PR #1914, opened 2026-04-01
refactor(backend): revamp the site module for a better customization and extension of internal features and behaviours

mercihabam:
Related: #1909 

Cypht already supports integration through the site module, but in practice some integrations (for example Tiki: https://doc.tiki.org/Webmail) still patch Cypht core code directly.

A likely reason is that site-level customization has not been flexible or explicit enough for key concerns such as authentication, sessions, cache, and configuration.

This PR improves and clarifies the customization model:

* Custom integration code can live in any accessible directory, configured through `SITE_MODULE_PATH`.
* Code in that path is automatically bound to Cypht internal modules (following the `modules/site` pattern).
* Core customization points now use predictable custom classes based on config type flags, without requiring extra class-name settings.

### Customization Points

* **Authentication**: If `AUTH_TYPE` is set to `custom`, Cypht uses `Hm_Custom_Auth` (no `AUTH_CLASS` needed).
* **Session handling**: If `SESSION_TYPE` is set to `custom`, Cypht uses `Hm_Custom_Session` (no `SESSION_CLASS` needed).
* **Configuration**: If `SITE_CONFIG_TYPE` is set to `custom`, Cypht uses `Hm_Custom_Site_Config`. If `USER_CONFIG_TYPE` is set to `custom`, Cypht uses `Hm_Custom_User_Config`.
* **Cache**: If `ENABLE_CUSTOM_CACHE` is enabled, Cypht uses `Hm_Custom_Cache` (no `CACHE_CLASS` needed).
* **Sieve client factory**: if `ENABLE_CUSTOM_SIEVE_FACTORY` is enabled, Cypht uses `Hm_Custom_Sieve_Client_Factory` (no `sieve_client_factory` setting needed).
* **Response handling**: Use `DISPATCH_RESPONSE_MODE` to specify behavior: `render` renders immediately, `store` stores for later retrieval via the dispatcher's `$output` property.
* **Routing**: Use `PAGE_PARAM_NAME` to specify custom page query parameter name. See: #1919.

### Integration Requirements

When using custom site configuration, you must explicitly call `$config->triggerInit()`, which returns the `$dispatcher` object.

Overall, this reduces the need for downstream core patches and makes supported integration points clearer.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -13,8 +13,10 @@ DB_USER=cypht_test
 DB_PASS=cypht_test
 DB_SOCKET=/var/lib/mysqld/mysqld.sock
 
-# Possible values 'DB', 'REDIS', 'MEM', and 'CUSTOM'. If set to 'CUSTOM', the 'SITE_MODULE_PATH' must be set and have the Hm_Custom_Session properly implemented. PHP session is the default if not set.
+# Possible values 'DB', 'REDIS', 'MEM', and 'custom'. If set to 'custom', the 'SITE_MODULE_PATH' must be set and have the Hm_Custom_Session properly implemented. PHP session is the default if not set.
 SESSION_TYPE=PHP
+
+# Possible values are `DB`, `LDAP`, `IMAP`, 'dynamic', and `custom`. If set to `custom`, the `SITE_MODULE_PATH` must be set and have the `Hm_Custom_Auth` properly implemented.
 AUTH_TYPE=DB
 
 IMAP_AUTH_NAME=localhost
--- a/.env.example
+++ b/.env.example
@@ -13,6 +13,7 @@ DB_USER=cypht_test
 DB_PASS=cypht_test
 DB_SOCKET=/var/lib/mysqld/mysqld.sock
 
+# Possible values 'DB', 'REDIS', 'MEM', and 'CUSTOM'. If set to 'CUSTOM', the 'SITE_MODULE_PATH' must be set and have the Hm_Custom_Session properly implemented. PHP session is the default if not set.
 SESSION_TYPE=PHP
 AUTH_TYPE=DB
 
--- a/.env.example
+++ b/.env.example
@@ -55,7 +55,8 @@ JS_COMPRESS=false
 CSS_COMPRESS=false
 
 ALLOW_SESSION_CACHE=false
-CACHE_CLASS=
+# If enabled, the 'SITE_MODULE_PATH' must be set and have the Hm_Custom_Cache properly implemented.
+ENABLE_CUSTOM_CACHE=false
 
 ENABLE_REDIS=true
 REDIS_SERVER='127.0.0.1'
--- a/.env.example
+++ b/.env.example
@@ -29,9 +29,13 @@ DEFAULT_SMTP_PORT=
 DEFAULT_SMTP_TLS=
 DEFAULT_SMTP_NO_AUTH=
 
+# Possible values are 'file', 'db', and 'custom'. If set to 'custom', the 'SITE_MODULE_PATH' must be set.
 USER_CONFIG_TYPE=file
 USER_SETTINGS_DIR=/var/lib/hm3/users
 ATTACHMENT_DIR=/var/lib/hm3/attachments
+
+# Possible values are 'file' and 'custom'. If set to 'custom', the 'SITE_MODULE_PATH' must be set.
+SITE_CONFIG_TYPE=file
 
 ADMIN_USERS=
 
@@ -87,9 +91,6 @@ DISABLE_EMPTY_SUPERGLOBALS=false
 DISABLE_OPEN_BASE_DIR=false
 DISABLE_INI_SETTINGS=false
 DISABLE_FINGERPRINT=false
-
-AUTH_CLASS=
-SESSION_CLASS=
 
 API_LOGIN_KEY=
 
@@ -238,9 +239,13 @@ WORKER_CUSTOM_IMPORTS=
 GLITCHTIP_DSN=
 GLITCHTIP_TRACES_SAMPLE_RATE=0.01
 
-# Your router may be using the 'page' parameter for other purposes when loading Cypht, so this allows you to specify a different parameter name not conflicting with downstream implementations. If not set, it defaults to 'page'.
-PAGE_PARAM_NAME=page
-
 # To be able to view Microsoft TNEF decoded messages, you need to enable the following setting
 # and ensure that `tnef` and `unrtf` are installed on your system.
 ENABLE_MSTNEF_VIEWER=false
+
+# Your router may be using the 'page' parameter for other purposes when loading Cypht, so this allows you to specify a different parameter name not conflicting with downstream implementations. If not set, it defaults to 'page'.
+PAGE_PARAM_NAME=page
+
+# Specify where the module for customized behaviour is located on the web server.
+# See the built-in example at modules/site
+SITE_MODULE_PATH=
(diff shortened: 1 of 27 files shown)


mercihabam:
Merge request in Tiki adapting the integration: https://gitlab.com/tikiwiki/tiki/-/merge_requests/9937

marclaporte:
@mercihabam 

FYI: @camilevahviraki is working on a proof of concept for a webmail module in Dolibarr:
https://github.com/camilevahviraki/dolibarr-cyphtWebmail

As you are making Cypht cleaner to use for Tiki, maybe this also helps for Dolibarr?

mercihabam:
@marclaporte this can definitely help for Dolibarr as well. It's also another opportunity to introduce more flexibility.

@camilevahviraki I'd be happy to help clarify how to leverage this and explore additional extension possibilities. If you'd like to pair on it, let's chat.

kroky:
Historically, we started Cypht integration with the idea that we have minimum amount of changes to request upstream (i.e. here in Cypht), so some of the integration details seem like patching core files. If you want to make cypht more modular and easier to integrate then there are definitely things to do but the current refactor seems way too big and prone to errors from my point of view. Removing class name constants is one example - it makes this less flexible and requires a considerable refactor on both sides.

I'd like to see 2 MRs on both sides with minimum amount of changes that allows Tiki integration code to not directly patch stuff - as vendor_path replace and css/js internal file modifications. Besides that, please keep everything else as it was. I'd like to avoid refactoring on the sake of refactoring but instead think about this strategically - what does Cypht need to change in order to have it more easily integrated into other software. Maybe @camilevahviraki can already tell us what is needed for Dollibar project integration.

mercihabam:
@kroky, here is my argument:

Class name constants are removed for the following reason:

When I scan the Cypht codebase on master, there is this almost unused module "site", described as:

> This module set provides sites a way to add features to Cypht, or to override existing behavior without modifying any Cypht code.

The need for class constants is to change default Cypht behaviour, so why not use the intended pattern rather than sneaking classes into the config files? This is wrong, in my opinion, because we end up with a way-too-large config file that isn't scope-focused.

So, with the current refactor, I think we keep the same flexibility, add a predictable structure to the customization process, and introduce a regression risk to downstream code that does not fully adhere to these changes.

If we want minimal changes, we would be skipping things surely, because Cypht can be customized in so many ways. One practical example is this tiny integration enhancement made in the past months: https://github.com/cypht-org/cypht/pull/1301. With an unnoticeable change to many users, it had changes spanning across a dozen files.

It would help reflect on this more if you could share additional points besides "the class name constants removal" on why you see these changes as not addressing the integration efficiently.

Thank you for your time!

kroky:
We are just heading in the wrong direction with a big refactor as I said. There were already several cases of such refactors in the past that made Cypht pretty unusable and cost us many hours to bring it back to a good shape. Besides regressions they also cause performance issues. Thus, I'd like to avoid big refactors unless they provide considerable value.

Site module is meant to add functionality in the form of custom modules to Cypht. Here, we are actually doing the reverse - integrating Cypht as kind of a module into an existing app. This means, we have to consider its core and how it handles requests, bundle classes, call dependencies, etc. Your example of js deps MR is a good one - the more we decrease dependencies or increase flexibility, the better.

That's my general idea and my main concern. We can also go over changes one by one but that will cost some time...

camilevahviraki:
> @mercihabam
> 
> FYI: @camilevahviraki is working on a proof of concept for a webmail module in Dolibarr: https://github.com/camilevahviraki/dolibarr-cyphtWebmail
> 
> As you are making Cypht cleaner to use for Tiki, maybe this also helps for Dolibarr?

@marclaporte, this will definitely help the Dolibarr module too. At the moment I've integrated Cypht as an iframe to avoid too much customization of the code in the Cypht Webmail Dolibarr module.

On that note, I'm also preparing a small fix upstream for one piece of this: Cypht's client-side AJAX requests (`Hm_Ajax_Request.xhr_fetch()` in site.js) always target `window.location.href`, which breaks when Cypht is embedded outside its own dispatch URL. I'll open a PR for it shortly

camilevahviraki:
> We are just heading in the wrong direction with a big refactor as I said. There were already several cases of such refactors in the past that made Cypht pretty unusable and cost us many hours to bring it back to a good shape. Besides regressions they also cause performance issues. Thus, I'd like to avoid big refactors unless they provide considerable value.

@kroky, @mercihabam, I'd propose splitting this PR so that each issue/refactor is in a dedicated Pull request, making it easier to follow in case of a regression.

mercihabam:
We tested this heavily to ensure that nothing breaks. But here the concern seems to be more about the approach followed.

@camilevahviraki and I just discussed and saw some valuable points that this could solve with the Dolibarr integration, and agreed to switch his POC to using this branch and see how that plays in practice.

I'm hoping we improve/change the bad approach here given the feedback we will have from the Dolibarr integration.

kroky:
I'd advise you start the dolibarr integration from master branch and see what exactly are you missing, then improve that in Cypht.

mercihabam:
@kroky that has already been done. The integration hit some limitations that we found out are addressed here.

kroky:
How are the class constant removal addressing these limitations? I feel like we have much more then what we need here and am against merging it. I'd like to see the dolibarr integration as well? Is anything on the dolibarr side changing? Adding Cypht in an iframe requires almost zero integration changes here but is also quite limited on the data share and transfer between the two systems.

mercihabam:
I don't understand why your main concern is class constant removal. This is something trivial that's being done for better harmony in the integration. Nevertheless, I have said that we can change that if it ends up not being better as I thought. I initially thought you were more concerned about things that could break with these changes, but seems like not.

The Dolibarr integration was started with an iframe and faced a few limitations. The suggestion was to only test this in Dolibarr, but I don't understand why that comes up as a big deal to you as well.

I could understand that you oversee this as much more than needed (but honestly, I don't think it's the case because I've seen its result in Tiki), so I'm leaving it to Dolibarr integration to prove otherwise.

kroky:
@mercihabam , due to your several big refactoring PRs here on this repo (I don't have time to list them one by one now but history can easily provide them) we faced a major issue with cypht 2.x and master including severe performance impact. Some were connected with the js handlers you introduced, some regarding other updates. If you really want to understand my concern, you can ask @marclaporte about the timesheets various developers have (including me) that was a direct result of your refactorings. Cypht has become completely unstable. I can't verify visually or logically just be doing code review. My concern is that no matter how much you test, we are still going to introduce needless regressions due to you insisting to merge  refactorings that are not needed. If you are not willing to start a new clean MR and do MINIMAL amount of updates just so Dolibarr or Tiki can integrate cleanly without doing preg_replace on core files, I will take care of that task and close this MR.

mercihabam:
A refactoring that was done many months/years ago and caused regressions can't suggest it's the case here too because evolution happens. But it seems like I can't convince of that. So, if closing this is the best action you can take, you might go ahead.

kroky:
So, you are not willing to consider adding another MR with more restricted set of updates?

mercihabam:
I'm not against going for another MR with fewer/more changes. But how to tell the changes we need if we can't even assess the current MR?

kroky:
I proposed 2 solutions:

1. start with Dolibarr - add cypht, ask - what is needed to integrate and add the changes.
2. start with Tiki - see what is causing issues - e.g. vendor_path and css/js overwrite - see how to make this configurable, make the changes.

Each change should have a clear answer to the question - why is it needed. Removing the class constants I keep giving as an example has no actual integration benefit and amount to almost half of the current PR changes (on a rough estimation).

mercihabam:
Ok, so I'll pick out the relevant work here in separate external MRs and eventually close this.

mercihabam:
Relevant changes here have been cherry-picked in individual PRs and merged into master. Also, more progress on the subject can be tracked here: https://github.com/cypht-org/cypht/issues/1909.

So, it's time to close this PR. Thanks to everyone!

PR #1919, opened 2026-04-07
feat(frontend/backend): support configurable page query parameter name

mercihabam:
Spin-off from: https://github.com/cypht-org/cypht/pull/1914


tests/phpunit/modules/core/output_modules.php
```diff
@@ -1003,28 +1003,28 @@ public function test_message_start() {
...
+        $this->assertEquals(array('<input type="hidden" class="msg_uid" value="5" /><div class="content_title"><a href="?page=search&list_path=search&list_page=1&sort=bar">Search</a><i class="bi bi-caret-right-fill path_delim"></i><a href="?page=message_list&list_path=">search<i class="bi bi-caret-right-fill path_delim"></i>bar</a></div><div class="msg_text">'), $res->output_response);
         $test->handler_response = array('list_page' => 1, 'list_filter' => 'foo', 'list_sort' => 'bar', 'uid' => 5, 'mailbox_list_title' => array('foo', 'bar'));
         $res = $test->run();
-        $this->assertEquals(array('<input type="hidden" class="msg_uid" value="5" /><div class="content_title"><a href="?page=message_list&amp;list_path=&list_page=1&filter=foo&sort=bar">foo<i class="bi bi-caret-right-fill path_delim"></i>bar</a></div><div class="msg_text">'), $res->output_response);
+        $this->assertEquals(array('<input type="hidden" class="msg_uid" value="5" /><div class="content_title"><a href="?page=message_list&list_path=&list_page=1&filter=foo&sort=bar">foo<i class="bi bi-caret-right-fill path_delim"></i>bar</a></div><div class="msg_text">'), $res->output_response);
```
kroky: &amp; html entity is syntactically correct here. We switch now back to ampersands which are susceptible to browser replacing them with valid entities (provided some are found). I think we should keep the old behavior. Ensure your link building functions construct and properly escape links but only when they are used in html context. If url is used for other purposes (curl calls, etc.), it should not be html encoded.

kroky:
Besides the encoding issue, everything else seems good to me.

tests/phpunit/modules/core/output_modules.php
mercihabam: That's a very good point. Thank you!

mercihabam:
> Besides the encoding issue, everything else seems good to me.

That's now addressed. Thanks!

kroky:
ok, but we have merge conflicts now, please resolve, thanks!

mercihabam:
Conflicts resolved, thanks!

PR #1920, opened 2026-04-07
test(unit): add tests for sievefilters

camilevahviraki:
### New
- Add  Unit Tests for sieve filter functionalities

### DIFF
--- a/tests/phpunit/modules/sievefilters/functions.php
+++ b/tests/phpunit/modules/sievefilters/functions.php
@@ -1,0 +1,163 @@
+<?php
+
+use PHPUnit\Framework\TestCase;
+
+class Hm_Test_Sieve_Functions_Client {
+    public static $scripts = array();
+
+    public function listScripts() {
+        return array_keys(self::$scripts);
+    }
+
+    public function getScript($name) {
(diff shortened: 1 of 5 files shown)


tests/phpunit/modules/sievefilters/functions.php
```diff
@@ -0,0 +1,314 @@
...
+        self::$scripts[$name] = $script;
+        return true;
+    }
+
+    public function getErrorMessage() {
+        return '';
+    }
+}
+
+class Hm_Test_Sieve_Functions_Save_Main_Script_Client extends Hm_Test_Sieve_Functions_Client {
```
kroky, line 26: Can you try to use phpunit mocks of the real classes with only functionality actually connecting to the sieve server being mocked? This way, you will test the classes as well instead of adding testing doubles here that hides some of the class functionality from the tests.

camilevahviraki: @kroky, Done. I have updated the Unit Tests

kroky:
A unit test is failing at the moment.

camilevahviraki:
> A unit test is failing at the moment.

@kroky , I have fixed the failing test

PR #1923, opened 2026-04-13
refactor(frontend): imap folders page to show wich folder is being edited

camilevahviraki:
cypht-folder-page

## Description
This PR redesigns the IMAP folders management page to make it easier to understand and manage folders per account. Instead of scattered per-server dialogs, each account now shows as an expandable card with a clear folder table. Folder actions (rename, delete, create subfolder, assign role) all open focused modals so the user always knows what they're acting on and gets feedback while the action runs.

### Issues
[https://avan.tech/item115321](https://avan.tech/item115321)

### Todo
- [X] Unit Tests

### DIFF
--- a/modules/imap_folders/site.js
+++ b/modules/imap_folders/site.js
@@ -1,3 +1,5 @@
+
+console.log('imap_folders/site.js loaded');
 'use strict';
 
 var folder_page_folder_list = function(container, title, link_class, target, id_dest, subscription = false) {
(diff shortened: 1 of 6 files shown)


modules/imap_folders/site.js
```diff
@@ -1,3 +1,5 @@
+
+console.log('imap_folders/site.js loaded');
```
kroky: All looks good here. Just remove the debugging code and we can merge.

camilevahviraki: Thanks @kroky, I have removed the console.log

Issue #1924, opened 2026-04-13
💬 "Inline Right" Message Preview not working despite correct configuration


Piskatje13:
<img width="980" height="201" alt="Image" src="https://github.com/user-attachments/assets/4504461e-b7d6-43ab-91c4-c053f684951d" />

Hello,

I am experiencing an issue with the inline message display feature in my Cypht installation. My settings are currently:

Cypht versions: 2.5.0 and 2.7.0
Browser I use: Firefox and Brave
Server environment: Local Docker (Portainer.io) on Zorin OS

Active Modules: core, contacts, local_contacts, feeds, imap, smtp, account, idle_timer, calendar, themes, nux, developer, history, saved_searches, advanced_search, highlights, profiles, inline_message, imap_folders, keyboard_shortcuts, tags

Description of the problem:
The "Inline Right" message preview (reading pane) does not appear in the interface. Despite the module inline_message being active and the settings being correctly configured in the UI, messages continue to open in full-screen/new-page mode instead of a split-pane layout.

Steps taken to resolve (without success):
Configuration Check: Verified that Show messages inline is checked and Inline Message Style is set to Right in the General Settings.

Module Verification: Confirmed inline_message is present in the CYPHT_MODULES environment variable and visible in the module list.

Current Result:
The setting is saved and remains active in the UI, but the layout does not change. There is no vertical divider or preview pane visible on the right side of the message list.

services:
  db:
    image: mariadb:10
    ports:
      - "3306:3306"
    volumes:
      - ./data/mysql:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=root_password
      - MYSQL_DATABASE=cypht
      - MYSQL_USER=cypht
      - MYSQL_PASSWORD=cypht_password
    healthcheck:
      test: ["CMD", "mysqladmin", "ping", "-h", "localhost", "-ucypht", "-pcypht_password"]
      interval: 10s
      timeout: 5s
      retries: 5

  cypht:
    image: cypht/cypht:2.7.0
    depends_on:
      db:
        condition: service_healthy
    ports:
      - "8080:80"
    environment:
      - AUTH_USERNAME=Admin
      - AUTH_PASSWORD=Password
      - DB_CONNECTION_TYPE=host
      - DB_DRIVER=mysql
      - DB_HOST=db
      - DB_NAME=cypht
      - DB_USER=cypht
      - DB_PASS=cypht_password
      - SESSION_TYPE=DB
[...]

PR #1930, opened 2026-04-24
fix(frontend): prevent SPA navigation from overriding inline message click

IrAlfred:
## Issue

When inline message mode is enabled, subject link clicks were intercepted
by the global `.cypht-layout a` navigation handler before the inline_message
module's capture_subject_click handler could act. This caused the full
message page to load (SPA navigate), briefly showing message content then
replacing the message list and this making inline mode non-functional.

Related issue: https://github.com/cypht-org/cypht/issues/1924


modules/core/navigation/navigation.js
```diff
@@ -53,7 +53,15 @@ $(document).on('click', '.cypht-layout a', function(event) {
         href.startsWith('https://') || 
         href.startsWith('//')
     ) && !href.includes(window.location.hostname);
-    
+
+    // When inline message mode is active, subject link clicks are handled by the
+    // inline_message module. Prevent full-page SPA navigation so the message is
+    // displayed inline instead of replacing the entire page content.
+    if (typeof inline_msg === 'function' && inline_msg() && $(this).closest('.subject').length) {
```
kroky: This is too much internal context leaked to a global event handler. I suggest you research 2 other possible approaches:
1. add special classes to inline msg links to avoid global navigation triggers. This can be reused in other modules as well. Something like `no-navigation` or similar.
2. change the order of event handlers, so inline msg handlers are executed before the global navigation ones

mercihabam: I’d like to suggest a third approach: when generating the message row and detecting the inline setting, use `#` for the href attribute and store the actual URL in a meaningful attribute (e.g., `data-src`). This way, navigation will properly ignore the link. This approach is already widely used in the codebase.

IrAlfred: Thank you @mercihabam for the suggestion. What do you think of this first approach I've implemented as suggested by @kroky

mercihabam: @IrAlfred I’d prefer to avoid introducing another branch, as it would add complexity. Maybe we can reuse existing approaches here instead.

kroky: Yes, that's definitely an option as well but it requires changes to the already established event handlers. I'd argue which one adds more complexity - `#` links with custom data attributes or no-navigation classes.

If the approach suggested by @mercihabam is already widely used, I think we should document it as the standard approach and surely add some comments to the global navigation handler, so people can know the right appraach to add links that are ignored by that in the future. I agree we should limit the ways we do one and the same thing in the codebase.

IrAlfred:
@mercihabam Please, I need your review here!

mercihabam:
@IrAlfred thank you for the changes. I'd advise generating elements with proper attributes from the backend instead of dynamically modifying them via JS. I think it's cleaner that way.

But also, the declaration of `get_inline_msg_details` should be modified so that the new attribute is used to extract the message details.

IrAlfred:
> @IrAlfred thank you for the changes. I'd advise generating elements with proper attributes from the backend instead of dynamically modifying them via JS. I think it's cleaner that way.
> 
> But also, the declaration of `get_inline_msg_details` should be modified so that the new attribute is used to extract the message details.

Thank you @mercihabam, I applied this. 
Try to check it again

mercihabam:
@IrAlfred, your changes looked good. I added one small change to fix the `capture_subject_click` function not getting triggered when navigating, which was hindered by the navigation not appending new scripts to the DOM.

mercihabam (approved):

IrAlfred:
> @IrAlfred, your changes looked good. I added one small change to fix the `capture_subject_click` function not getting triggered when navigating, which was hindered by the navigation not appending new scripts to the DOM.

Thank you @mercihabam

kroky:
I still don't see any documentation or comments in the global navigation code that shows what is the correct way of adding links that are not affected by it. Please add one.

IrAlfred:
> I still don't see any documentation or comments in the global navigation code that shows what is the correct way of adding links that are not affected by it. Please add one.

It's done !

Issue #23, opened 2015-10-09
Prepopulate port fields for "Add server"


R-J:
When I was adding IMAP servers I always had to type "993" into the port field. It would be nice if those port fields are prepopulated with meaningful values. I'd suggest
IMAP: 993
POP3: 995
SMTP: 465

jasonmunro:
Great idea! Added in https://github.com/jasonmunro/hm3/commit/483dcb39f7a36ff2f21f6b06ccd40e9f1b7a9329

PR #1935, opened 2026-04-25
fix(other): disable reply all when only recipient is current user

camilevahviraki:
This MR updates the logic for the "Reply All" button in the message view. Now, "Reply All" is disabled when the only recipient is the current user, preventing unnecessary replies to oneself.

Issue: [test case #23](https://github.com/cypht-org/cypht/wiki/Comprehensive-review-of-to-cc-bcc-when-reply-reply%E2%80%90to%E2%80%90all-forward-in-all-possible-contexts)


modules/imap/output_modules.php
```diff
@@ -367,25 +367,34 @@ protected function output() {
...
             if (array_key_exists('from', $lc_headers)) {
                 $imap_server_id = explode('_', $this->get('msg_list_path'))[1];
                 $server = Hm_IMAP_List::get($imap_server_id, false);
                 if ($server) {
-                    $addr_list = process_address_fld($lc_headers['from']);
-                    $addr_list = array_filter($addr_list, function ($addr) use($server) {
-                        return $addr['email'] != $server['user'];
-                    });
-                    $size += count($addr_list);
+                    $server_user = $server['user'];
```
kroky, line 408: This used to check the 'from' header and now it doesn't. Isn't this a regression?

marclaporte: @camilevahviraki Any thoughts?

camilevahviraki: @kroky,  It was a regression.
I have fixed it.

Thank you.

PR #1937, opened 2026-04-26
fix(backend): fix errors related to EWS when manipulating folders

Shadow243:
Some errors:
```
PHP Warning:  hex2bin(): Input string must be hexadecimal string in /cypht/modules/imap/hm-ews.php on line 218
PHP Warning:  hex2bin(): Input string must be hexadecimal string in /cypht/modules/imap/hm-ews.php on line 268
```

While assigning a folder to Trash, we used to get a long string instead of a human-readable string, e.g.:
Trash Folder: 
```
AAMkADk5Njk2ZjM0LTZmMWItNGIwZS05YTBjLTUzZWE4ZTExN2I5OQAuAAAAAADoTzuWq76VRIeNlYRsLpeUAQDwSoBO6tE6QZYs8Ixm76GUAAAzNmMdAAA=
```

### DIFF
--- a/modules/imap/hm-ews.php
+++ b/modules/imap/hm-ews.php
@@ -215,7 +215,7 @@ class Hm_EWS {
         if ($this->is_distinguished_folder($folder)) {
             $folderObj = new Type\DistinguishedFolderIdType($folder);
         } else {
-            $folderObj = new Type\FolderIdType(hex2bin($folder));
+            $folderObj = new Type\FolderIdType($folder);
         }
         $new_folder = new Type\FolderType();
         $new_folder->displayName = $new_name;
@@ -245,8 +245,7 @@ class Hm_EWS {
             if ($this->is_distinguished_folder($parent)) {
                 $parentObj = new Type\DistinguishedFolderIdType($parent);
             } else {
-                // Convert hex ID to binary for EWS
-                $parentObj = new Type\FolderIdType(hex2bin($parent));
+                $parentObj = new Type\FolderIdType($parent);
             }
             $request = [
                 'FolderIds' => Utilities\getFolderIds([$folderObj]),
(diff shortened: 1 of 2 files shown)


modules/imap/hm-ews.php
```diff
@@ -215,7 +215,7 @@ public function rename_folder($folder, $new_name, $parent = null) {
         if ($this->is_distinguished_folder($folder)) {
             $folderObj = new Type\DistinguishedFolderIdType($folder);
         } else {
-            $folderObj = new Type\FolderIdType(hex2bin($folder));
+            $folderObj = new Type\FolderIdType($folder);
```
kroky, line 218: You should be quite careful with this as normally, the hex representation is used throughout Cypht and passing around folder names in links, etc. while the bin representation is expected internally by EWS. If you receive a hex where you expect a bin representation in just one use-case - you should trace the issue back to the place here bin is sent - it could be somewhere in a quite different place in the code. I think the change here will introduce regressions in other places where rename_folder is already called with hex params.

Shadow243: We've noticed the following function:

```
/**
 * @subpackage imap/functions
 */
if (!hm_exists('decode_folder_str')) {
function decode_folder_str($folder) {
    $folder_name = false;
    $parts = explode('_', $folder, 3);
    if (count($parts) == 3) {
        $folder_name = hex2bin($parts[2]);
    }
    return $folder_name;
}}
```

It is called in Hm_Mailbox.php here:
`https://github.com/cypht-org/cypht/blob/master/modules/core/hm-mailbox.php#L170` for rename_folder, and
`https://github.com/cypht-org/cypht/blob/master/modules/core/hm-mailbox.php#L186` for delete_folder.

This appears to apply only to non-IMAP servers. From what we can see, `Hm_EWS::rename_folder` is only called via `Hm_Mailbox::rename_folder`, which already applies `decode_folder_str()` before passing the value. As a result, by the time it reaches EWS, the folder name is already in its decoded (binary) form, which is why we removed the `hex2bin` handling in `Hm_EWS`.

@kroky, could you please confirm if this understanding is correct, and whether there might be any other cases we may have overlooked ?

kroky: That looks correct. Thanks for checking.

Issue #94, opened 2016-06-28
Create a brute force login protection module set


jasonmunro:
lots of great ideas on this here:
https://www.owasp.org/index.php/Blocking_Brute_Force_Attacks

marclaporte:
@Danelif please advise.

Danelif:
Alright

Danelif:
I have read this article carefully https://www.owasp.org/index.php/Blocking_Brute_Force_Attacks and found it very relevant. But some techniques are not included in it.
If we want to create brute-force login protection, consider including the (2-3) FA technique (2-3 Factor Authentication). The user might provide a unique OTP sent to his email address or mobile phone once the username/password is correct. 
Also, The limitation of attempting to log in can be considered from a single IP address. If the limit is reached we can suggest the user to recover it password by emailing an OTP to the email in our database.

marclaporte:
@Danelif Thank you, please look at how it is done in Tiki to get some more good ideas.

Danelif:
In tiki 2FA is done using Google2FA php library. Good idea

marclaporte:
Some docs:
- https://doc.tiki.org/PluginTOTP
- https://doc.tiki.org/Two-factor-authentication

TOTP uses time, so the code changes every 30 seconds.

Danelif:
@marclaporte I have seen how 2FA works in tiki. It could be great to to the same in cypht

marclaporte:
ok, please proceed as a medium priority. High priority is fixing bugs before adding new features.

Danelif:
PR: https://github.com/cypht-org/cypht/pull/1938

PR #1938, opened 2026-04-27
feat(backend): add brute force login lockout protection

Danelif:
This PR adds a brute_force module to protect against repeated login failures. The module tracks failed logins by IP address and username, locks requests after the configured threshold, and clears counters after a successful login.

Changes: 

Added brute_force handler module.
Added configurable lockout settings:
BRUTE_FORCE_MAX_ATTEMPTS
BRUTE_FORCE_LOCKOUT_DURATION
BRUTE_FORCE_DATA_DIR
Enabled brute_force in the default module list.
Added PHPUnit coverage for lockout tracking, blocking, clearing, and active-session edge cases.
Ignored the runtime lockout JSON file.

image
image
image

Issue : https://github.com/cypht-org/cypht/issues/94

### DIFF
--- a/modules/brute_force/modules.php
+++ b/modules/brute_force/modules.php
@@ -1,0 +1,398 @@
+<?php
+
+/**
+ * Brute force login protection modules
+ * @package modules
+ * @subpackage brute_force
+ */
+
+if (!defined('DEBUG_MODE')) { die(); }
+
+/**
+ * Database-backed tracker for failed login attempts.
+ * Stores attempt counts keyed by IP address and username (hashed).
+ * @subpackage brute_force
+ */
+class Hm_Brute_Force_Tracker {
+
+    /**
+     * Database connection
+     * @var PDO|false
+     */
+    private $dbh;
+
+    /**
+     * Create a tracker for the configured site database
+     * @param object $config site config
+     * @return void
+     */
+    public function __construct($config) {
+        $this->dbh = Hm_DB::connect($config);
+    }
+
+    /**
+     * Load attempt data from the database
+     * @return array
+     */
+    public function load() {
+        if (!$this->dbh) {
+            return [];
+        }
+        $rows = Hm_DB::execute($this->dbh, 'select attempt_key, attempt_count, locked_until, last_attempt from hm_login_attempts', [], false, true);
+        if (!is_array($rows)) {
+            return [];
+        }
+        $data = [];
+        foreach ($rows as $row) {
+            $data[$row['attempt_key']] = [
+                'count' => (int) $row['attempt_count'],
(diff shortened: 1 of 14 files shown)


modules/brute_force/modules.php
```diff
@@ -0,0 +1,277 @@
...
+class Hm_Brute_Force_Tracker {
+
+    /**
+     * Return the path to the attempts data file
+     * @param object $config site config
+     * @return string
+     */
+    public static function get_file($config) {
+        $dir = $config->get('brute_force_data_dir', sys_get_temp_dir());
+        return rtrim($dir, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . 'cypht_failed_logins.json';
```
kroky: We already have sqlite, mysql or postgre support. Can we use a table there to store these?

Danelif: sure let me do this

Danelif: pushed changes now failed logins will be  stored on databases

modules/brute_force/modules.php
```diff
@@ -0,0 +1,278 @@
...
+ */
+class Hm_Brute_Force_Tracker {
+
+    /**
+     * Return a DB connection for the configured site database
+     * @param object $config site config
+     * @return PDO|false
+     */
+    private static function connect($config) {
+        return Hm_DB::connect($config);
```
kroky: Cache this and reuse as a local private variable. Even though connect db method caches the instance, it still looks very weird to connect each time here you use the db. Convert all those methods to non-static methods and reuse context as object variables.

modules/brute_force/modules.php
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * Record a failed attempt for a key
+     * @param array $data
+     * @param string $key
+     * @param int $max_attempts
+     * @param int $lockout_duration seconds
+     * @return array updated data
+     */
+    public static function record_failure($data, $key, $max_attempts, $lockout_duration) {
```
kroky: methods like this should save the data, not only modify it. Otherwise, you are breaking the abstraction and make many different methods and classes work on the same set of data -> makes the maintenance of the code quite harder.

Danelif: Thanks, I’ve refactored this so the tracker now saves the changes itself instead of passing modified data back to the handler. I also updated the related tests, and the brute-force test suite is passing now.

modules/brute_force/modules.php
Danelif: Thanks, updated this now. The tracker now keeps the DB connection on the object, and the methods are instance-based instead of static.

PR #1942, opened 2026-04-28
fix(e2e): fix CI sqlite setup and replace flaky navigation wait with deterministic counter + AJAX idle check

IrAlfred:
### Issue 1: CI login failures (sqlite DB_NAME misconfiguration)

All Selenium tests were failing in CI at `good_login` and `good_logout`. The setup script `.github/tests/setup.sh` patched `DB_DRIVER=sqlite` in the `.env` file but left `DB_NAME=cypht_test`. Cypht builds its SQLite DSN as `sqlite:{DB_NAME}`, so it tried to open a file literally named `cypht_test` with no tables, causing every authentication attempt to fail.

Related to this change: https://github.com/cypht-org/cypht/pull/1929

### Issue 2: Intermittent navigation timing failures

Selenium tests were failing intermittently on `load_profile_page`, `flagged`, and other navigation-dependent tests in CI and locally. The root cause was a race condition in `wait_for_navigation_to_complete()`:

The previous implementation checked `window.routingToast === null` to detect when SPA navigation had finished. Since `routingToast` is `null` both _before_ a navigation starts and _after_ it completes, the check would often pass immediately — causing the test to assert on page content before it had actually loaded.

A second issue compounded this: some pages (e.g. Profiles) load their actual content via a background AJAX call _after_ the page shell is injected by `navigate()`. No wait existed for that second async step.

### DIFF
--- a/tests/selenium/base.py
+++ b/tests/selenium/base.py
@@ -23,6 +23,7 @@ import os
 class WebTest:
 
     driver = None
+    _pre_click_nav_count = None
 
     def __init__(self, cap=None):
         self.read_ini()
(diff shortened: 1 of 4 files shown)


tests/selenium/base.py
```diff
@@ -228,16 +229,28 @@ def wait_on_sys_message(self, timeout=60):
 
     def wait_for_navigation_to_complete(self, timeout=60):
         print(" - waiting for the navigation to complete...")
-        # Wait for the main content to be updated and any loading indicators to disappear
+        # Wait for the SPA navigation to finish. The counter is captured before
+        # the triggering click and checked here so we wait for the specific
+        # navigation we initiated, not a previous one.
+        expected = (self._pre_click_nav_count or 0) + 1
```
kroky, line 235: Thanks for that! I hope we will finally be able to close that intermittent navigation issue in selenium tests...

PR #1945, opened 2026-05-02
fix(other): add validation for hex2bin in decode_folder_str

christer77:
This PR fixes an ErrorException: hex2bin(): Input string must be hexadecimal string. The error occurred when the `decode_folder_str` function tried to convert a string that was not in a valid hexadecimal format or whose length was odd.

### DIFF
--- a/modules/imap/functions.php
+++ b/modules/imap/functions.php
@@ -1301,7 +1301,10 @@ function decode_folder_str($folder) {
     $folder_name = false;
     $parts = explode('_', $folder, 3);
     if (count($parts) == 3) {
-        $folder_name = hex2bin($parts[2]);
+        $hex_part = $parts[2];
+        if (ctype_xdigit($hex_part) && strlen($hex_part) % 2 == 0) {
+            $folder_name = hex2bin($hex_part);
+        }
     }
     return $folder_name;
 }}


modules/imap/functions.php
```diff
@@ -1312,7 +1312,10 @@ function decode_folder_str($folder) {
     $folder_name = false;
     $parts = explode('_', $folder, 3);
     if (count($parts) == 3) {
-        $folder_name = hex2bin($parts[2]);
+        $hex_part = $parts[2];
+        if (ctype_xdigit($hex_part) && strlen($hex_part) % 2 == 0) {
+            $folder_name = hex2bin($hex_part);
+        }
```
kroky, line 1318: hex2bin and bin2hex are used to encode folder names and other ID strings that might contain binary data or unsafe data for urls. It was the originally chosen way to encode and even though there exist a more appropriate url encoding schemes, we kept it in all modules (imap, ews, smtp). The most probable reason that here it tries to decode a hex string that is not a hex string is that something is passing the plain value and not the hex string. Your changes will mask the problem and not fix it. Please see the stack trace of the bug you experience and check what is passing a plain name instead of a hex-encoded name and fix the issue where appropriate.

christer77: Noted

IrAlfred:
This issue doesn't occur anymore and the potential root cause has been fixed in these two pull request https://github.com/cypht-org/cypht/pull/1923 and https://github.com/cypht-org/cypht/pull/1937

Issue #1466, opened 2025-03-09
🐛 [Bug] submission (smtp) authentication with scram tries using imap AUTHENTICATE command instead of smtp AUTH command


mjl-:
## 🐛 Bug

I was setting up cypht 2.4.0 to test for interoperability with https://github.com/mjl-/mox. When adding a submission/smtp server for outgoing email, cypht tries to log in to check the credentials (presumably). It tries to pick SCRAM (which is good!). However, it tries to use command AUTHENTICATE on smtp, which does not exist in smtp. It does in IMAP. I suspect the SCRAM code is shared between IMAP and SMTP, but doesn't take this difference into account.

This is what the mox submission server is seeing:

```
> 220 komijn.test.xmox.nl ESMTP mox v0.0.15-0.20250308080341-0857e81a6ccc+dirty-go1.24.1\r\n
< EHLO 0f5831cd80bb\r\n
> 250-komijn.test.xmox.nl\r\n250-PIPELINING\r\n250-SIZE 104857600\r\n250-REQUIRETLS\r\n250-AUTH SCRAM-SHA-256-PLUS SCRAM-SHA-256 SCRAM-SHA-1-PLUS SCRAM-SHA-1 CRAM-MD5 PLAIN LOGIN\r\n250-FUTURERELEASE 5184000 2025-05-08T11:28:17Z\r\n250-ENHANCEDSTATUSCODES\r\n250-8BITMIME\r\n250-LIMITS RCPTMAX=1000\r\n250 SMTPUTF8\r\n
< AUTHENTICATE SCRAM-SHA-1\r\n\r\n
> 500 5.5.1 unknown command (T1ziWP-GGO7qrINl8MLFNw)\r\n
> 500 5.5.1 unknown command (T1ziWP-GGO7qrINl8MLFNw)\r\n
```

https://github.com/cypht-org/cypht/blob/1ead675272181d5b9e4cf5672fd30f3acc1b4c07/modules/smtp/hm-smtp.php#L345

https://github.com/cypht-org/cypht/blob/1ead675272181d5b9e4cf5672fd30f3acc1b4c07/lib/scram.php#L51

Btw, if I'm reading the code right, the scram mechanisms are chosen in order from weakest to strongest. Is that intentional? See:

https://github.com/cypht-org/cypht/blob/1ead675272181d5b9e4cf5672fd30f3acc1b4c07/modules/smtp/hm-smtp.php#L122

[...]

PR #1959, opened 2026-05-14
fix(module/smtp): correct SCRAM authentication command and mechanism selection

IrAlfred:
Related issue: https://github.com/cypht-org/cypht/issues/1466


Neustradamus:
@IrAlfred: Thanks for your PR!

cc: @mjl-.

Issue #412, opened 2020-06-25
Better handling of spam folder: Special folder, aggregate view, tag as "not spam", etc.


marclaporte:
I received an email. Went in spam folder. I didn't think to check. It was an important email. Not spam. I had nothing prompting me to go check there. Not good.

We should:
- Have a obvious way to see if there is new/unread spam.
- Add Spam as a special folder like Trash and Sent, knowing that an external process is likely set up to move suspected spam.
- Have a button to identify a message as "not spam" and move it back to relevant inbox. And if possible, teach the server that it's not spam (is there a standard for this?)

Thanks!

dumblob:
>Have a button to identify a message as "not spam" and move it back to relevant inbox. And if possible, teach the server that it's not spam (is there a standard for this?)

If there is any standard for that, I'm ok with this proposal. But if not (and I don't know of any), I'm **strongly against** yeat another button (or any UI extension in this regard) because the functionality is equivalent to "move to another folder" (duplicating functionality is UX anti-pattern).

Yamakasi:
> 
> 
> > Have a button to identify a message as "not spam" and move it back to relevant inbox. And if possible, teach the server that it's not spam (is there a standard for this?)
> 
> If there is any standard for that, I'm ok with this proposal. But if not (and I don't know of any), I'm **strongly against** yeat another button (or any UI extension in this regard) because the functionality is equivalent to "move to another folder" (duplicating functionality is UX anti-pattern).

I thought directly, did Marc not asked for the same a couple of months ago ?

marclaporte:
Just happened again. spam folder is too hidden.

ElvisAns:
I join the thread... I hope to find a solution with your help

marclaporte:
Here is how Gmail and Roundcube do it:
![2022-09-24_122529](https://user-images.githubusercontent.com/1004261/192109135-c7e53197-a010-4043-961e-805ce610f9c3.png)
![2022-09-24_122152](https://user-images.githubusercontent.com/1004261/192109136-62a00592-a66a-45a0-a68c-0710bd82e119.png)

marclaporte:
https://support.mozilla.org/en-US/kb/thunderbird-and-junk-spam-messages#w_tell-thunderbird-what-is-not-junk
http://kb.mozillazine.org/Junk_Mail_Controls

NonJunk is visible here:
https://datatracker.ietf.org/doc/rfc5232/
https://datatracker.ietf.org/doc/rfc9051/
[...]

PR #1961, opened 2026-05-15
feat(frontend/backend): add button mark a message as not spam/junk

Grandi0z:
[Related task](https://avan.tech/item96465)

Adds a “Not junk” action for Junk folders, allowing messages to be moved back to Inbox (mirror of the existing Junk action).

The control appears only in junk-related views. For` IMAP` accounts, spam-related flags are cleaned up before the move when supported by `PERMANENTFLAGS` (fail-soft behavior).

Also adds unread badges for Junk folders in the navigation


kroky:
702 lines added for a simple feature as "not junk". Please revise. Permanent keywords should not be needed here. Not just action is available in junk folder and what it does is simply - remove junk flag of the msg + move to inbox. We already have both actions available implemented, so I don't expect an MR with more than 40-50 lines of changes...

Grandi0z:
> 702 lines added for a simple feature as "not junk". Please revise. Permanent keywords should not be needed here. Not just action is available in junk folder and what it does is simply - remove junk flag of the msg + move to inbox. We already have both actions available implemented, so I don't expect an MR with more than 40-50 lines of changes...

I've simplified the implementation significantly.

Removed:

PERMANENTFLAGS support
$NotJunk / $Junk keyword handling
REMOVE_KEYWORD action
RFC9051-related logic

The Not Junk action now simply:

clears the IMAP \Junk flag
reuses the existing restore flow to move the message back to Inbox

I kept the unread Junk badge because it addresses the first requirement from the original issue: [412](https://github.com/cypht-org/cypht/issues/412)

marclaporte:
Joseph Kausi is no longer available. So I put this in draft, and we can re-explore when other devs are less busy (ex.: Alfred and Steven)

Issue #1982, opened 2026-05-27
🐛 [Bug] Unread mails foldr amd counter not updated


B3rt:
The unread mail counter stays always on 0
When opening the mailbox itself if shows new mails, but the total unread stay 0

Worked fine for 1 or 2 weeks and then it suddenly stopped working.

## 🐛 Bug

<img width="401" height="489" alt="Image" src="https://github.com/user-attachments/assets/9efb9e15-9639-40e8-b7ac-706de92bd908" />

<img width="342" height="583" alt="Image" src="https://github.com/user-attachments/assets/a1a74012-26b9-4162-a22e-79e85bc61eef" />

### Version & Environment
OS:  Docker latest  

<img width="319" height="62" alt="Image" src="https://github.com/user-attachments/assets/9528d8ac-54a2-4471-96ef-aec5b83cc097" />

IrAlfred:
Thank you @B3rt for submitting this issue, we are going to look into it.

IrAlfred:
The issue has been resolved. We're closing this ticket now, but you're welcome to reopen it if the problem persists.

PR #1988, opened 2026-05-28
fix(other): prevent unread total from resetting to zero

FrizkyTrixter:
## Summary

This PR fixes an issue where the global unread counter would incorrectly reset to `0` when opening the `Unread` page, even though the actual mail folders still contained unread messages.

While reproducing issue #1982 locally, I noticed that the unread badge briefly displayed the correct value and then immediately reset to `0`. After tracing the JavaScript logic, I found that the unread counter was being recalculated using the number of visible unread rows in the current message list view instead of the actual unread folder counts stored internally by Cypht.

The problematic logic was inside:

```js
this.set_message_list_state = function(list_type)
```

Specifically:

```js
if (list_type == 'formatted_unread_data') {
    self.adjust_unread_total(Hm_Utils.rows().length, true);
}
```

This recalculated the unread total from the currently rendered rows, which could evaluate to `0` on the Unread page even when unread messages still existed in IMAP folders.

## Fix

Instead of recalculating unread totals from visible rows, this PR updates the unread counters using the existing folder unread data already stored in:

```js
Hm_Folders.unread_counts
```

The logic was changed to:

```js
if (list_type == 'formatted_unread_data') {
    Hm_Folders.update_unread_counts();
}
```

Additionally, the unread total calculation was updated to aggregate unread counts only from actual mail source folders:

* IMAP
* JMAP
* EWS

using:

```js
name.startsWith('imap_')
name.startsWith('jmap_')
name.startsWith('ews_')
```

This ensures the unread badge reflects the real unread counts from connected mail sources rather than the currently displayed rows in the UI.

## Testing

Tested locally on Windows/XAMPP with an IMAP account containing unread messages.

Verified that:

* the global unread badge updates correctly
* the unread badge no longer resets to `0`
* unread totals persist after refresh
* marking emails read/unread updates counts correctly
* Inbox and Unread views remain synchronized
* IMAP folder unread counters still function correctly

I was able to consistently reproduce the original issue before the fix and verify the corrected behavior afterward.


IrAlfred:
Thank you, @FrizkyTrixter, for your first contribution to this project. It will be reviewed shortly.

IrAlfred:
Hey @FrizkyTrixter  !

Welcome again, and thanks so much for your recent first contribution! 🎉

We just wanted to follow up now that things have settled a bit. You probably saw that we recently shipped Cypht 2.11.0 – and your work is now officially part of it! We'd love to hear your thoughts on the release, especially since you've just gotten your hands dirty with the codebase. Any feedback (good or bad) is super welcome.

Also, if you'd like to stick around and stay connected with the team, feel free to hop into our community chat on Gitter: https://gitter.im/cypht-org/community – it's a great place for casual dev chat and quick questions.

We also hold monthly community meetings (https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings) – it would be awesome to have you join and share your perspective live, now that you're familiar with the project flow.

Looking forward to seeing you around!

PR #1998, opened 2026-06-09
refactor(frontend/backend): add single data sources to specific combined views

### LINKED ISSUE #283: Add single data sources to specific combined views
Currently the ability to control what data sources are included in combined views is very limited:

- IMAP folders can be added to combined views (INBOX folders are included automatically, but can be excluded after the fact). But only to all combined views, or none
- RSS feeds, WordPress notifications, and Github notices can be excluded from the Unread combined view, but not the others

It would be great to be able to include any data source to any combined view page.

josaphatim:
- fixes #283

### DIFF
--- a/modules/core/js_modules/Hm_MessagesStore.js
+++ b/modules/core/js_modules/Hm_MessagesStore.js
@@ -3,6 +3,8 @@
  */
 class Hm_MessagesStore {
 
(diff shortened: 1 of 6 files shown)


marclaporte:
@josaphatim Some tests are failing

PR #1999, opened 2026-06-09
fix(backend): "Uncaught Error: Cannot use object of type stdClass as array" thrown when deleting a folder from a mailbox with active Sieve support

mercihabam:


modules/sievefilters/functions.php
```diff
@@ -615,6 +615,10 @@ function get_sieve_linked_mailbox ($imap_account, $module) {
             $scripts = $client->listScripts();
             $folders = [];
             foreach ($scripts as $s) {
+                if (! mb_strstr($s, 'cypht')) {
```
kroky: Script name might not contain the keyword 'cypht'. I am not sure what are you trying to do here. If you see an error because we are handling a sieve script that is an external one not defined by Cypht, then we need a better way to distinguish between our own Cypht-managed sieve scripts and external ones. Hardcoding keywords is never a good idea. Maybe we use a specific script header or filter namespace prefix?

mercihabam: The error is thrown because non-filter scripts like blocked_senders/main_script/Default are not parsed like normal filter payloads. In other places where metadata isn't available, the 'cypht' keyword is used and serves to distinguish between non-filter scripts and filter ones. But using a specific script header here is a good idea, so now using `'CYPHT GENERATED CONDITION'` as the inclusion criteria.

Issue #1871, opened 2026-02-17
🐛 [Bug] Settings Save page save permanently - not working cannot explicitly allow it


PeopleInside:
## 🐛 Bug

On my .env file I have:
`DEFAULT_SETTING_NO_PASSWORD_SAVE=true`

But If I made settings change, for example the language and save the page then going into ?page=save I'm asked for password and some settings appear to be saved because an edit has been done.

I'm in production mode.
For save settings I have to save the page after edit settings (for example language) then to made permanent I need to go at `?page=save` insert password and save.

As I set in the .env file 
`DEFAULT_SETTING_NO_PASSWORD_SAVE=true`

I should not going to the page ?page=save to save.

If I dont do this all settings edit are lost after logout

### Version & Environment

Version 2.6.0 (17 February 2026)

OS:  Ubuntu 24 LTS

IrAlfred:
@PeopleInside 

The environment variable `DEFAULT_SETTING_NO_PASSWORD_SAVE` does **not** control whether user preferences/settings need to be saved via the `?page=save` page. This setting only controls whether **email account passwords (IMAP/SMTP)** are persisted between sessions.

### What This Setting Actually Does

**When `DEFAULT_SETTING_NO_PASSWORD_SAVE=true`:**
- Email server passwords (IMAP/SMTP) are NOT saved on the server
- Users must re-enter their email account passwords each session
- **More secure** but less convenient

**When `DEFAULT_SETTING_NO_PASSWORD_SAVE=false`:**
- Email server passwords ARE saved on the server  
- Users don't need to re-enter email passwords after logging in
- More convenient but less secure

### Expected Behavior (Working as Designed)

The requirement to visit `?page=save` and enter your password to permanently save settings is **intentional by design** for security reasons:

1. **During session:** Settings changes are temporary (stored in session only)
2. **To persist permanently:** Users must explicitly navigate to `?page=save`, enter their password, and click "Save"
3. **Without permanent save:** All changes are discarded on logout

**There is no environment variable (for now) to bypass the `?page=save` workflow.** This is a core security feature of Cypht. To permanently save any settings changes (language, timezone, etc.), you must:

1. Make your changes in settings
2. Navigate to `?page=save`
3. Enter your password
4. Click "Save" or "Save and Logout"

PeopleInside:
Understood. Many time i loose settings and thing because of that. Pressed save and was thinking was saved then lost all.
[...]

Issue #1815, opened 2025-12-23
🐛 [Bug] Setting is not saved correctly


kevinend51:
## 🐛 Bug

Here's my cypht's docker compose:
```
services:
  mariadb:
    image: mariadb:12.1.2
    restart: unless-stopped
    volumes:
      - cypht-mysql:/var/lib/mysql
    healthcheck:
      test: healthcheck.sh --connect --innodb_initialized
    environment:
      MYSQL_ROOT_PASSWORD: XXXXX
      MYSQL_DATABASE: cypht
      MYSQL_USER: cypht
      MYSQL_PASSWORD: XXXXX
      MARIADB_AUTO_UPGRADE: true
  cypht:
    depends_on:
      mariadb:
        condition: service_healthy
    image: cypht/cypht:2.5.1
    ports:
      - XXXXX
    environment:
      DB_CONNECTION_TYPE: host
      DB_DRIVER: mysql
      DB_HOST: mariadb
      DB_NAME: cypht
      AUTH_TYPE: LDAP
      DB_USER: cypht
      DB_PASS: XXXXX
      SESSION_TYPE: LDAP
      USER_CONFIG_TYPE: DB
      LDAP_AUTH_PORT: XXXXX
      LDAP_AUTH_SERVER: XXXXX
      LDAP_AUTH_TLS:
      LDAP_AUTH_BASE_DN: XXXXX
      LDAP_UID_ATTR: cn
      LDAP_AUTH_UID_ATTR: cn
volumes:
  cypht-mysql:
```

After a fresh start, I perform the following step:
1. Login with LDAP account
2. Add a email (gmail)
3. Save settings with entering LDAP password again to confirm
4. Open a new private tab, login with the same LDAP account, it shows "no email setup":
    
<img width="1853" height="970" alt="Image" src="https://github.com/user-attachments/assets/fe7f1aad-8e0d-4d89-b065-bef7aeefe4db" />

### Version & Environment

Rev: cypht/cypht:2.5.1

OS:  []

IrAlfred:
Thank you @kevinend51 for reporting this, we are going to investigate about it !

IrAlfred:
Hi @kevinend51 

We have been unable to reproduce this issue on our end. Could you please retest to confirm it's still occurring?

If the problem persists, we can schedule a brief screenshare session to investigate further. To coordinate, please send me a direct message (DM) on Matrix at https://matrix.to/#/@iralfred:gitter.im (user: @iralfred).

[...]

PR #2007, opened 2026-06-13
feat(frontend/backend): add feature auto save user settings

IrAlfred:
## Summary

Cypht stages configuration changes in the session and normally requires a visit to ?page=save with a password before anything is written to disk. This PR adds an optional Automatically save settings preference so users who opt in no longer need that extra step.

When enabled, the login password is stored in the encrypted session for the current session only and reused to persist settings after changes. On logout, any pending changes are saved before the session is destroyed. The manual save page and flow remain available for users who do not enable auto-save.

Related issues: https://github.com/cypht-org/cypht/issues/1871 https://github.com/cypht-org/cypht/issues/1815#issuecomment-3829491743


kroky:
500+ changed lines for this feature seems quite a lot to me. Also, I think it is fragile at the moment - changing many handler modules, js code places ,etc. while we have a central place where all the encryption and config storage goes. Yet another problem I see is that it totally defeats the purpose of requesting a password from the user - it is meant to confirm that the user behind the scenes is the actual user and only then save the settings - standard procedure in many web apps. I think it was broken in some aspects (reading the comments you reference), so people did not notice or did not get that warning that they have unsaved changes. This bug should be fixed.

Regarding the auto-save as an option, I agree it is preferred by many users, so we can update cypht to support it. I propose a much simpler way to implement this:
1. Have a global cypht installation setting (valid for all users) - store encrypted config - true or false. If true, password is required on config save and read. This is the current way cypht works and should not require any changes. If false, modify the config storage and read operations - store as plain json encoded string. It already has crypt_state method that switched between the two modes, so implementation here should be quite straightforward.
2. Upon reading or saving config - check also the current state of the file/db entry. If it is not decodable, then still require a password. This is to prevent users from locking out of the system if they change the global setting but have their existing configuration encrypted.
3. Add help text to the global setting that it basically switches cypht into 2 modes - first one is more secure and stores encrypted configuration on disk but requires users to manually save operations via password entry. Second one is less secure but gives users the ability to auto-save the changes without the risk of loosing them. Most people will prefer the second approach.

IrAlfred:
> 500+ changed lines for this feature seems quite a lot to me. Also, I think it is fragile at the moment - changing many handler modules, js code places ,etc. while we have a central place where all the encryption and config storage goes. Yet another problem I see is that it totally defeats the purpose of requesting a password from the user - it is meant to confirm that the user behind the scenes is the actual user and only then save the settings - standard procedure in many web apps. I think it was broken in some aspects (reading the comments you reference), so people did not notice or did not get that warning that they have unsaved changes. This bug should be fixed.
> 
> Regarding the auto-save as an option, I agree it is preferred by many users, so we can update cypht to support it. I propose a much simpler way to implement this:
> 
> 1. Have a global cypht installation setting (valid for all users) - store encrypted config - true or false. If true, password is required on config save and read. This is the current way cypht works and should not require any changes. If false, modify the config storage and read operations - store as plain json encoded string. It already has crypt_state method that switched between the two modes, so implementation here should be quite straightforward.
> 2. Upon reading or saving config - check also the current state of the file/db entry. If it is not decodable, then still require a password. This is to prevent users from locking out of the system if they change the global setting but have their existing configuration encrypted.
> 3. Add help text to the global setting that it basically switches cypht into 2 modes - first one is more secure and stores encrypted configuration on disk but requires users to manually save operations via password entry. Second one is less secure but gives users the ability to auto-save the changes without the risk of loosing them. Most people will prefer the second approach.

Thank you for the comprehensive guidelines, @kroky. I will implement them.

PR #2015, opened 2026-06-19
fix(backend): combined pages not loading because `enable_exclude_auto_bcc_setting` is mishandled

mercihabam:
Related: https://github.com/cypht-org/cypht/pull/1740


kroky:
This is actually related/a fix for https://github.com/cypht-org/cypht/pull/1733 that was a spinoff from 1740...

kroky:
@mercihabam actually explains the issue well here: https://github.com/cypht-org/cypht/pull/1733#discussion_r3443630449

PR #2020, opened 2026-06-23
docs(docker): switch to php 8.3 for docker version

camilevahviraki:
### Issue
- The Docker image was based on Debian with PHP 8.1, which is [End Of Life](https://www.php.net/supported-versions.php)


IrAlfred:
@camilevahviraki Please update this too https://github.com/cypht-org/cypht/blob/master/docker/DOCKERHUB-README.md

Shadow243:
> @camilevahviraki Please update this too https://github.com/cypht-org/cypht/blob/master/docker/DOCKERHUB-README.md

@camilevahviraki, what @IrAlfred was referring to is here: https://github.com/cypht-org/cypht/blob/master/docker/DOCKERHUB-README.md?plain=1#L7

it should be something like this: `* Debian (bookworm) based image`

PR #2024, opened 2026-06-25
fix(module/feeds): block SSRF to private and internal feed URLs

IrAlfred:
https://avan.tech/item165627#threadId107239


modules/feeds/modules.php
```diff
@@ -366,17 +366,22 @@ public function process() {
                 if (feed_exists($form['new_feed_address'])) {
                     Hm_Msgs::add(sprintf('Feed url %s already exists', $form['new_feed_address']), 'warning');
                 } else {
-                    $connection_test = address_from_url($form['new_feed_address']);
-                    if ($con = @fsockopen($connection_test, 80, $errno, $errstr, 2)) {
+                    if (!feed_url_is_allowed($form['new_feed_address'])) {
+                        Hm_Msgs::add('Feed address is not allowed', 'danger');
+                    } else {
                         $feed = is_news_feed($form['new_feed_address']);
                         if (!$feed) {
```
kroky, line 373: Previously it tried an actual connection to the address and now it just tries to get the contents and skips otherwise - shouldn't it show an error in case the remote address is not a news feed?

tests/phpunit/modules/feeds/OpmlTest.php
```diff
@@ -194,15 +194,14 @@ public function testUrlFallback()
...
     </body>
 </opml>';
         
-        require_once APP_PATH.'modules/feeds/hm-opml.php';
         $parser = new Hm_Opml_Parser();
         $parser->parse($opml);
         $feeds = $parser->getFeeds();
         
-        $this->assertEquals('https://example.com/feed.xml', $feeds[0]['name']);
+        $this->assertEquals('https://93.184.216.34/feed.xml', $feeds[0]['name']);
```
marclaporte: Is this an intentional change? What is this IP?

modules/feeds/modules.php
IrAlfred: The old removed fsockopen(..., 80) was only a TCP reachability check on port 80, not a feed validation. The actual feed test has always been is_news_feed() (HTTP fetch + XML parse).

We removed fsockopen because it was the SSRF from the security report and was redundant: is_news_feed() already performs the real HTTP request. Non-feed URLs still get errors as previously

tests/phpunit/modules/feeds/OpmlTest.php
IrAlfred: Yes, it's intentional. validateUrl() now delegates to feed_url_is_allowed(), which resolves hostnames before accepting a URL. The old example.com fixtures failed in environments without DNS (including our CI sandbox), because resolution returns nothing and the URL is rejected.

93.184.216.34 is the public IP historically used for example.com; we use it only as opaque valid feed URL input in tests. We never connect to it. Same for 1.1.1.1 where we need a second distinct URL.

modules/feeds/modules.php
kroky: The code is a bit confusing to read this way but I see how it is handled now - checking first for negative match...

tests/phpunit/modules/feeds/OpmlTest.php
kroky: Can we make them constants, at least? Also, add some comments explaining why we are using these specific ones.

IrAlfred:
@kroky Could you please take another look at this PR

PR #2027, opened 2026-06-30
fix(module/imap): improve MSTNEF attachment parsing safety and temp file handling

mercihabam:


IrAlfred:
@mercihabam Could you please verify whether this issue also exists on the 1.4.x branch? If so, please apply the same fix there as well.

mercihabam:
@IrAlfred the feature does not exist in the 1.4.x branch. So, the issue isn't there either.

IrAlfred:
> @IrAlfred the feature does not exist in the 1.4.x branch. So, the issue isn't there either.

Thank you @mercihabam for your feedback !

PR #2030, opened 2026-07-05
feat(frontend/backend): enhance Tags module with colors, refactor, and improved test coverage

Shadow243:
As part of improving the **test coverage**, this PR enhances the **Tags** module by integrating the latest changes from Cypht, adding support for tag colors, and improving the overall implementation.

## Changes

* Integrate the latest changes from Cypht into the **Tags** module.
* Add support for tag colors.
* Add tests to improve code coverage.
* Simplify the test structure and improve maintainability.
* Remove unnecessary code and reduce duplication while keeping the implementation clean and straightforward.

This PR does not introduce any functional changes beyond the addition of tag colors; its main focus is improving code quality, maintainability, and test coverage.

Screenshot 2026-07-06 021850


modules/tags/site.js
```diff
@@ -0,0 +1,263 @@
...
+                        triggerLink.html(originalLinkHtml);
+                    }
+                }
+            }
+        );
+    });
+    modal.open();
+};
+
+$(document).on('click', '.tag_add_new_btn', function(e) {
```
kroky, line 228: Don't we need off handlers here in case this is loaded several times?

Shadow243: I think, no need for .off() here: navigation between pages happens via fetch() which only replaces #cypht-main.

 site.js is never re-injected or re-executed. So this document-level delegated binding is only ever attached once and can't be duplicated.

Issue #2037, opened 2026-07-11
🐛 [Bug] Middle click mouse doesn't open the message


petitsurfeur:
Hi 
I just updated from 2.8.0 to 2.11.1. 
I'm on the https://xxx/?page=message_list&list_path=email page
With 2.8.0, middle-click (mouse) on a mail opened the message in a new browser tab. 
With 2.11.1, middle-click (mouse)  on a mail opens a new browser tab with the list of all mails (https://xxx/?page=message_list&list_path=email). 

=> I can't read a mail in a new tab.

IrAlfred:
Thank you @petitsurfeur for submitting this issue, we are going to look into it!

Shadow243:
Thanks, @petitsurfeur, for raising this! Fixed via #2038 (pending review)

PR #2038, opened 2026-07-12
fix(frontend): inline message issue related to middle click mouse doesn't open the message

Shadow243:
related issue https://github.com/cypht-org/cypht/issues/2037

### DIFF
--- a/modules/inline_message/site.css
+++ b/modules/inline_message/site.css
@@ -1,4 +1,13 @@
-.search_content .inline_right, .message_list .inline_right { float: right; width: 50%; max-width: 50%; }
+.search_content .inline_right, .message_list .inline_right {
+    float: right;
+    width: 50%;
+    max-width: 50%;
+    position: sticky;
+    top: 20px;
+    max-height: calc(100vh - 40px);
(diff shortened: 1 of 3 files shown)


modules/inline_message/site.css
```diff
@@ -1,4 +1,13 @@
-.search_content .inline_right, .message_list .inline_right { float: right; width: 50%; max-width: 50%; }
+.search_content .inline_right, .message_list .inline_right {
+    float: right;
+    width: 50%;
+    max-width: 50%;
+    position: sticky;
```
kroky, line 5: @Shadow243 keep in mind Cypht is embedded in Tiki - how would this affect it?

Shadow243: Noted. I'll test it when I get the chance.

Shadow243:
@mbagalwa I'd like to get this one wrapped up, but I'm tied up with a lot of work at the moment. Could you please take a look at the current changes and finish up the remaining items?

mbagalwa:
Got it @Shadow243 , I'll take care of the task

PR #2049, opened 2026-07-24
fix(module/feeds): block SSRF to private and internal feed URLs in 1.4.x

IrAlfred:
Related to #2024


IrAlfred:
Hi @kroky! Could you please take a look at this PR when you get a moment? We are aiming to get this merged quickly if possible

PR #2059, opened 2026-07-29
fix(module/smtp): prevent path traversal in chunked attachment upload

IrAlfred:
Related to https://avan.tech/item167923

Other issues fixed in this pull request

- Compose/send behavior breaks for common real-world names: accented characters are HTML-entity-encoded before storage (so send-time lookup misses the file), 
- comma-separated attachment lists cannot represent names that contain commas, 
- saving a sent copy over IMAP can fail when the message contains multi-byte UTF-8 (APPEND size calculated in characters instead of bytes). 
- In some APPENDUID edge cases, the post-send “review sent message” redirect hits a fatal and leave a blank page.

### DIFF
--- a/modules/smtp/modules.php
+++ b/modules/smtp/modules.php
@@ -261,10 +261,13 @@ class Hm_Handler_process_enable_exclude_auto_bcc extends Hm_Handler_Module {
  */
 class Hm_Handler_get_test_chunk extends Hm_Handler_Module {
     public function process() {
-        $filepath = $this->config->get('attachment_dir');
-        $filepath = $filepath.'/chunks-'.$this->request->get['resumableIdentifier'];
-        $chunk_file = $filepath.'/'.$this->request->get['resumableFilename'].'.part'.$this->request->get['resumableChunkNumber'];
-        if (file_exists($chunk_file)) {
+        $resumable_identifier = trim(basename((string) ($this->request->get['resumableIdentifier'] ?? '')));
+        $resumable_filename = trim(basename((string) ($this->request->get['resumableFilename'] ?? '')));
+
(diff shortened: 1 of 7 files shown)


modules/smtp/modules.php
```diff
@@ -261,10 +261,16 @@ function enable_exclude_auto_bcc_callback($val) { return $val; }
...
+        $resumable_identifier = trim(basename((string) ($this->request->get['resumableIdentifier'] ?? '')));
+        $resumable_filename = trim(basename((string) ($this->request->get['resumableFilename'] ?? '')));
+
         $filepath = $this->config->get('attachment_dir');
-        $filepath = $filepath.'/chunks-'.$this->request->get['resumableIdentifier'];
-        $chunk_file = $filepath.'/'.$this->request->get['resumableFilename'].'.part'.$this->request->get['resumableChunkNumber'];
-        if (file_exists($chunk_file)) {
+        $filepath = $filepath.'/chunks-'.$resumable_identifier;
+        $chunk_file = $filepath.'/'.$resumable_filename.'.part'.$this->request->get['resumableChunkNumber'];
+        if (!in_array($resumable_identifier, ['', '.', '..'], true) &&
```
kroky: If basename does this, why do you do it again here?

Also, as a final check - it would be enough to check dirname of the target file against the expected dir name.

kroky: OK, I see that you check if the target name is not one in the list but it seems better to me to check the actual dirname of the file.

IrAlfred: Agreed! I switched to dirname check

IrAlfred: I've added other changes to fix various issues I found when sending attachments, please review all the changes again

PR #2061, opened 2026-08-04
fix(backend): check select_mailbox() result in get_mailbox_page()

mkraus666:
`get_mailbox_page()` calls `select_mailbox()` without checking the result. The SORT/SEARCH/FETCH commands that follow carry no mailbox argument — they operate on whichever mailbox is currently selected on the connection. When the switch fails, the function keeps going and either returns an empty page or, if another mailbox is still selected, returns messages from *that* mailbox under the requested name. The caller has no way to tell either case from a genuinely empty folder.

This returns `[0, []]` when the select fails, matching what `Hm_Mailbox::get_messages()` already does when `select_folder()` fails, so callers that expect `$messages[1]` keep working.

**How I ran into it**

Debugging an account that consistently returned zero messages. In that particular case the mailbox really was empty, so this was not the root cause — but the silent fall-through cost several rounds of investigation before that was clear, and the wrong-mailbox path is a correctness issue in its own right.

**On the test**

I added a regression test using an anonymous subclass to force `select_mailbox()` to fail, so it needs neither a broken server nor a hand-crafted IMAP fixture.

I could not get it to run green, and not because of the test:

- the `modules_imap` suite is commented out in `tests/phpunit/phpunit.xml` (lines 141-142), so the IMAP tests do not run in CI at all;
- the file it points at is `./modules/imap/hm-imap.php`, but the test file is `hm_imap.php`;
- running the file directly fails on duplicate class declarations (`HTMLToText`, `Test_Uid_Cache`) because `setUp()` uses `require` rather than `require_once` — this hits the existing `test_get_mailbox_page` too, not just my addition.

Happy to drop the test from this PR, or to look at reviving the suite separately — whichever you prefer. I did not want to bundle that into a nine-line fix.

### DIFF
--- a/modules/imap/hm-imap.php
+++ b/modules/imap/hm-imap.php
@@ -2561,7 +2561,15 @@ if (!class_exists('Hm_IMAP')) {
 
             /* select the mailbox if need be */
             if (!$this->selected_mailbox || $this->selected_mailbox['name'] != $mailbox) {
-                $this->select_mailbox($mailbox);
+                /* Bail out if the mailbox cannot be selected. Without this check the
+                 * SORT/SEARCH/FETCH commands below run against whatever mailbox is still
+                 * selected on the connection, so a failed switch either returns an empty
+                 * list or - worse - messages from the previously selected mailbox.
+                 * Same contract as Hm_Mailbox::get_messages(), which already returns
+                 * [0, []] when select_folder() fails. */
+                if (!$this->select_mailbox($mailbox)) {
+                    return array(0, array());
+                }
             }
 
             /* use the SORT extension if we can */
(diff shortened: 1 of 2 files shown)


marclaporte:
Thank you @mkraus666 for the investigation!

@IrAlfred will do a first review, and @kroky the final approval.

FYI, you may want to be more involved in the community:
* monthly community meetings: https://github.com/cypht-org/cypht/wiki/Monthly-Community-Meetings (recordings of previous months are available)
* a community chat: https://gitter.im/cypht-org/community

kroky:
The update looks good. The pipeline failed because of the PR title not conforming to the rules. I've fixed it and merging it now. Thanks!

PR #2070, opened 2026-08-14
refactor(backend): resolve VENDOR_PATH dynamically based on installation layout

mercihabam:
Because `VENDOR_PATH` is always assumed at the APP_PATH level, integrating Cypht in other software often requires overwriting that constant definition with the real path.

This PR provides a way to dynamically resolve that path when Cypht is installed either as a standalone app or a package via Composer.

Related change in Tiki: https://gitlab.com/tikiwiki/tiki/-/merge_requests/10893.

Related issue: https://github.com/cypht-org/cypht/issues/1909

### DIFF
--- a/modules/api_login/api.php
+++ b/modules/api_login/api.php
@@ -7,12 +7,14 @@
 
 /* Constants */
 define('APP_PATH', dirname(dirname(dirname(__FILE__))).'/');
-define('VENDOR_PATH', APP_PATH.'vendor/');
 define('WEB_ROOT', '');
 
 /* Init the framework */
-require_once VENDOR_PATH.'autoload.php';
 require_once APP_PATH.'lib/framework.php';
+
+Hm_Functions::define_vendor_path();
+
+require_once VENDOR_PATH.'autoload.php';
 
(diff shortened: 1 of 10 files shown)


modules/api_login/api.php
```diff
@@ -7,13 +7,15 @@
...
-define('VENDOR_PATH', APP_PATH.'vendor/');
 define('WEB_ROOT', '');
 
 /* Init the framework */
-require_once VENDOR_PATH.'autoload.php';
 require_once APP_PATH.'lib/framework.php';
 
+Hm_Functions::define_vendor_path();
+
+require_once VENDOR_PATH.'autoload.php';
```
kroky, line 14: Ok, but now we load framework and almost all project files before autoloading composer dependencies which means that as soon as one of the files references a composer dependency at load time, it will fail. It is better to keep the order and usually is better to autoload composer before anything else. Maybe define the vendor path loader as a shared function somehow in a single file and require it in all those places before requiring autoload?

mercihabam: Yes, that's better.

PR #2074, opened 2026-08-20
feat(other): make auth configurable

camilevahviraki:
### Issues

- None

### Todo

- [ ] None


.env.example
```diff
@@ -245,5 +275,153 @@ PAGE_PARAM_NAME=page
...
+CYPHT_CONFIG_ENCRYPTION_SECRET=your_encryption_key
+
+# ============================================
+# CYPHTSITE - DATABASE CONFIGURATION
+# ============================================
+
+# Database table prefix used by the integrating application.
+# - Dolibarr: "llx_"
+# - Tiki: "tiki_"
+# - WordPress: "wp_"
```
marclaporte, line 382: AFAIK, no one volunteered to work on WordPress, Joomla! and  Drupal so maybe this should be not there?

PR #2091, opened 2026-09-05
chore(backend): allow isolated execution of handler and output modules

mercihabam:
This PR provides a simple way to run individual modules without necessarily running the entire app. An integration can simulate the necessary components of a module (configuration, request data, etc.), pass them to the caller, and receive real output.

This is leveraged by [Tiki](https://gitlab.com/tikiwiki/tiki/-/merge_requests/11072).


mercihabam:
Thanks! An ajax call to Cypht gives me the expected result. So, closing this.

PR #2098, opened 2026-09-16
test(unit): enhance setup for unit tests by adding user, dovecot, and…

Shadow243:


kroky:
Please request a review when the PR is ready and not in draft mode...

Issue #17, opened 2015-08-19
Web based configuration script


jasonmunro:
Currently we have a set of CLI scripts to create the Cypht configuration and add/remove/modify user accounts (when using the DB authentication method). It would be great to have a web based version of these to make installation easier, especially for shared hosting environments.

dumblob:
In my opinion this is quite a minor issue. In fact, this is a thing of a system-specific configuration tools or principles. It might be some packaging system (rpm, deb, pkgbuild, ebuild, etc.), it might be some configuration system (ansible, chef, puppet, etc.), it might be some specific GUI application, it might be just a text editor (nvim, emacs, nano, etc.). The only solution which from my experience works is well-defined configuration file semantics with well-known structure (e.g. JSON) at a well-known location (e.g. `/etc/`).

jasonmunro:
I agree, but this is a big blocker to adoption, and if people can't try out the software it's hard to build momentum behind it. We include CLI helper scripts for a few things already, it's not too much of a stretch to "web-ify" them. My main concern with both the configuration arrangement we already have and a web based setup script is security.

dumblob:
Right, what I meant was *"begin with a thoroughly defined configuration file (in terms of semantics and syntax)"* as the core feature and *"end with a simple web GUI operating upon this well-defined configuration file"*.

And it's true, that unification is the way to go. Not a set of different user-space scripts with varying interfaces.

Regarding security, do you mean the need to introduce authentication and at the same time providing bug-free web app being safe when reading/writing the configuration file(s)?

jasonmunro:
thanks for the clarification. That is very much the approach I'm taking. The "defined" file is the hm3.ini file, the output is the rc file (serialized array). The web interface is just a way to manage both without a text editor and cli commands, but it ultimately will use the same code path.

WRT security, the issue is authentication. CLI commands have an implicit authentication associated with them (even that is not ideal), but providing a web interface to "bootstrap" an installation is worrisome, if for no other reason than it gives people an easy way to shoot themselves in the foot :)

Kroesss:
Is there any news on this ticket yet? It would be great if it could be used on a shared hosting-environment.

jasonmunro:
@Kroesss, Sorry, this is still pretty low on my priority list right now.

jasonmunro:
As of today I have disabled the half ass web based configuration builder we had in the scripts directory. I did for a number of reasons as described in the commit:

1. It does not work
[...]

PR #2100, opened 2026-09-20
feat(frontend/backend): reat/web based configurations

Shadow243:
Addresses #17 ("Web based configuration script") by providing a simpler install path, especially for shared hosting, without introducing the security risks.

**What was done (in increasing phases):**
- A reusable installation core (Hm_Installer) and a CLI wizard (scripts/install.php), covering directory creation, .env generation, account creation, and config unit tested.
- A dedicated install guide for non-root shared hosting.
- A web installer, disabled by default, gated by a token (served as a dotfile, blocked from direct HTTP access via .htaccess), with session regeneration after token verification and escaping of Dotenv metacharacters in .env values.
- Redesigned as a multi-step wizard with server-side validation, step-scoped required fields, stopping the flow on the first failed step (instead of continuing regardless), and running setup_database/create_account/config_gen in-process (no more shell_exec).

Security: access token, dotfile blocked by .htaccess, disabled by default, in-process execution (no shell), escaped .env values,  the goal being to avoid reopening the attack surface that motivated removing the original installer.


Shadow243:
The DB wizard setup is now working with validation, next step make it work with  `IMAP` & `LDAP`.

PR #2101, opened 2026-09-21
fix(backend): keep the session when the origin carries a port the target lost

### LINKED ISSUE #1989: 🐛 Login page loop
## 🐛 Bug
Cannot log in, login page reloads.

Clean install with docker compose 
```
services:
  db:
    image: mariadb:10
    volumes:
      - ./data/mysql:/var/lib/mysql
    environment:
      - MYSQL_ROOT_PASSWORD=root_password
      - MYSQL_DATABASE=cypht
      - MYSQL_USER=cypht
      - MYSQL_PASSWORD=cypht_password
    healthcheck:
      test: ["CMD", "mysqladmin", "ping", "-h", "localhost", "-ucypht", "-pcypht_password"]
      interval: 10s
      timeout: 5s
      retries: 5

  cypht:
    image: cypht/cypht:2.9.0
    depends_on:
      db:
        condition: service_healthy
    ports:
      - "80:80"
    environment:
      - AUTH_USERNAME=admin
      - AUTH_PASSWORD=admin
      - DB_CONNECTION_TYPE=host
      - DB_DRIVER=mysql
      - DB_HOST=db
      - DB_NAME=cypht
      - DB_USER=cypht
      - DB_PASS=cypht_password
      - SESSION_TYPE=DB
      - USER_CONFIG_TYPE=DB
    volumes:
      - ./data/log/nginx:/var/log/nginx
      - ./data/log/php:/var/log/php
      - ./data/log/supervisord:/var/log/supervisord
      - ./data/fonts:/usr/local/share/cypht/site/fonts
```
no error in logs
```
docker compose up
[+] up 2/2
 ✔ Container test_-db-1    Running                                                                                                                                                        0.0s
 ✔ Container test_-cypht-1 Recreated                                                                                                                                                      4.7s
Attaching to cypht-1, db-1
Container test_-db-1 Waiting 
Container test_-db-1 Healthy 
cypht-1  | Cannot load Xdebug - it was already loaded
cypht-1  | 
cypht-1  | Database connection successful ...
cypht-1  | Database detected. Running migrations...
cypht-1  | Running migrations...
cypht-1  | Migrations completed.
cypht-1  | 
cypht-1  | Db setup finished
cypht-1  | Creating directory for attachments /var/lib/hm3/attachments
cypht-1  | Cannot load Xdebug - it was already loaded
cypht-1  | User 'admin' already exists. Skipping creation...
cypht-1  | Cannot load Xdebug - it was already loaded
cypht-1  | 
cypht-1  | checking dependencies ...
cypht-1  | all dependency checks passed.
cypht-1  | 
cypht-1  | scanning module core ...
cypht-1  | scanning module contacts ...
cypht-1  | scanning module local_contacts ...
cypht-1  | scanning module ldap_contacts ...
cypht-1  | scanning module gmail_contacts ...
cypht-1  | scanning module feeds ...
cypht-1  | scanning module jmap ...
cypht-1  | scanning module imap ...
cypht-1  | scanning module smtp ...
cypht-1  | scanning module account ...
cypht-1  | scanning module idle_timer ...
cypht-1  | scanning module desktop_notifications ...
cypht-1  | scanning module calendar ...
cypht-1  | scanning module themes ...
cypht-1  | scanning module nux ...
cypht-1  | scanning module developer ...
cypht-1  | scanning module profiles ...
cypht-1  | scanning module imap_folders ...
cypht-1  | scanning module sievefilters ...
cypht-1  | scanning module tags ...
cypht-1  | scanning module history ...
cypht-1  | scanning module scheduled_sends ...
cypht-1  | site.css file created
cypht-1  | site.js file created
cypht-1  | dynamic.php file written
cypht-1  | creating production site
cypht-1  | KindEditor assets copied successfully
cypht-1  | copying assets directory...
cypht-1  | 2026-05-29 19:30:48,386 CRIT Supervisor is running as root.  Privileges were not dropped because no user is specified in the config file.  If you intend to run as root, you can set user=root in the config file to avoid this message.
cypht-1  | 2026-05-29 19:30:48,389 INFO supervisord started with pid 1
cypht-1  | 2026-05-29 19:30:49,393 INFO spawned: 'nginx' with pid 19
cypht-1  | 2026-05-29 19:30:49,396 INFO spawned: 'php-fpm' with pid 20
cypht-1  | Cannot load Xdebug - it was already loaded
cypht-1  | [29-May-2026 19:30:49] NOTICE: fpm is running, pid 20
cypht-1  | [29-May-2026 19:30:49] NOTICE: ready to handle connections
cypht-1  | 2026-05-29 19:30:50,487 INFO success: nginx entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
cypht-1  | 2026-05-29 19:30:50,487 INFO success: php-fpm entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
cypht-1  | 127.0.0.1 -  29/May/2026:19:30:56 +0000 "POST /index.php" 303
cypht-1  | 127.0.0.1 -  29/May/2026:19:30:57 +0000 "GET /index.php" 200
```

### Version & Environment
v2.9.0 : not working

I tried version 2.8.0, it logs in no problem. No change in the logs.

kirinriki7777-sys:
## 🍰 Pullrequest

`validate_origin()` compares the source (`Origin`/`Referer`) with the target (`HTTP_HOST`). The source always carries the port the client used. The target does not always carry it, because that depends on what the web server passes to PHP:

- `fastcgi_param HTTP_HOST $host;` drops the client supplied port. This is what the official image ships — `/etc/nginx/fastcgi_params` inside `cypht/cypht:2.12.2` (nginx 1.22.1, Debian bookworm). The comments in that same file say so: `$host` "does not preserve the client-supplied port, while `$http_host` typically does".
- `$http_host`, and Apache, keep it.

`validate_source()` appended the source port before a strict comparison, so on the first kind of deployment every request from a browser using a non-default port looked cross-origin:

```
target = HTTP_HOST          = "localhost"        <- port dropped before PHP saw it
source = parse_url(Origin)  = "localhost:8081"   <- port re-attached by Cypht
"localhost:8081" !== "localhost"  ->  $session->destroy()
```

Authentication itself succeeds. The login response sets the session cookies and expires them in the very same response, which is exactly the reported "login succeeds, page just reloads":

```
Set-Cookie: hm_id=...;      path=/; domain=localhost; HttpOnly; SameSite=Lax
Set-Cookie: hm_session=...; path=/; domain=localhost; HttpOnly; SameSite=Lax
Set-Cookie: hm_session=...; expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; ...
Set-Cookie: hm_id=...;      expires=Thu, 01-Jan-1970 00:00:01 GMT; Max-Age=0; ...
```

This is also why the issue has been hard to reproduce, and why it still occurs for some reporters after the earlier fix: a deployment reached on port 80 or 443 never shows it, while anything on a non-default port shows it every time. It matches the reports in the issue from people using a raw IP with a port, or a proxy forwarding to one.

### Changes

- `lib/module.php` — new `source_matches_target()`. It compares hosts, and compares ports only when both sides actually have one. `validate_source()` and the proxy disambiguation in `source_and_target()` both go through it, so the three places that used to build a `host:port` string by hand are now one.
- `tests/phpunit/module.php` — the ported-origin cases.

The target is parsed with `parse_url('//'.$target)` rather than split on `:`, so an IPv6 literal such as `[::1]:8081` is handled by the same code path.

### What this relaxes, and what it does not

| case | before | after |
| --- | --- | --- |
| port on the source only (`Origin: http://localhost:8081` vs `HTTP_HOST: localhost`) | rejected | **accepted** |
| port on both sides, matching | accepted | accepted |
| port on both sides, different (`http://localhost:9999` vs `localhost:8081`) | rejected | rejected |
| different host, with or without ports | rejected | rejected |

Only the first row changes. When the target has no port, the port was dropped before PHP could see it, so its absence is not evidence of a mismatch — there is nothing to compare against. When both sides do carry a port, the comparison is as strict as before.

This changes one existing expectation in `tests/phpunit/module.php`, which asserted that `HTTP_ORIGIN=http://localhost:123` against `HTTP_HOST=localhost` must be rejected. That assertion is the bug: it is the exact shape of every request from a browser on a non-default port. The two cases that keep the check meaningful — matching ports and mismatching ports — are now asserted explicitly, so the protection is pinned rather than assumed.

### Verification

End to end against the released image: `cypht/cypht:2.12.2` published on `8081:80` with MariaDB, stock configuration, `DISABLE_ORIGIN_CHECK` **not** set. `lib/module.php` is the only file swapped between runs; the requests are otherwise identical, differing only in whether `Host`, `Origin` and `Referer` carry `:8081`.

| `lib/module.php` | request | session cookies expired in the login response | result |
| --- | --- | --- | --- |
| master | with port | 4 | login page served again |
| master | without port (control) | 0 | logged in |
| this PR | with port | 0 | **logged in** |
| this PR | without port (control) | 0 | logged in |

`bash tests/phpunit/run.sh` (sqlite, PHP 8.1.34): 911 tests, 1860 assertions. The same two `Open basedir` failures appear on pristine `master` in this container and are unrelated to the change.

Reverting only `lib/module.php` and keeping the new assertions fails, so the added cases cover the bug rather than describe the new code:

```
 ✘ Validate origin
   │ Failed asserting that false is true.
Tests: 1, Assertions: 5, Failures: 1.
```

### Issues

- fixes #1989

### Todo

- [X] None


IrAlfred (approved):

lib/module.php
```diff
@@ -248,21 +248,9 @@ private function source_and_target($request, $config) {
...
-                }
-                $referer_host = $referer_parts['host'];
-                if (array_key_exists('port', $referer_parts)) {
-                    $referer_host .= ':'.$referer_parts['port'];
-                }
-                if ($origin_host !== $target && $referer_host === $target) {
-                    $source = $referer;
-                }
+            if (!$this->source_matches_target(parse_url($origin), $target) &&
+                $this->source_matches_target(parse_url($referer), $target)) {
```
kroky, line 252: For the sake of consistency, please use parse_url inside the method and not here when calling it. target is already handled like that, so let's also handle source like that. It will be easier to read and understand as well.

PR #2103, opened 2026-09-22
feat(backend): add TLS options for MySQL/MariaDB connections

iiroak:
## 🍰 Pullrequest

Cypht builds its PDO DSN with no driver options, so there is currently no way to connect to a MySQL or MariaDB server that enforces encrypted transport. On a server with `require_secure_transport=ON` the connection is rejected with:

```
SQLSTATE[HY000] [3159] Connections using insecure transport are prohibited
while --require_secure_transport=ON.
```

The UI only surfaces `Unable to connect to database`, and no setting exists to work around it — `lib/db.php` calls `new PDO($dsn, $user, $pass)` with three arguments, and `build_dsn()` returns a fixed string with no room for TLS parameters. Since the Docker image ships PHP with **mysqlnd**, the usual `my.cnf` client-side workaround does not apply either: mysqlnd does not read option files.

### What this adds

Four optional settings in `config/database.php`, read from the environment like every other database setting:

| Setting | Env var | Default |
|---|---|---|
| `db_ssl_ca` | `DB_SSL_CA` | `''` |
| `db_ssl_cert` | `DB_SSL_CERT` | `''` |
| `db_ssl_key` | `DB_SSL_KEY` | `''` |
| `db_ssl_verify_server_cert` | `DB_SSL_VERIFY_SERVER_CERT` | `true` |

A new `build_pdo_options()` translates them into `PDO::MYSQL_ATTR_SSL_*` attributes.

### Notes on the implementation

- **`DB_SSL_CA` is what enables TLS.** I verified that passing `MYSQL_ATTR_SSL_VERIFY_SERVER_CERT` on its own does *not* enable encryption — the server still rejects the connection as insecure. So `VERIFY_SERVER_CERT` is only applied when another SSL attribute is already set, and the docs say as much rather than leaving users to discover it.
- **Non-mysql drivers are untouched**: `build_pdo_options()` returns early unless the driver is `mysql`, so `pgsql` and `sqlite` never receive MySQL-specific attributes.
- **The SSL paths are part of the connection identity**, so they are included in `db_key()`. Without that, two connections differing only in TLS settings would share a cached handle.
- **`DB_SSL_VERIFY_SERVER_CERT=false` is useful in practice** because MySQL and MariaDB auto-generate self-signed certificates on first start, which is what most self-hosted instances end up with.

### Backwards compatibility

With none of the new settings defined, `build_pdo_options()` returns `[]`, and `new PDO($dsn, $user, $pass, [])` behaves identically to the previous three-argument call. Existing deployments see no change.

### Verification

Tested against MariaDB 11.8.6 with `require_secure_transport=ON`, from the `cypht/cypht:2.12.2` image (PHP 8.1, mysqlnd):

- Before the change: connection rejected with SQLSTATE `[HY000] [3159]`.
- After the change, with `DB_SSL_CA=/etc/ssl/certs/ca-certificates.crt` and `DB_SSL_VERIFY_SERVER_CERT=false`: connection established, login page renders, and `SHOW STATUS LIKE 'Ssl_cipher'` reports `TLS_AES_256_GCM_SHA384`.
- `php -l` clean on both modified PHP files.

One thing I could not verify: I did not run the PHPUnit suite, as it needs a database fixture I did not set up. Reading `tests/phpunit/db.php` and `Hm_Mock_Config::get()`, the new keys fall through to their defaults, so `build_pdo_options()` returns `[]` and the asserted `build_dsn()` output is unchanged — but that is reasoning, not a test run, so CI is the real check here.

### Issues
- None

### Todo
- [X] None


IrAlfred:
Thank you @iiroak for your contribution on Cypht project, this will be reviewed shortly

